PR #20, opened 2019-09-26
one common gql call for fundingEligibility, checkoutCustomization

vishakha94:


bluepnume:
(incorporated manually)

PR #24, opened 2019-12-06
Feat/finalize black button

khanghoang:
Provide content for the inline guest black button for JS SDK.

### DIFF
--- a/server/components/buttons/middleware.js
+++ b/server/components/buttons/middleware.js
@@ -98,11 +98,14 @@ export function getButtonMiddleware({ logger = defaultLogger, content: smartCont
         logger.info(req, `button_render_version_${ render.version }`);
         logger.info(req, `button_client_version_${ client.version }`);
 
+        const content = smartContent[locale.country][locale.lang];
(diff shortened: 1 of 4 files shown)


server/components/buttons/middleware.js
```diff
@@ -98,11 +98,18 @@ export function getButtonMiddleware({ logger = defaultLogger, content: smartCont
...
+        const blackButtonText = content.payWithDebitOrCreditCard;
+        
+        // logs for the missing content for the black button
+        if (!blackButtonText) {
+            logger.info(req, `missing_content`, { info: JSON.stringify({ params, locale }) });
+        }
+        
         const buttonHTML = render.button.Buttons({
-            ...params, nonce: cspNonce, csp:   { nonce: cspNonce }, fundingEligibility, personalization
+            ...params, nonce: cspNonce, csp:   { nonce: cspNonce }, fundingEligibility, personalization, blackButtonText
```
bluepnume: Can you just pass the entire content object? Don't want to have to keep changing this as we consume more and more content from xo-content

PR #25, opened 2020-02-05
Inline wallet component

vishakha94:
WIP in this open PR
- [ ] The SSR for this is currently broken. When I use hooks along with `preact-render-to-string`, I keep getting the error `Uncaught TypeError: Cannot read property '__H' of undefined`. This issue has been mentioned(https://github.com/preactjs/preact/issues/1373) and fixed here(https://github.com/preactjs/preact-render-to-string/pull/77) but the solve has not been able to work for me yet. 
- [ ] Responsive code & CSS for wallet component needs polish
- [ ] To call handler for access code to access token exchange
- [ ] Event handling for wallet component
- [ ] flow


server/service/checkoutSession.js
```diff
@@ -0,0 +1,143 @@
...
+        
+        const checkoutSession = result.checkoutSession;
+        console.log('the checkout session is: ', checkoutSession);
+        
+        return checkoutSession;
+        
+    } catch (err) {
+        console.log('the error is: ', err);
+        logger.error(req, 'checkout_session_error_fallback', { err: err.stack ? err.stack : err.toString() });
+        // TODO: what should be the default behaviour??
```
bluepnume: Just throw the error.

PR #28, opened 2020-02-25
update OrdervValidation on msp.

weihou:
has dependency of https://github.paypal.com/Checkout-R/xobuyernodeserv/pull/1325


bluepnume:
Will wait for this PR to be fully ready before review/merge. We do need full validation here before we go live with this, otherwise we'll end up with bad integrations we can never fix.

bluepnume:
This looks great, thanks. Assuming this uses new GQL schema? If so we'll merge and publish once that is ready/deployed.

bluepnume:
Can you rebase and resolve? Had to make some changes for a different feature. Pulled out the gql call to another file.

PR #34, opened 2020-03-09
PayPal Credit wallet components

Seavenly:


src/wallet/components/credit/transition/index.jsx
```diff
@@ -0,0 +1,32 @@
+/* @flow */
```
bluepnume: I would say let's decouple these two things:

- Helper function to build the credit url and do the redirect
- Generic loading component (with context to enable/disable loading) that can be used for any loading scenario

bluepnume:
Minor comment, otherwise looks good to me!

PR #35, opened 2020-03-10
Update plan id for approveOrder, add rewards schema gql

vishakha94:
- updated plan id to be used for approveOrder
- adding rewards schema as part of the xosession gql call

### DIFF
--- a/src/wallet/wallet.jsx
+++ b/src/wallet/wallet.jsx
@@ -5,11 +5,12 @@
 import renderToString from 'preact-render-to-string';
 import { h, render } from 'preact';
 import { ZalgoPromise } from 'zalgo-promise/src';
+import { useState, useEffect, useRef } from 'preact/hooks';
 
 import { getBody } from '../lib';
 import { approveOrder } from '../api';
 
-import type { CheckoutSession } from './types';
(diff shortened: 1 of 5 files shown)


src/wallet/wallet.jsx
```diff
@@ -45,13 +38,38 @@ type SetupOptions = {|
...
+
+export function App({ props, cspNonce, checkoutSession, buyerAccessToken } : AppProps) : Node {
+    const { fundingOptions } = checkoutSession;
+    const [ selectedWalletItem, setSelectedWalletItem ] = useState(fundingOptions[0]);
+    useSetup({ props, selectedWalletItem, buyerAccessToken });
+    
     return (
         <StyleSheet cspNonce={ cspNonce }>
-            <Page checkoutSession={ checkoutSession } />
+            <Page checkoutSession={ checkoutSession } selectedWalletItem={ selectedWalletItem } setSelectedWalletItem={ setSelectedWalletItem } />
```
bluepnume: Instead of passing both of these down, can we just store the state in both places, and pass down an `onChangeWalletItem` callback? That keeps the parent and child components a little less coupled.

bluepnume:
Looks great 👍 one minor comment

PR #36, opened 2020-04-08
fix payee in gql query 

weihou:


bluepnume:
Can we please get a test case that fails when the GQL params are incorrect? Thanks.

PR #38, opened 2020-04-27
Feat - Support 3rd party token for subscriptions

ravishekhar:
* Support 3rd party token for subscription 

* Recreate the access token with 3rd party credentials with target as merchantID if merchantID is passed. 
* Create the token on button render when createSubscription prop is passed to Button so that memoization can store the access token for use while clicking the button.
* On Button Click call create Token method and let memoized value to be used if available.


src/api/auth.js
```diff
@@ -9,12 +9,19 @@ import { HEADERS } from '../constants';
 
 import { callGraphQL } from './api';
 
-export function createAccessToken (clientID : string) : ZalgoPromise<string> {
+export function createAccessToken (clientID : ?string, targetSubject? : $ReadOnlyArray<string>) : ZalgoPromise<string> {
```
bluepnume, line 12: Can we make this an object with optional options? 

```
createAccessToken(clientID, { targetSubject: [ 'foo', 'bar' ] })
```

bluepnume:
Looks good to me, just one minor comment & will merge. Thanks!

src/api/auth.js
ravishekhar: Updated!

PR #40, opened 2020-05-05
Add parentDomain query param to native popup url

abarco:


src/payment-flows/native.js
```diff
@@ -248,8 +248,9 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     });
 
     const getNativePopupUrl = memoize(({ sessionUID }) : string => {
-        return extendUrl(`${ getNativePopupDomain() }${ NATIVE_CHECKOUT_POPUP_URI[fundingSource] }`, {
-            query: { sdkMeta, sessionUID, buttonSessionID }
+        const parentDomain = getNativePopupDomain();
```
bluepnume, line 251: This should be `getDomain()`. `getNativePopupDomain()` will return the child domain.

src/payment-flows/native.js
```diff
@@ -248,8 +248,9 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
+        return extendUrl(`${ parentDomain }${ NATIVE_CHECKOUT_POPUP_URI[fundingSource] }`, {
```
bluepnume, line 252: this should still be `getNativePopupDomain()`

PR #42, opened 2020-05-14
msp support

weihou:
* get branded from the 2nd fundingEligibility call
* support msp in validateOrder - get payees from purchaseUnits of xosession or from getPayee (for BA transaction)


src/button/orders.js
```diff
@@ -27,13 +27,64 @@ type ValidateOptions = {|
...
+    const foundPayee = payees.every(payee => {
+        return (merchantIds.indexOf(payee.merchantId) > -1 || merchantEmails.indexOf(payee.email && payee.email.stringValue && payee.email.stringValue.toLowerCase()) > -1);
+    });
+    return foundPayee;
+}
+
 export function validateOrder(orderID : string, { env, clientID, merchantID, expectedCurrency, expectedIntent } : ValidateOptions) : ZalgoPromise<void> {
     return ZalgoPromise.hash({
         order: getSupplementalOrderInfo(orderID),
         payee: getPayee(orderID)
```
bluepnume, line 85: Hi -- I missed that this was re-added to the PR. From Slack I thought the plan was to:

> 1. update the getPayee call response in xo-hermes-checkout so that it will have the payee_details array.
> 2. add payee_details into gql get payee response in xobuyernodeserv
> 3. change the spb order validation function.

weihou: xobuyernodeserv will return all payees in checkoutSession.payees

PR #45, opened 2020-06-03
Support paylater funding source

Seavenly:


src/button/button.js
```diff
@@ -97,6 +97,10 @@ export function setupButton(opts : ButtonOpts) : ZalgoPromise<void> {
             if (isEnabled()) {
                 paymentProcessing = true;
 
+                if (fundingSource === FUNDING.PAYLATER) {
+                    payment.creditProductIdentifier = 'PAY_LATER_GB';
```
bluepnume, line 101: Hi, can you give some context on where this is used?

bluepnume:
Added do-not-merge label. Please remove when gql changes are live so this can be safely merged/released. Thanks!

src/button/button.js
Seavenly: This value is passed to the checkout zoid component to append a new query parameter (`credit_product_identifier=PAY_LATER_GB`) which is required to open up the application.

src/payment-flows/checkout.js
```diff
@@ -285,7 +285,9 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
             locale,
             commit,
             cspNonce,
-            clientMetadataID: cmid
+            clientMetadataID: cmid,
+            creditProductIdentifier,
```
bluepnume, line 289: For now, if we need, let's derive this in xo-landing-router. I'd prefer to come up with a more generic key for child products of funding sources.

PR #46, opened 2020-06-17
change the message on incorrect client ID

weihou:
the original message is a little confusing because there is no merchant id passed. The client used the wrong client ID in this case.


westeezy:
Closing old PRs. If I closed this erroneously please feel free to reopen and I apologize for any inconvenience

PR #47, opened 2020-07-10
Introducing upgradeLSAT calls on onAuth callback

abarco:
This feature is meant to deprecate clientside API calls to smartcomponentnodeweb with the intention of fulfilling the transaction.

### DIFF
--- a/src/props/onAuth.js
+++ b/src/props/onAuth.js
@@ -1,20 +1,21 @@
 /* @flow */
 
 import { ZalgoPromise } from 'zalgo-promise/src';
-import { memoize } from 'belter/src';
-import type { CreateOrder } from './createOrder';
-import { getLogger } from '../lib';
+
 import { upgradeFacilitatorAccessToken } from '../api';
+import { getLogger } from '../lib';
+
+import type { CreateOrder } from './createOrder';
 
 export type XOnAuthDataType = {|
-    accessToken: string
+    accessToken : string
 |};
 
 export type OnAuth = () => ZalgoPromise<void>;
 
-export function getOnAuth({ facilitatorAccessToken, createOrder, isLSATExperiment } : {| facilitatorAccessToken : string, createOrder: CreateOrder, isLSATExperiment: boolean |}) : OnAuth | void {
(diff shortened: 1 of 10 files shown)


src/props/onAuth.js
```diff
@@ -0,0 +1,32 @@
...
+    return ({ accessToken }) => {
+        getLogger().info(`spb_onauth_access_token_${ (accessToken || buyerAccessToken)  ? 'present' : 'not_present' }`);
+
+        if (accessToken) {
+            return ZalgoPromise.try(() => {
+               // if (experiment) {
+               //     do the thing
+               // }
+
+               return createOrder().then(orderID => upgradeFacilitatorAccessToken(facilitatorAccessToken, {buyerAccessToken: accessToken, orderID}));
```
bluepnume, line 26: Do we also wanna check isLSATExperiment here? Otherwise we'll go straight to upgrading 100% of tokens, right?

src/props/onAuth.js
```diff
@@ -0,0 +1,32 @@
...
+            }).then(() => {
+                return accessToken;
+            });
```
bluepnume, line 29: I think we should catch and log here, and not fail. Since most of the time everything will still work even if upgrade fails (e.g. if it's a server side integration)

src/props/onAuth.js
```diff
@@ -0,0 +1,32 @@
...
+export type OnAuth = () => ZalgoPromise<void>;
+
+export function getOnAuth({ facilitatorAccessToken, createOrder } : {| facilitatorAccessToken : string, createOrder: CreateOrder |}) : OnAuth | void {
+
+    return ({ accessToken }) => {
+        getLogger().info(`spb_onauth_access_token_${ (accessToken || buyerAccessToken)  ? 'present' : 'not_present' }`);
+
+        if (accessToken) {
+            return ZalgoPromise.try(() => {
+               // if (experiment) {
```
bluepnume, line 22: Lol I missed this.

bluepnume:
Ready for merge?

PR #48, opened 2020-07-16
Native instrumentation

abarco:
Instrumentation for the following usecases:
 - When setting the props, (we need to know if all of them will be making it to firebase `process_set_props_attempt`)
 - When attempting the app-switch, (we need to know what will be making it in the url to native `app_switch_attempted`)
 - When coming back from native, (we need a log that confirms that the popup was closed `process_popup_closed`)

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -107,7 +107,7 @@ function isNativeOptedIn({ props } : {| props : ButtonProps |}) : boolean {
 let initialPageUrl;
 
 function isNativeEligible({ props, config, serviceData } : IsEligibleOptions) : boolean {
-    
+
     const { platform, onShippingChange, createBillingAgreement, createSubscription, env } = props;
     const { firebase: firebaseConfig } = config;
     const { eligibility } = serviceData;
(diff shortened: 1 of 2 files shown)


bluepnume:
Looks good to me just needs `npm run lint -- --fix`

src/payment-flows/native.js
```diff
@@ -198,6 +198,21 @@ type NativeSDKProps = {|
...
+        const sanitizedProps = {
+            ...props,
+            facilitatorAccessToken : props && props.facilitatorAccessToken ? '********************' : ''
+        };
+
+        getLogger().info('native_setprops_request', sanitizedProps).track({
+            [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_SET_PROPS_ATTEMPT
+        });
+    } else {
+        getLogger().info('native_setprops_request_has_enpty_props');
```
bluepnume, line 212: empty

src/payment-flows/native.js
```diff
@@ -198,6 +198,21 @@ type NativeSDKProps = {|
     forceEligible : boolean
 |};
 
+function instrumentNativeSDKProps(props : NativeSDKProps) {
```
bluepnume, line 201: First null check for props

src/payment-flows/native.js
```diff
@@ -198,6 +198,21 @@ type NativeSDKProps = {|
...
+    if (props) {
```
bluepnume, line 202: Second null check for props

src/payment-flows/native.js
```diff
@@ -198,6 +198,21 @@ type NativeSDKProps = {|
...
+        const sanitizedProps = {
+            ...props,
+            facilitatorAccessToken : props && props.facilitatorAccessToken ? '********************' : ''
```
bluepnume, line 205: Third null check for props. 

you really don't trust props are there eh!? hahaha

abarco:
re-test this please

PR #58, opened 2020-07-23
map 4xx to invalid clientd id client error

vishakha94:


server/components/buttons/middleware.js
```diff
@@ -104,7 +104,7 @@ export function getButtonMiddleware({ logger = defaultLogger, content: smartCont
             try {
                 facilitatorAccessToken = await facilitatorAccessTokenPromise;
             } catch (err) {
-                if (err && err.code === AUTH_ERROR_CODE.INVALID_CLIENT) {
+                if ((err && err.code === AUTH_ERROR_CODE.INVALID_CLIENT) || (err && err.statusCode >= 400 && err.statusCode < 500)) {
```
bluepnume, line 107:
```suggestion
                if (err && err.statusCode && err.statusCode >= 400 && err.statusCode < 500) {
```

PR #62, opened 2020-08-12
maint/logging

mnicpt:
For app switch funnel drop-off logging.
Provides logging for Firebase connection and changes the name of one log to `popup_shown` as outlined in this doc where @abarco is mentioned: https://engineering.paypalcorp.com/confluence/display/Checkout/New+Startup+Funnel

### DIFF
--- a/src/api/socket.js
+++ b/src/api/socket.js
@@ -512,7 +512,7 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
     
                 getLogger().info('firebase_connection_opened').track({
                     [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
-                    [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.FB_CONNECTION_OPENED
+                    [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.FIREBASE_CONNECTION_OPENED
                 }).flush();
 
                 for (const handler of onOpenHandlers) {
@@ -539,7 +539,7 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
         databasePromise.catch(err => {
             getLogger().info('firebase_connection_errored', { err: stringifyError(err) }).track({
                 [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
-                [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.FB_CONNECTION_ERRORED
+                [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.FIREBASE_CONNECTION_ERRORED
             }).flush();
         });
 
--- a/src/api/socket.js
+++ b/src/api/socket.js
@@ -510,7 +510,7 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
 
                 open = true;
     
-                getLogger().info('smart_button').track({
+                getLogger().info('firebase_connection_opened').track({
                     [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
                     [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.FB_CONNECTION_OPENED
                 }).flush();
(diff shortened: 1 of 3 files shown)


src/payment-flows/native.js
```diff
@@ -513,7 +513,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
 
                     getLogger().info(`native_attempt_appswitch_url_popup`, { url: nativeUrl })
                         .track({
-                            [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_ATTEMPT_APP_SWITCH
+                            [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.POPUP_SHOWN
```
mnicpt, line 516: @abarco Why is this here rather than after line 483?

abarco: if you want to track the popup shown, it should be in line 483. But this event is meant to track the attempt to appswitch. Maybe there's a confusion on the confluence doc. I think this line shouldn't be modified. But we should create this tracking log on 483

src/api/socket.js
```diff
@@ -507,6 +510,11 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
 
                 open = true;
     
+                getLogger().info('smart_button').track({
```
bluepnume, line 513: Instead of 'smart_button' can we have 'firebase_connection_opened'? This will be the event name in CAL

src/payment-flows/native.js
```diff
@@ -436,12 +436,16 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const initDirectAppSwitch = ({ sessionUID } : {| sessionUID : string |}) => {
         const nativeUrl = getNativeUrl({ sessionUID });
 
-        getLogger().info(`native_attempt_appswitch_url_direct`, { url: nativeUrl })
+        const nativeWin = popup(nativeUrl);
+        getLogger().info(`native_attempt_appswitch_popup_shown`, { url: nativeUrl })
```
bluepnume, line 440: Not clear on what the value is of having two events triggered at the exact same time?

mnicpt: The reason is that there was a requirement to have `popup_shown` and for Android, these 2 events happen at the same time; whereas for iOS, they don't so you see the `native_attempt_appswitch_popup_shown` happening in the `createOrder` call within `initPopupAppSwitch`.

mnicpt: @bluepnume Going to sync with @abarco about removing the second one.

bluepnume: Fair. Let's chain them together:

```
getLogger()
    .info(`native_attempt_appswitch_popup_shown`, { url: nativeUrl })
    .info(`native_attempt_appswitch_url_popup`, { url: nativeUrl })
    .track({
        [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.POPUP_SHOWN
    });
    .track({
        [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_ATTEMPT_APP_SWITCH
    }).flush();
```

src/constants.js
```diff
@@ -112,7 +115,8 @@ export const FPTI_TRANSITION = {
     NATIVE_ATTEMPT_APP_SWITCH: ('app_switch_attempted' : 'app_switch_attempted'),
     NATIVE_POPUP_CLOSED:       ('process_popup_closed' : 'process_popup_closed'),
 
-    HONEY_IDENTIFY:           ('honey_identify')
+    POPUP_SHOWN:               ('popup_shown' : 'popup_shown'),
```
bluepnume, line 118: Can we make this `NATIVE_POPUP_SHOWN`

src/constants.js
```diff
@@ -104,6 +104,9 @@ export const FPTI_TRANSITION = {
 
     CONNECT_REDIRECT:         ('process_connect_redirect' : 'process_connect_redirect'),
 
+    FB_CONNECTION_OPENED:     ('fb_connection_opened' : 'fb_connection_opened'),
```
bluepnume, line 107: Let's use `FIREBASE_CONNECTION_OPENED`. "FB" is a loaded acronym, also means "fraudnet beacon" :D

bluepnume:
Thx a mil!

PR #64, opened 2020-08-20
maint/logging

mnicpt:
**Changes**
- Log failure for setProps
- Update logging with state name for popup shown and native attempt app switch
- Log failure for native attempt app switch for iOS and Android functions
- Remote expected `awaitRedirectError` from test


test/client/native.js
```diff
@@ -2736,7 +2736,7 @@ describe('native ios cases', () => {
                         data:   {
                             pageUrl: `${ window.location.href }#close`
                         }
-                    }).catch(expect('awaitRedirectError'));
```
bluepnume, line 2739: Hmm, this should still pass right?

mnicpt: It's a question I have.  Notice that I am catching errors in that method now and rethrowing in order to close Firebase connection and log for app switch funnel.

mnicpt: The error I get if I leave this test is: `Error: Expected awaitRedirectError to be called`

PR #70, opened 2020-09-03
trigger validation for smartfields

andyfen:
for specific fundingSources

### DIFF
--- a/src/lib/comms.js
+++ b/src/lib/comms.js
@@ -1,0 +1,24 @@
+/* @flow */
+import { getAllFramesInWindow, isSameDomain } from 'cross-domain-utils/src';
+
+import type { SmartFields } from '../types';
+import { FRAME_NAME } from '../constants';
+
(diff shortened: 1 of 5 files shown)


src/lib/comms.js
```diff
@@ -0,0 +1,18 @@
...
+import { getAllFramesInWindow, isSameDomain } from 'cross-domain-utils/src';
+
+import type { SmartFields } from '../types';
+
+export function getFieldsByFundingSource(fundingSource : string) : ?SmartFields {
+
+    try {
+        for (const win of getAllFramesInWindow(window)) {
+        // $FlowFixMe
+            if (isSameDomain(win) && win.exports && win.exports.fundingSource === fundingSource) {
```
bluepnume, line 11: Can we keep the `exports.name`? We may have other components in future and need to be able to differentiate.

bluepnume:
Looks great! One small comment then happy to merge this in.

PR #71, opened 2020-09-09
add confirm call for alt payments

andyfen:
adds the confirm call when it is for a smartfield

### DIFF
--- a/src/button/button.js
+++ b/src/button/button.js
@@ -108,7 +108,7 @@ export function setupButton(opts : ButtonOpts) : ZalgoPromise<void> {
             if (isEnabled()) {
                 paymentProcessing = true;
 
-                return initiatePaymentFlow({ payment, config, serviceData, components, props: paymentProps }).finally(() => {
+                return initiatePaymentFlow({ payment, config, serviceData, components, props: paymentProps, smartFields }).finally(() => {
                     paymentProcessing = false;
                 });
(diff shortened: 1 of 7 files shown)


src/button/button.js
```diff
@@ -100,6 +100,12 @@ export function setupButton(opts : ButtonOpts) : ZalgoPromise<void> {
                     }
                     return;
                 }
+
```
bluepnume, line 103: Can we do this?

```javascript
if (smartFields) {
    if (!smartFields.isValid()) {
        if (win) {
            win.close();
        }
        return;
    }
}

if (onClick) {
    onClick({ fundingSource: paymentFundingSource });
}

const confirm = (orderID) => {
        if (smartFields && smartFields.confirm) {
            return createOrder().then(smartFields.confirm);
        }
    });
}

if (isEnabled()) {
    paymentProcessing = true;

    return initiatePaymentFlow({ payment, config, serviceData, components, props: paymentProps, confirm }).finally(() => {
        paymentProcessing = false;
    });
} else  {
    if (win) {
        win.close();
    }
}
```

Otherwise you're skipping onClick, which may interrupt the flow if the merchant rejects due to some other validation.

bluepnume:
Looks good -- one small re-sequencing required though. Could we also get some tests? Thanks!

PR #74, opened 2020-09-21
Move 'Powered by PayPal' below card fields container

allen-tong:
repositioned "Powered by PayPal" as requested

before:
Screen Shot 2020-09-21 at 4 45 42 PM

after:
Screen Shot 2020-09-21 at 4 45 59 PM

### DIFF
--- a/server/components/buttons/middleware.js
+++ b/server/components/buttons/middleware.js
@@ -146,7 +146,6 @@ export function getButtonMiddleware({
                     <style nonce="${ cspNonce }">${ buttonStyle }</style>
                     
                     <div id="buttons-container" class="buttons-container">${ buttonHTML }</div>
-                    <div id="card-fields-container" class="card-fields-container"></div>
 
                     ${ meta.getSDKLoader({ nonce: cspNonce }) }
                     <script nonce="${ cspNonce }">${ client.script }</script>


mstuart:
Whoa, is that an "X" icon in inline guest? :D

bluepnume:
A few comments:

- Can we fix the width of the fields to match the width of the buttons?
- Having a cross button sitting between the buttons and the fields, feels kinda weird. Can we not do something like add a link beneath 'pay now' saying something like 'cancel and return to merchant?'

src/payment-flows/card-fields.js
```diff
@@ -10,12 +10,25 @@ import { unresolvedPromise, promiseNoop } from '../lib';
...
+
+    return { buttonsContainer, cardButtonsContainer, cardFieldsContainer };
+};
+
 function setupCardFields() {
-    // pass
+    const { buttonsContainer, cardFieldsContainer } = getElements();
+    const buttonContainer = buttonsContainer.querySelector('.paypal-button-container');
+    const poweredBy = buttonContainer.querySelector('.paypal-powered-by');
+    buttonContainer.insertBefore(cardFieldsContainer, poweredBy);
```
bluepnume, line 29: Can we just add `#card-fields-container` to the correct place in the dom from the start rather than swapping around elements on the fly?

allen-tong:
> A few comments:
> 
>     * Can we fix the width of the fields to match the width of the buttons?
> 
>     * Having a cross button sitting between the buttons and the fields, feels kinda weird. Can we not do something like add a link beneath 'pay now' saying something like 'cancel and return to merchant?'

Agreed, but is it ok for us to make changes like this :o since this is what design team wanted

src/payment-flows/card-fields.js
allen-tong: yeah that sounds a lot better, do you know where in the code the SPBs are originally added to the DOM? i was trying this way in the beginning but couldn't find it

allen-tong:
This issue has been resolved in the following PRs in `paypal-checkout-components`:
https://github.com/paypal/paypal-checkout-components/pull/1471
https://github.com/paypal/paypal-checkout-components/pull/1473

`card-fields-container` is now located [there](https://github.com/paypal/paypal-checkout-components/pull/1471/files#diff-3fba3feb650bae3f6bfc93d232bac1d8cfde682d5bfffbdd40d8cdfeda7fcd8aR172), so we can remove it here :)

gregjopa (approved):
Nice! Get that div out of here 😄

PR #76, opened 2020-09-23
maint/logging

mnicpt:
Adds better defined logging around Firebase connection errors and closes native connection when it occurs.  This fixes the scenario where Android hangs because of error on web and close message is not sent to alert them of SPB error.

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -344,10 +344,12 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
                 return socket.send(SOCKET_MESSAGE.SET_PROPS, sdkProps);
             }).then(() => {
                 getLogger().info(`native_response_setprops`).track({
+                    [FPTI_KEY.STATE]:           FPTI_STATE.BUTTON,
                     [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_APP_SWITCH_ACK
                 }).flush();
             }).catch(err => {
                 getLogger().info(`native_response_setprops_error`).track({
+                    [FPTI_KEY.STATE]:           FPTI_STATE.BUTTON,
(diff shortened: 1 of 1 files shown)


src/payment-flows/native.js
```diff
@@ -63,8 +63,21 @@ const getNativeSocket = memoize(({ sessionUID, firebaseConfig, version } : Nativ
...
+
     nativeSocket.onError(err => {
-        getLogger().error('native_socket_error', { err: stringifyError(err) });
+        getLogger().error('native_socket_error', { err: stringifyError(err) })
+            .track({
+                [FPTI_KEY.STATE]:           FPTI_STATE.BUTTON,
+                [FPTI_KEY.TRANSITION]:      FPTI_TRANSITION.NATIVE_APP_SWITCH_ACK,
+                [FTPI_CUSTOM_KEY.ERR_DESC]: `[Native Socket Error] ${ stringifyError(err) }`
+            }).flush();
+        closeNative();
```
gregjopa, line 80: I'm not familiar with the types of errors this socket `onError()` callback receives. Is it safe to call `closeNative()` for any type of error received by this callback? Do we need a specific check on the `err` payload to ensure its an error we can't recover from?

mnicpt: Great question, @gregjopa .  This error callback will catch any errors when the Firebase socket tries to send a message.  Specifically, in our case, calling `setProps`.  What this is doing is sending a message to the native app, which then signals to them to close their app and will fallback to web checkout.  This also cleans up any remaining tasks by running the rest that are queued up if they haven't been run.  This does not close the Firebase connection.

gregjopa: Thanks for clarifying @mnicpt. The `closeNative()` behavior makes more sense now.

gregjopa (approved):
👍

PR #78, opened 2020-09-25
iOS App Switch Bypass Firebase

amyegan:
Provide props via query parameters on the universal link so we can avoid errors associated with Firebase. Existing Firebase route remains as a fallback.


mstuart (approved):

gregjopa:
@amyegan these changes look great! 

Would you be able to add some new unit tests for these native use cases? It looks like there's already some [tests for ios](https://github.com/paypal/paypal-smart-payment-buttons/blob/master/test/client/native.js#L1567).

src/payment-flows/native.js
```diff
@@ -430,9 +460,10 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         };
     });
 
-    const detectAppSwitch = once(({ sessionUID } : {| sessionUID : string |}) => {
+    const detectAppSwitch = once(({ sessionUID, orderId } : {| sessionUID : string, orderId : string |}) => {
```
mnicpt, line 463: I'm not seeing where you are passing `orderId` to this method.  Also, I'm wondering if this is the place to log `orderId`.  Not feeling good about changing this interface purely for logging.

mnicpt: Maybe log in `instrumentNativeSDKProps` as this method will only get called if it is an app switch.

src/payment-flows/native.js
```diff
@@ -430,9 +460,10 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
         getLogger().info(`native_detect_app_switch`).track({
-            [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_DETECT_APP_SWITCH
+            [FPTI_KEY.TRANSITION]:      FPTI_TRANSITION.NATIVE_DETECT_APP_SWITCH,
+            [FTPI_CUSTOM_KEY.INFO_MSG]: orderId
```
bluepnume, line 466: I think we already have `FPTI_KEY.ORDER_ID` for this.

src/payment-flows/native.js
```diff
@@ -284,12 +289,32 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
     });
 
-    const getNativeUrl = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
+    const getNativeUrlForAndroid = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
         return extendUrl(`${ getNativeDomain() }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
             query: { sdkMeta, sessionUID, buttonSessionID, pageUrl }
         });
     });
 
+    const getNativeUrl = memoize(({ sessionUID, pageUrl = initialPageUrl, sdkProps } : {| sessionUID : string, pageUrl : string, sdkProps : NativeSDKProps |}) : string => {
```
bluepnume, line 298: Is there a reason we can't send the same URL for Android too, for consistency? My feeling is the less we know/care about the platform, the easier this will be to maintain.

mnicpt: @bluepnume The idea was to leave Android alone for now and isolate these changes to iOS.  We plan to do a POC for Android and see using the same code we've done here for Android as well so we no longer have 2 paths.  Just not sure if we send more in the universal link if that will break Android.  It shouldn't.

src/payment-flows/native.js
```diff
@@ -579,19 +609,39 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
         const onCompleteListener = listen(popupWin, getNativePopupDomain(), POST_MESSAGE.ON_COMPLETE, () => {
             getLogger().info(`native_post_message_on_complete`).flush();
             popupWin.close();
         });
 
+        const onErrorListener = listen(popupWin, getNativePopupDomain(), POST_MESSAGE.ON_ERROR, (data) => {
+            onErrorCallback(data);
+        });
+
+        const onShippingChangeListener = listen(popupWin, getNativePopupDomain(), POST_MESSAGE.ON_SHIPPING_CHANGE, (data) => {
```
bluepnume, line 630: We decided against handling this one from the popup, right? I think we should just make sure for now, iOS + onShippingChange is marked as ineligible for the native flow.

mnicpt: @bluepnume Yes.  I will remove it.

bluepnume:
This is looking pretty good to me! Few minor comments but no major concerns.

src/payment-flows/native.js
amyegan: I changed it to log the url when we attempt app switch in awaitRedirectListener so we capture the difference between old and new web and track potential problems in the url.

src/payment-flows/native.js
amyegan: This was done to leave the Android flow as unchanged as possible to avoid issues with that path. If it makes more sense to use the same URL for both then we could do that. It shouldn't hurt anything to pass the extra information to the Android app

src/payment-flows/native.js
amyegan: I removed onShippingChangeListener from the iOS path

src/payment-flows/native.js
amyegan: Steve brought up that this doesn't seem to be the best place to log this. I've changed it to log the native URL when we attempt the app switch instead

mnicpt (approved):
LGTM!

bluepnume:
Thanks all!

PR #79, opened 2020-10-05
fix/AndroidWebListener

mnicpt:
Android app switching logs are present after web switching and causing the Android to hang waiting for props from Firebase.  This PR proposes that we send a close message to native when web switching to ensure a better user experience by letting the app know it needs to close.

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -438,10 +438,12 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         return connectNative({ sessionUID }).setProps();
     });
 
-    const detectWebSwitch = once((fallbackWin : ?CrossDomainWindowType) => {
+    const detectWebSwitch = once(({ sessionUID, fallbackWin } : {| sessionUID : string, fallbackWin : ?CrossDomainWindowType |}) => {
         getLogger().info(`native_detect_web_switch`).track({
             [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_DETECT_WEB_SWITCH
         }).flush();
+
+        connectNative({ sessionUID }).close();
 
         return fallbackToWebCheckout(fallbackWin);
     });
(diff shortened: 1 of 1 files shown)


bluepnume:
Hmm. This suggests something is iffy in our app switch detection, right?

mnicpt:
@bluepnume My guess is there is a validation error occurring causing the SPB to fallback and not in any of the other error scenarios where we close native.  This is for Android so it doesn't seem to be a timing issue as we attempt to app switch immediately.

bluepnume:
Hmm, the validation case already calls `close()`: https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/payment-flows/native.js#L496

mnicpt:
@bluepnume Correct.  A question I have where this may occur is do you know why we have `detectWebSwitchListener` for Android if we don't open a popup url?  Not sure which window would be posting this message? https://github.com/paypal/paypal-smart-payment-buttons/blob/f449b3d9a6826a467bbc0ad192294f1eea531c81/src/payment-flows/native.js#L485 This is the only place I could think that we weren't covering validation errors but wasn't sure who was posting a message to this listener.

bluepnume:
We do open a popup in Android. That popup will point to `/smart/checkout/native` and will send the detectWebSwitch post message.

It's just, in the Android case, if the app is installed, the popup will never fully open.

https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/payment-flows/native.js#L469

mnicpt:
I see now.  It opens with `nativeCheckoutFallback.jsx`. It immediately sends a `detectWebSwitch` message back.  So if the app switch doesn't happen fast enough, it's possible that we go down the web switch path and app switch at the same time?  That's what we see in the logs.

bluepnume:
We do also have this additional redundant check here: https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/payment-flows/native.js#L504

That brings us back to the question: is the app-switch detection (`didAppSwitch()`) failing here, or giving unexpected results? 

On Android, when the app switch happens, `!popupWin || isWindowClosed(popupWin)` should always be `true`. If that assumption is incorrect, then yes we have a problem and potentially we could be going down both paths at the same time.

There is a bit of a race condition here. `!popupWin || isWindowClosed(popupWin)` is only accurate after a few hundred ms, which is why we're waiting for the `createOrder()` call to complete right now. It may be we need to add an artificial delay here in case `createOrder()` returns too fast. Maybe `max<createOrder, 500>`?

bluepnume:
Artificial delays are awful though :( It may be good to revisit if there are more reliable ways to detect a successful app switch, than the current `!popupWin || isWindowClosed(popupWin)` check.

mnicpt:
@bluepnume There is.  I'll post here with a change in architectural direction.

PR #80, opened 2020-10-06
maint/logging

mnicpt:
**Issue:**
Firebase connection fails if button is clicked more than once due to sessionUID being regenerated.  Firebase is expecting the first sessionUID and therefore fails with PERMISSION_DENIED due to sessionUID being different.  This can occur if the user closes or cancels the native flow and tries to go back in.  This does not result in native being closed and wondering where their props are. This change is purely a logging change to not create false alarm.

FPTI timestamps use server-side stamps.  Adding a client-side timestamp to tracking builder.


mstuart (approved):

src/api/socket.js
```diff
@@ -506,6 +510,25 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
...
+                    }
+                }
+            };
+
+            const user = firebase.auth().currentUser;
+            if (user) {
+                const uid = user.uid;
+                firebase.database().ref(`users/${ uid }/messages`).off('value', valueCallback);
+                firebase.database().ref(`users/${ uid }/messages`).remove();
+                firebase.auth().signOut();
```
bluepnume, line 529: Does this all happen synchronously, or do we need to wait for a promise here or something?

mnicpt: Remove and signOut do.  Maybe signout after removal.  Will make that change.

src/api/socket.js
```diff
@@ -506,6 +510,29 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
...
+                    }
+                }
+            };
+
+            const user = firebase.auth().currentUser;
+            if (user) {
+                const uid = user.uid;
+                firebase.database().ref(`users/${ uid }/messages`).off('value', valueCallback);
+                firebase.database().ref(`users/${ uid }/messages`).remove()
+                    .then(() => {
```
bluepnume, line 529: Could we make this a `.finally()` so it covers both cases?

src/api/socket.js
```diff
@@ -506,6 +510,29 @@ export function firebaseSocket({ sessionUID, config, sourceApp, sourceAppVersion
...
+                        handler(message);
+                    }
+                }
+            };
+
+            const user = firebase.auth().currentUser;
+            if (user) {
+                const uid = user.uid;
+                firebase.database().ref(`users/${ uid }/messages`).off('value', valueCallback);
+                firebase.database().ref(`users/${ uid }/messages`).remove()
```
bluepnume, line 528: Thinking we should save this promise, then await it on line 536? To avoid a race condition if we try to sign out and then re-auth immediately.

bluepnume:
So essentially we're making `firebaseSocket` a singleton here, right? Since it can only be used with one "user" (sessionUID) at a time.

What happens if someone:

1. Creates a firebase instance A
2. Creates a new firebase instance B with a different sessionUID
3. Tries to use the original instance A to send a message

Can we make sure that if this happens, the original instance A throws an error or something? Maybe we could validate currentUser here:

https://github.com/paypal/paypal-smart-payment-buttons/blob/14d5cb3c390a6a85aa8845b32682553581acb43d/src/api/socket.js#L570

mnicpt:
@bluepnume Pretty much. Yeah, definitely need to think about this.

src/api/socket.js
mnicpt: Yeah, good catch!  Looking.

mstuart (approved):

gregjopa (approved):
The code change looks good 👍 Seems like a low-risk logging improvement. I'm approving but wanted to note that I did not test this change locally.

src/lib/logger.js
```diff
@@ -50,7 +50,8 @@ export function setupLogger({ env, sessionID, clientID, sdkCorrelationID, locale
             [FPTI_KEY.SDK_NAME]:               FPTI_SDK_NAME.PAYMENTS_SDK,
             [FPTI_KEY.SDK_VERSION]:            sdkVersion,
             [FPTI_KEY.USER_AGENT]:             window.navigator && window.navigator.userAgent,
-            [FPTI_KEY.CONTEXT_CORRID]:         sdkCorrelationID
+            [FPTI_KEY.CONTEXT_CORRID]:         sdkCorrelationID,
+            t:                                 Date.now().toString()
```
bluepnume, line 54: Looks good, can we throw this in `FPTI_KEY` const though, so others can use the same?

mnicpt: Fo sho! ;)

bluepnume:
Looks good! One minor tweak and happy to merge this one.

src/lib/logger.js
mnicpt: https://github.com/paypal/paypal-sdk-constants/pull/42

PR #83, opened 2020-10-16
fix/v4 logging

mnicpt:


abarco (approved):

abarco:
LGMT 👍

mstuart (approved):

bluepnume:
Few things:

- This is v5 spb. Is this fix in the right repo?
- It's already defaulted here, do we need another default? https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/button/props.js#L204?
- If we need to default this value in v4, we should make this fix in xo-buttonjs repo
- Please make changes to `src/` and `test/` only in PRs. `dist/` and `cdn/` are autogenerated when running `npm run release`

mnicpt:
@bluepnume Yeah, ran the release too soon.  `setupButtonLogger` is defined in v5.  Yes, it is defaulted on props, but when xo-buttonjs calls `setupButtonLogger`, it doesn't pass `getQueriedEligibleFunding`, and so is undefined at https://github.com/paypal/paypal-smart-payment-buttons/blob/a303b22421c44ede3f104e6a77b37b857a6e493d/src/button/logger.js#L74 .

mnicpt:
@bluepnume I see what you're saying now.  What would you suggest from here?  We haven't activated the release.  If we don't activate, and make the change in xo-buttonjs, we should be good, right?

PR #86, opened 2020-11-10
Hide APM buttons when user pays with card

allen-tong:
Before:
Screen Shot 2020-11-10 at 11 17 45 AM
After:
Screen Shot 2020-11-10 at 11 20 57 AM


bluepnume:
Hmm, I think the real issue here is that the card buttons should always be at the bottom of the stack. Can you check on the ordering of the buttons?

allen-tong:
Thanks @bluepnume! I've submitted a PR to fix that in `@paypal/sdk-constants`, please check that out here https://github.com/paypal/paypal-sdk-constants/pull/39

Closing this PR :)

mnicpt:
I think @bluepnume is correct here and we shouldn't be making the change to the enum as it shouldn't dictate the rendering order.

PR #87, opened 2020-11-17
Fix memoize functionality for onApprove callback

gregjopa:
This PR passes in a reference to the `restart` function to fix memoizing the `onApprove` callback.

### Additional Info

We have been troubleshooting why memoize isn't working with the `onApprove` callback. We are thinking it's related to the `restart` function getting declared on every `onApprove` call which makes it's the arguments unique and breaks the memoization. 

We haven't been able to test this locally yet which is why I made it a draft PR.


bluepnume:
This is a great hunch. let's give it a try.

bluepnume (approved):

gregjopa:
I just pushed up a fix for the linting error. Hopefully CI passes 🤞

PR #88, opened 2020-11-18
Add support for negative testing

elizabethmv:
## What
This PR utilizes the `PayPal-Mock-Response` header (documented [here](https://developer.paypal.com/docs/api-basics/sandbox/request-headers/)) to allow merchants to pass in error codes with actions to invoke negative responses.

```
createOrder: function(data, actions) {
                return actions.order.create({
                   mockErrorCode: "AMOUNT_MISMATCH",
                    purchase_units: [{
                        amount: {
                            value: '1.00'
                        }
                    }]
                });
            }
```
```
onApprove: function(data, actions) {
                return actions.order.capture({
                    mockErrorCode: "INSTRUMENT_DECLINED"
                  }).then(function(details) {
                    alert('Transaction completed by ' + details.payer.name.given_name + '!');
                });
            }
```
## Why
Negative testing on the server with request headers is possible, but is inconvenient is not comprehensive with the current docs. Merchants want to be able to easily test out negative responses on the client.

## Steps to reproduce
1. Full local sdk setup
3. Add a valid `mockErrorCode` in the `actions.order.create()`and click on a button
2. In the network tab look for the `orders` request and confirm the `paypal-mock-response: {"mock_application_codes": "APPROPRIATE_ERROR_CODE"}` header is being passed

## Important notes
- The mocking component is not enabled in stage and is a blocker to test this locally end to end. Instead this has been tested with a combination of confirming the header is being passed correctly locally and by confirming in sandbox with a resend request in FireFox, adding in the header manually, and seeing a negative response (422).
Screen Shot 2020-11-09 at 3 30 10 PM

- The current `sync-browser-mocks` library does not support mocking by request header, and therefore is blocking adding tests.

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -1,4 +1,5 @@
 /* @flow */
+/* eslint  max-lines: off */
 
 import type { ZalgoPromise } from 'zalgo-promise/src';
 import { FPTI_KEY, FUNDING, WALLET_INSTRUMENT, INTENT } from '@paypal/sdk-constants/src';
@@ -20,7 +21,13 @@ export type OrderCreateRequest = {|
             payee? : {|
                 merchant_id? : string
             |}
-        |}>
+        |}>,
+    mockErrorCode? : string
+|};
+
+export type OrderCaptureRequest = {|
+    orderID : string,
+    mockErrorCode? : string
 |};
 
 export type OrderResponse = {||};
@@ -35,19 +42,25 @@ type OrderAPIOptions = {|
     forceRestAPI? : boolean
 |};
 
+
 export function createOrderID(order : OrderCreateRequest, { facilitatorAccessToken, partnerAttributionID } : OrderAPIOptions) : ZalgoPromise<string> {
     getLogger().info(`rest_api_create_order_id`);
+
+    const headers : Object = { [HEADERS.PARTNER_ATTRIBUTION_ID]: partnerAttributionID || '' };
+
+    if (order.mockErrorCode) {
+        headers[HEADERS.PAYPAL_MOCK_RESPONSE] = `{"mock_application_codes": "${ order.mockErrorCode }"}`;
+    }
+
+    delete order.mockErrorCode;
 
     return callRestAPI({
         accessToken: facilitatorAccessToken,
         method:      `post`,
         url:         `${ ORDERS_API_URL }`,
         data:        order,
-        headers:     {
-            [ HEADERS.PARTNER_ATTRIBUTION_ID ]: partnerAttributionID || ''
-        }
+        headers
     }).then((body) : string => {
-
         const orderID = body && body.id;
 
         if (!orderID) {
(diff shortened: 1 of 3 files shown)


src/api/order.js
```diff
@@ -83,22 +96,26 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
         });
 }
 
-export function captureOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
+export function captureOrder(order : OrderCaptureRequest, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
```
elizabethmv, line 99: I noticed that `captureOrder` defaults `forceRestAPI` to false, and seems to use `smart/api`. It looks like this route does not support negative testing. Are there cases when we do call the REST api here?

https://www.sandbox.paypal.com/v2/checkout/orders/{{order_id}}/capture https://www.sandbox.paypal.com/smart/api/order/{{order_id}}/capture

src/api/order.js
```diff
@@ -35,19 +42,25 @@ type OrderAPIOptions = {|
...
 export function createOrderID(order : OrderCreateRequest, { facilitatorAccessToken, partnerAttributionID } : OrderAPIOptions) : ZalgoPromise<string> {
     getLogger().info(`rest_api_create_order_id`);
 
+    const headers : Object = { [HEADERS.PARTNER_ATTRIBUTION_ID]: partnerAttributionID || '' };
+
+    if (order.mockErrorCode) {
+        headers[HEADERS.PAYPAL_MOCK_RESPONSE] = `{"mock_application_codes": "${ order.mockErrorCode }"}`;
+    }
+
+    delete order.mockErrorCode;
```
gregjopa, line 55: Can you move this delete operation inside the condition? Ex:

```js
    if (order.mockErrorCode) {
        headers[HEADERS.PAYPAL_MOCK_RESPONSE] = `{"mock_application_codes": "${ order.mockErrorCode }"}`;
        delete order.mockErrorCode;
    }
```

src/api/order.js
```diff
@@ -35,19 +42,25 @@ type OrderAPIOptions = {|
     forceRestAPI? : boolean
 |};
 
+
```
gregjopa, line 45: nit: we can remove this extra line.

```suggestion
```

bluepnume:
My biggest concern is, adding something new to the body for this call, that technically isn't supported at the api level. Right now the body is just passed-through to the API. So a developer knows that any of the params they can pass into `actions.order.create()` are exactly the same as the params that are passed to [api.paypal.com/v2/checkout/orders](https://developer.paypal.com/docs/api/orders/v2/#orders_create) -- no more, no less. 

Would one of the following make sense instead?

```javascript
return actions.order.create({
    purchase_units: [{
        amount: {
            value: '1.00'
        }
    }]
}, {
    mockErrorCode: "AMOUNT_MISMATCH"
});
```

or

```javascript
return actions.order({
    mockErrorCode: "AMOUNT_MISMATCH"
}).create({
    purchase_units: [{
        amount: {
            value: '1.00'
        }
    }]
});
```

Just thinking out loud. Will sleep on it some or lemme know if you have any other ideas.

src/api/order.js
gregjopa: Nice find! I didn't realize that the capture api call defaults to hitting a different endpoint (ex: https://www.sandbox.paypal.com/smart/api/order/3C284553N6060542P/capture). 

This is a tricky one. It may take some time to set up smart/api to support negative testing 🤔  

Let's chat about this more tomorrow.

gregjopa:
> a developer knows that any of the params they can pass into actions.order.create() are exactly the same as the params that are passed to api.paypal.com/v2/checkout/orders -- no more, no less.

@bluepnume that's a good point. I like your code examples. The second example with  `order` becoming a function (ex: `actions.order({})`) seems like a bigger change for implementations. On the other hand, it would be more consistent with the capture function with a pattern of 

```js
  actions.orders({
    // config for things like headers with mockErrorCode
  })
  .[apiName]({
    // payload for create/capture order api calls
  })
```

🤔

bluepnume:
The alternative is, we actually allow the api to take a mock error code in the body. That could be the most consistent approach.

elizabethmv:
@bluepnume Ah, I initially thought that adding the error code as a header conditionally and removing it from the payload resolved this, but I understand the concern more now and where it’s convenient that the payload matches exactly what is sent in the body of the api call. This makes sense.

The approach with the optional param felt inconsistent when it came to `capture()`, since it seems to not take in a body for the api call. For example:
```
return actions.order.capture({
                    mockErrorCode: "INSTRUMENT_DECLINED"
                  });
```

```
return actions.order.create({
                    purchase_units: [{
                        amount: {
                            value: '1.00'
                        }
                    }]
                }, { mockErrorCode: "AMOUNT_MISMATCH"});
```

Maybe not though? I’m certainly not opposed.

Another option could be to have them pass in the full headers either as an optional param or with the 'payload'. Since headers is more intuitive in regards to what/why it is being sent with the api call vs a key we created that is not technically supported by the api.
```
return actions.order.create({
                    headers: {
                      'paypal-mock-response': `{"mock_application_codes": "ERROR_CODE_HERE"}`
                    }
                    purchase_units: [{
                        amount: {
                            value: '1.00'
                        }
                    }]
                });
```
On the other hand, I thought it might be a little bit tedious, vs pointing them to the [error docs](https://developer.paypal.com/docs/api/orders/v2#errors) and then just having them pass in the error code the want to see.

Thoughts?

gregjopa:
Thanks for working on this @elizabethmv! I agree with closing it out for now. 

We learned a lot from hacking on this. Documenting these learnings for future reference:
- the create order api call can take a paypal-mock-response header for negative testing. This can technically be done today with a browser extension like [Requestly](https://chrome.google.com/webstore/detail/requestly-redirect-url-mo/mdnleldcmiljblolnjhpnblkcekpdkpa?hl=en) that can intercept the http request and add the paypal-mock-response header. not ideal, but technically possible.
- the smart capture api call does not currently support the paypal-mock-response header so it can't be used to test transaction failures.
- the createOrder and onApprove callbacks can reject a Promise to simulate those callbacks failing. It will trigger the onError callback to confirm the global error handler works.

At the end of the day we could tweak `actions.order.create` to support passing header values but it doesn't really provide much benefit from a negative testing perspective. Most developers want test credit cards for doing negative testing so let's explore that instead.

PR #90, opened 2020-11-20
fix/popup

mnicpt:
- Ensure unload event listener that closes popup when parent closes is removed by passing in a named function
- Ensure multiple popups cannot occur by closing the old one and opening a new one if the button is tapped again when cancelling out of native checkout flow by multi-tasking

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-354

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -20,6 +20,8 @@ import { checkout } from './checkout';
 const SOURCE_APP = 'paypal_smart_payment_buttons';
 const TARGET_APP = 'paypal_native_checkout';
 
+const NATIVE_CACHED_POPUP = '__native_popup__';
+
 const POST_MESSAGE = {
     AWAIT_REDIRECT:     'awaitRedirect',
     DETECT_APP_SWITCH:  'detectAppSwitch',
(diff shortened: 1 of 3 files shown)


src/payment-flows/native.js
```diff
@@ -248,7 +248,12 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         throw new Error(`Can not run native flow without firebase config`);
     }
 
-    const clean = cleanup();
+    if (window.clean) {
```
bluepnume, line 251: Can we use a module-level global at the top of `native.js`, i.e. just a `let clean;`

src/payment-flows/native.js
```diff
@@ -496,9 +503,11 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         const nativeUrl = getNativeUrlForAndroid({ sessionUID });
 
         const nativeWin = popup(nativeUrl);
-        window.addEventListener('unload', () => {
+
+        const closePopup = () => {
             nativeWin.close();
-        });
+        };
+        window.addEventListener('pagehide', closePopup);
```
bluepnume, line 510: Interesting. Why `pagehide` vs `unload`?

mnicpt: I found that pagehide was fired more consistently than unload when redirecting.

gregjopa:
I pulled this down locally to test. I noticed that there's one failing test:

```
TOTAL: 1 FAILED, 103 SUCCESS


1) should render a button with createOrder, click the button, and render checkout via popup to native path in Chrome with multiple clicks
     native chrome cases
     Error: Expected window to not close
```

There's a version of this test for both Android/Chrome and iOS. Only the test using the android user agent string seems to fail. How do we best resolve this failing test?

mnicpt:
@gregjopa Yeah, I'm still working on that.  It's expecting the popup to not be closed.

mnicpt:
@bluepnume @gregjopa I have fixed the tests to comply with the new changes.

bluepnume (approved):

gregjopa (approved):
Looks good to me! I tested the pagehide event on my phone using this site and it looks good: https://event-logger.glitch.me/

PR #91, opened 2020-12-01
[DO NOT MERGE]feature/sticky

mnicpt:
Provides the ability to track conversion and engagement of test user in the native flow that have successfully gone through the entire flow successfully.

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-366

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -120,6 +120,8 @@ export const FPTI_TRANSITION = {
     NATIVE_ATTEMPT_APP_SWITCH_ERRORED:  ('app_switch_attempted_errored' : 'app_switch_attempted_errored'),
     NATIVE_POPUP_SHOWN:                 ('popup_shown' : 'popup_shown'),
     NATIVE_POPUP_CLOSED:                ('process_popup_closed' : 'process_popup_closed'),
+    NATIVE_STICKY_SET:                  ('native_sticky_set' : 'native_sticky_set'),
+    NATIVE_IS_STICKY:                   ('native_is_sticky' : 'native_is_sticky'),
 
     HONEY_IDENTIFY:                     ('honey_identify')
 };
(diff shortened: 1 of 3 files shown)


bluepnume:
My concerns here are:

1. For most users this won't stick. Local storage in iframe is blocked unless you have 3rd-party-tracking enabled. Which is disabled by default.
2. We're skipping a bunch of GQL eligibility rules if the user is sticky. Just because a user was eligible for native one time, doesn't mean the next transaction will be eligible.

I don't know how to fix #1 -- but for #2, we should fix this by making sure the elmo call on the server side (which runs after all of the eligibility checks) is sticky, when possible. Normally that is done by passing a persisted id to Elmo.

So we can:

1. Try to use localStorage to keep a persistent id, else generate a new id
2. Pass that id through GQL -> Elmo, only if all eligibility checks pass
3. Rely on Elmo to provide stickiness

mnicpt:
@bluepnume Just spoke with Rohith and he brought up that this would also bias the existing experiment metrics.  Idea was to convert the current Ramp to an A/B Test so we could add another treatment so that the tests don't conflict with each other and to also make the Sticky Treatment sticky by a persistent id.  We could then get metrics around how control compares to random selection and sticky users receiving the native experience and measure conversion and engagement in all three populations.

If we can't use localStorage to persist stickiness, how would we use localStorage to persist a persistent id?  Do we have a persistent user id already being stored we can use?

bluepnume:
> allow eligiblility to trump

This won't work :) you're ignoring a "false" result from GraphQL, which could mean "this transaction is not eligible for native".

> Idea was to convert the current Ramp to an A/B Test, if possible, so we could add another treatment so that the tests don't conflict with each other and to also make the Sticky Treatment sticky by a persistent id.

Makes sense.

> If we can't use localStorage to persist stickiness, how would we use localStorage to persist a persistent id

We can't, there's no strong guarantee we can get a persistent id. The best we can do is:

1. Try to find an existing id in localStorage, if we're lucky, send it to GQL
2. If we can't, fall back to the id from the parent page (which will be specific to the domain of the parent page, not global) -- for which we'll need to add logic in the SDK to pass it down.

bluepnume:
This is the domain-specific id we can pass down from the parent page: https://github.com/paypal/paypal-sdk-client/blob/master/src/session.js#L22

mnicpt:
@bluepnume Ha!  You're right on eligibility trumping. Ope!  I think most of this code is going to go away in favor of setting up another experiment as an A/B test so we can have a second Treatment for stickiness.  Awesome on the PayPal Storage id!

mstuart:
@bluepnume are you aware that "ope" is a thing? https://www.youtube.com/watch?v=qb_-taYLRfY

bluepnume:
Hah! this is a new one to me :D

PR #93, opened 2020-12-08
feat: Add eligibility check for donate SPB

arielxin:
Added the check to restrict all items in the order to be of `DONATION` category when the `donate` label is used for SPB.

Screen Shot 2021-06-28 at 2 17 28 PM

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -8,7 +8,7 @@ import { request, noop, memoize, uniqueID } from 'belter/src';
 import { SMART_API_URI, ORDERS_API_URL, VALIDATE_PAYMENT_METHOD_API } from '../config';
 import { getLogger } from '../lib';
 import { FPTI_TRANSITION, FPTI_CONTEXT_TYPE, HEADERS, SMART_PAYMENT_BUTTONS,
-    INTEGRATION_ARTIFACT, USER_EXPERIENCE_FLOW, PRODUCT_FLOW, PREFER, LSAT_UPGRADE_FAILED } from '../constants';
+    INTEGRATION_ARTIFACT, ITEM_CATEGORY, USER_EXPERIENCE_FLOW, PRODUCT_FLOW, PREFER, LSAT_UPGRADE_FAILED } from '../constants';
 import type { ShippingMethod, ShippingAddress } from '../payment-flows/types';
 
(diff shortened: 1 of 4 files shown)


bluepnume (approved):

gregjopa:
@arielxin I'm getting a couple flow errors when running `npm run build` on your branch locally. Can you take a look?
Screen Shot 2020-12-08 at 1 59 57 PM

arielxin:
@gregjopa Thanks for catching this error. I've updated my PR with the fix. I also included `dist/` updates in a separate commit, I'm not sure if this should be part of my PR, pls let me know if I should remove it. Thanks!

bluepnume:
Yeah, please leave `dist` out, this will be autogenerated during release. Thanks!

src/api/order.js
```diff
@@ -417,7 +417,8 @@ type SupplementalOrderInfo = {|
                     currencyValue : string,
                     currencyCode : string
                 |}
-            |}
+            |},
+            category? : string
```
bluepnume, line 421: Can you use `$Values<typeof ITEM_CATEGORY>` rather than `string` for this one? Thanks!

src/button/validation.js
```diff
@@ -173,6 +174,20 @@ export function validateOrder(orderID : string, { env, clientID, merchantID, cur
...
         }
 
+        if (buttonLabel === BUTTON_LABEL.DONATE) {
+            const category = ITEM_CATEGORY.DONATION;
+            const itemCategory = cart.category || '';
+
+            if (!itemCategory || itemCategory !== category) {
+                triggerIntegrationError({
+                    error:         'smart_button_validation_error_incorrect_item_category',
+                    message:       `Expected item category from order api call to be ${ category }, got ${ itemCategory }. Please ensure you are passing ${ SDK_QUERY_KEYS.CATEGORY }=${ category } to the sdk url. https://developer.paypal.com/docs/checkout/reference/customize-sdk/`,
```
bluepnume, line 184: Looks like this is erroring -- probably need to remove `SDK_QUERY_KEYS` here, since this value is passed in order payload, not as a query param to the sdk.

arielxin: Thanks! Have updated the PR with the changes (and we will update the dev doc soon).

PR #95, opened 2020-12-10
Add badges and license

gregjopa:
This PR adds new badges similar to https://github.com/paypal/paypal-checkout-components/pull/1494. 

We were also missing our license so I added that as well.


gregjopa:
I wanted to mention that the license badge in the readme will be fixed after we publish a new version to npm.

PR #97, opened 2020-12-29
Add confirm api call

andyfen:
Retrieves the payload from sessionStorage and makes confirm payment call.

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -1,5 +1,6 @@
-/* @flow */
+/* eslint max-lines: 0 */
 
+/* @flow */
 import type { ZalgoPromise } from 'zalgo-promise/src';
 import { FPTI_KEY, FUNDING, WALLET_INSTRUMENT, INTENT } from '@paypal/sdk-constants/src';
 import { request, noop, memoize } from 'belter/src';
@@ -26,6 +27,7 @@ export type OrderCreateRequest = {|
 
 export type OrderResponse = {||};
 export type OrderCaptureResponse = {||};
+export type OrderConfirmResponse = {||};
 export type OrderGetResponse = {||};
 export type OrderAuthorizeResponse = {||};
 
@@ -182,6 +184,31 @@ export function patchOrder(orderID : string, data : PatchData, { facilitatorAcce
                 [HEADERS.CLIENT_CONTEXT]: orderID
             }
         });
+}
+
+export type ConfirmData = {|
+    payment_source : {
+        [$Values<typeof FUNDING>] : {|
+            country_code? : string | null,
+            name? : string | null,
+            email? : string | null,
+            bic? : string | null,
+            bank_id? : string | null
+        |}
+      }
+|};
+
+export function confirmOrderAPI(orderID : string, data : ConfirmData, { facilitatorAccessToken, partnerAttributionID } : OrderAPIOptions) : ZalgoPromise<OrderConfirmResponse> {
+    return callRestAPI({
+        accessToken: facilitatorAccessToken,
+        method:      `post`,
+        url:         `${ ORDERS_API_URL }/${ orderID }/confirm-payment-source`,
+        data,
+        headers:     {
+            [HEADERS.PARTNER_ATTRIBUTION_ID]: partnerAttributionID || '',
+            [HEADERS.PREFER]:                 PREFER.REPRESENTATION
+        }
+    });
 }
 
 export type ValidatePaymentMethodOptions = {|
--- a/src/button/pay.js
+++ b/src/button/pay.js
@@ -9,7 +9,6 @@ import { getLogger, promiseNoop, sendBeacon } from '../lib';
 import { FPTI_TRANSITION } from '../constants';
 import { updateButtonClientConfig } from '../api';
 import { nativeFakeoutExperiment } from '../experiments';
-import type { SmartFields } from '../types';
 
 import { type ButtonProps, type Config, type ServiceData, type Components } from './props';
 import { enableLoadingSpinner, disableLoadingSpinner } from './dom';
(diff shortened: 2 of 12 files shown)


src/button/pay.js
```diff
@@ -113,7 +111,24 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 return validateOrder(orderID, { env, clientID, merchantID, intent, currency, vault });
             });
 
-            const confirmOrderPromise = smartFields && smartFields.confirm && createOrder().then(smartFields.confirm);
+            // Check if there is a confirm order payload in local storage for the specific funding source
+            // if so use this to make a confirm call with.
+            const confirmOrderPayloadLocalStorage = localStorage.getItem(`${ fundingSource }-payload`);
```
gregjopa, line 116: It's possible that `localStorage.getItem()` can throw an error. It's best practice to wrap in a try/catch like so: https://github.com/paypal/paypal-smart-payment-buttons/blob/40058b156e67cee97e504b730c7baf77e24839ff/src/payment-flows/native.js#L108

src/button/pay.js
```diff
@@ -113,7 +111,21 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
...
+                if (confirmOrderPayloadSessionStorage) {
+                    confirmOrderPayload = JSON.parse(confirmOrderPayloadSessionStorage);
+                }
+            } catch (err) {
+                getLogger()
+                    .warn(`retrieving confirm payload from sessionStorage error`)
+                    .flush();
+            }
+ 
+            const confirmOrderPromise = confirmOrderPayload && createOrder()
```
bluepnume, line 127: We should only make this call if `confirmOrderPayload` was there in localStorage right?

src/button/pay.js
```diff
@@ -113,7 +111,21 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 return validateOrder(orderID, { env, clientID, merchantID, intent, currency, vault });
             });
 
-            const confirmOrderPromise = smartFields && smartFields.confirm && createOrder().then(smartFields.confirm);
+            let confirmOrderPayload;
+
+            try {
+                const confirmOrderPayloadSessionStorage = sessionStorage.getItem(`${ fundingSource }-payload`);
```
bluepnume, line 117: Should we clear this value once it's been read? To avoid it being used in future sessions

andyfen: using `sessionStorage` now, gets cleared when the page session ends, you feel it needs to be cleared here also ? https://github.com/paypal/paypal-smart-payment-buttons/blob/a38ebe98d858d2ff3c4e1d54c8ec4e7a0027cbb4/src/button/pay.js#L142-L144

bluepnume: Also recommend we call this something like `__confirm_${ fundingSource }_payload__` to avoid any ambiguity about what this payload is specifically for.

src/props/confirmOrder.js
```diff
@@ -0,0 +1,46 @@
...
+
+export type ConfirmOrder = (payload : ConfirmPaload) => ZalgoPromise<void>;
+
+type ConfirmOrderXProps = {|
+    orderId : string,
+    payload : ConfirmData,
+    partnerAttributionID : ?string
+|};
+
+export function getConfirmOrder({ orderId, payload, partnerAttributionID } : ConfirmOrderXProps, { facilitatorAccessToken } : {| facilitatorAccessToken : string |}) : ZalgoPromise<void> {
```
bluepnume, line 24: Minor nitpick but please use `orderID` for consistency with the rest of the repo.

src/api/order.js
```diff
@@ -149,6 +152,30 @@ export function patchOrder(orderID : string, data : PatchData, { facilitatorAcce
...
+|};
+
+export function confirmOrderAPI(orderID : string, data : ConfirmData, { facilitatorAccessToken, partnerAttributionID } : OrderAPIOptions) : ZalgoPromise<OrderConfirmResponse> {
+    return callRestAPI({
+        accessToken: facilitatorAccessToken,
+        method:      `post`,
+        url:         `${ ORDERS_API_URL }/${ orderID }/confirm-payment-source`,
+        data,
+        headers:     {
+            [HEADERS.PARTNER_ATTRIBUTION_ID]: partnerAttributionID || ''
```
bluepnume, line 174: Please also add `Prefer` header here to match the other order api calls

src/button/pay.js
andyfen: if its not in `sessionStorage` `confirmOrderPayload` will be undefined  and so will `confirmOrderPromise`

src/api/order.js
```diff
@@ -149,6 +152,30 @@ export function patchOrder(orderID : string, data : PatchData, { facilitatorAcce
         });
 }
 
+export type ConfirmData = {|
+    payment_source : {
+        [string] : {|
```
bluepnume, line 157: Rather than `[string]` can you define an enum like `PAYMENT_SOURCE_INSTRUMENT` and use `[$Values<typeof PAYMENT_SOURCE_INSTRUMENT>]`? Just so it's obvious to future callers which instruments are allowed.

src/button/pay.js
bluepnume: Ohhh I missed the `confirmOrderPayload &&`. We're good!

src/button/pay.js
bluepnume: > using sessionStorage now, gets cleared when the page session ends

True enough. I do have one concern though; what's going to happen when localstorage doesn't work or 3rd party tracking is blocked e.g. in safari?

Thinking maybe we should consider storing this in storage on the parent page?

bluepnume: FYI have added this to make it easier:

https://github.com/paypal/paypal-checkout-components/commit/049ac3b35329bf762dbec5b07c748d78914a6326

This allows for, e.g.

```javascript
sessionState.get('someKey').then(value => { ... })
```

```javascript
sessionState.set('someKey', someValue).then(() => { ... })
```

This way the data is stored on parent page and not subject to third-party-tracking restrictions.

src/api/order.js
andyfen: can you link me to this ? i don't see it

gregjopa: Hi @andyfen we recently added `Prefer` here: https://github.com/paypal/paypal-smart-payment-buttons/blob/d321c19f728e63ddf32fc77ffb77a7aeed838108/src/api/order.js#L49

src/button/pay.js
```diff
@@ -114,8 +112,35 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
...
+                                return;
+                            }
+
+                            let confirmOrderPayload;
+                            
+                            try {
+                                confirmOrderPayload = JSON.parse(confirmOrderPayloadSessionStorage);
+                            } catch (err) {
+                                getLogger()
+                                    .warn(`parsing confirm payload from sessionStorage`)
```
bluepnume, line 133: Please make cal events underscored_like_this

src/button/pay.js
```diff
@@ -114,8 +112,35 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
...
+                            if (!confirmOrderPayloadSessionStorage) {
+                                // noop when there is no confirm call to make (ie. regular flow).
+                                return;
+                            }
+
+                            let confirmOrderPayload;
+                            
+                            try {
+                                confirmOrderPayload = JSON.parse(confirmOrderPayloadSessionStorage);
+                            } catch (err) {
```
bluepnume, line 131: Do we want to re-throw the error here? If the data isn't available the buyer shouldn't be able to check out right?

test/client/happy.js
```diff
@@ -346,6 +391,15 @@ describe('happy cases', () => {
                 }
             }));
 
+            window.xprops.sessionState = {
```
bluepnume, line 394: Instead of mocking this in every test, can you just add to `mocks.js` please?

bluepnume:
Looks good! Couple of minor comments and looks like it needs a rebase, then happy to merge 👍

PR #103, opened 2021-01-20
Pass 'enableFunding' option to graphql mobileSDKEligibility query

gregjopa:
Working on adding support for the enableFunding query param

### DIFF
--- a/src/api/eligibility.js
+++ b/src/api/eligibility.js
@@ -87,7 +87,8 @@ type NativeEligibilityOptions = {|
     shippingCallbackEnabled : boolean,
     platform : $Values<typeof PLATFORM>,
     cookies : string,
-    orderID? : ?string
+    orderID? : ?string,
+    enableFunding : ?$ReadOnlyArray<$Values<typeof FUNDING>>
 |};
 
 type NativeEligibility = {|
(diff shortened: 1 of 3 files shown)


bluepnume (approved):

mnicpt (approved):

mnicpt (approved):

bluepnume:
Looks good! But if this is for venmo I think you'll also need to update native sdk eligibility call, to pass enableFunding there too

gregjopa:
> if this is for venmo I think you'll also need to update native sdk eligibility call, to pass enableFunding there too

@bluepnume great point. This is for venmo. I'll push an update for the native sdk eligibility call.

mstuart:
Lgtm! Does the funding eligibility stuff from earlier need to be there too? Or just the SDK eligibility part?

mnicpt:
> Lgtm! Does the funding eligibility stuff from earlier need to be there too? Or just the SDK eligibility part?

Was wondering the same.  Will both GQL calls be expecting `enableFunding`?

gregjopa:
> Does the funding eligibility stuff from earlier need to be there too? Or just the SDK eligibility part?

Good question. I looked up the graphql query and `fundingEligibility` does support enableFunding so it may be fine to just keep it:

Screen Shot 2021-01-20 at 4 54 18 PM

Happy to remove it as well. Let me know your thoughts.

mstuart:
Just the mobileSDKEligibility one will 👍

gregjopa:
I went ahead and reverted the other changes since they are not needed. I kept the mobileSDKEligibility change.

mstuart:
lgtm!

bluepnume:
Looks good to me! The previous changes would be good to close some tech debt if you still have them. But will merge this.

PR #106, opened 2021-02-11
feature/eligibility

mnicpt:
Checks for Android app installed to use for eligibility conditions.


server/components/native/params.js
```diff
@@ -42,7 +42,7 @@ function getParentDomain(params : NativePopupInputParams) : string {
     }
 
     // eslint-disable-next-line security/detect-unsafe-regex
-    if (!parentDomain.match(/\.paypal\.com(:\d{1,4})?$/)) {
+    if (env !== ENV.LOCAL && !parentDomain.match(/\.paypal\.com(:\d{1,4})?$/)) {
```
bluepnume, line 45: Can we rely on `NODE_ENV` here? Since `env` is passed in as a query param, an attacker could just set `env=local`

src/native/popup/popup.js
```diff
@@ -25,10 +29,62 @@ type NativePopup = {|
...
+type AndroidApp = {|
+    id? : string,
+    installed : boolean,
+    version? : string
+|};
+
+function isAndroidAppInstalled(appId : string) : ZalgoPromise<AndroidApp> {
+    // assume true unless we can prove false
+    if (window.navigator && window.navigator.getInstalledRelatedApps) {
+        return new ZalgoPromise(resolve => {
```
bluepnume, line 41: Don't need this extra promise wrapper -- can just do `return window.navigator.getInstalledRelatedApps().then(  ...`

src/native/popup/popup.js
```diff
@@ -25,10 +29,62 @@ type NativePopup = {|
...
+            window.navigator.getInstalledRelatedApps().then(result => {
+                if (result && result.length) {
+                    const apps = result.filter(app => app.id === appId);
+                    if (apps && apps.length) {
+                        const id = apps[0].id;
```
bluepnume, line 46: Can we add some logs for these? Would be good to be able to query and see which apps and versions we detect on a given device.

src/native/popup/popup.js
```diff
@@ -25,10 +29,62 @@ type NativePopup = {|
...
+                        const version = apps[0].version;
+
+                        resolve({ id, installed: true, version });
+                    } else {
+                        resolve({ installed: false });
+                    }
+                }
+                
+                resolve({ installed: true });
```
bluepnume, line 55: Assuming this is just for safety for now -- once we know this works, we can switch out for `installed: false` ya?

src/native/popup/popup.js
```diff
@@ -113,7 +169,7 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
 
     const pageUrl = `${ window.location.href  }#${  HASH.CLOSE }`;
 
-    sendToParent(MESSAGE.AWAIT_REDIRECT, { pageUrl }).then(({ redirect = true, redirectUrl }) => {
+    sendToParent(MESSAGE.AWAIT_REDIRECT, { appInstalledPromise, pageUrl }).then(({ redirect = true, redirectUrl }) => {
```
bluepnume, line 172: Can we resolve the promise first, like `appInstalledPromise.then(() => sendToParent(MESSAGE.AWAIT_REDIRECT, ... ))`? Feels a little cleaner than sending the promise in the message body (and will result in fewer back and forth messages between these windows).

src/native/popup/popup.js
mnicpt: I'm adding logs in SPB when the promise is returned.  Want them here too?

src/native/popup/popup.js
bluepnume: Or better yet -- if app not installed, should we just go straight to fallback and not even send await_redirect?

src/native/popup/popup.js
mnicpt: Right.  We will send this out ahead of Android so they can test it.  This will assume installed for now and will change after.

src/native/popup/popup.js
bluepnume: Oh right on, missed that! Let's send the logs to fpti too so they're queryable.

src/payment-flows/native.js
```diff
@@ -913,21 +913,22 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
             return ZalgoPromise.hash({
-                valid:    validatePromise,
-                eligible: eligibilityPromise
-            }).then(({ valid, eligible }) => {
-
+                valid:      validatePromise,
+                eligible:   eligibilityPromise,
+                app:        appInstalledPromise
+            }).then(({ valid, eligible, app }) => {
+                getLogger().info(`native_installed_app`, { app: { ...app } }).flush();
```
bluepnume, line 924: Can we loop over this and send a unique log name like `native_app_installed_android_4.3.3`? That will make it super easy to check in CAL and see the raw numbers without having to query.

src/native/popup/popup.js
```diff
@@ -25,10 +29,60 @@ type NativePopup = {|
...
+                    const version = apps[0].version;
+
+                    return ZalgoPromise.resolve({ id, installed: true, version });
+                } else {
+                    return ZalgoPromise.resolve({ installed: false });
+                }
+            }
+            
+            return ZalgoPromise.resolve({ installed: true });
+        });
```
bluepnume, line 55: Just thinking, let's add a .catch() block here, log the error, and default to `installed: true`.

src/payment-flows/native.js
```diff
@@ -457,6 +449,11 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     });
 
     const getNativePopupDomain = memoize(() : string => {
+        // eslint-disable-next-line no-process-env
+        if (process.env.NODE_ENV === 'development') {
```
bluepnume, line 453: Let's use the `NATIVE_POPUP_DOMAIN` constant for this. Also I think we can use `getDomain()` as the value, to avoid hard-coding the domain/port here. We can probably lose the `process.env.NODE_ENV` check too if we're relying on `getDomain()`

src/payment-flows/native.js
```diff
@@ -913,21 +918,29 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
             getLogger().info(`native_post_message_await_redirect`).flush();
 
             return ZalgoPromise.hash({
-                valid:    validatePromise,
-                eligible: eligibilityPromise
+                valid:      validatePromise,
+                eligible:   eligibilityPromise
             }).then(({ valid, eligible }) => {
-
+                Object.keys(app).forEach(key => {
```
bluepnume, line 928: Can we get a corresponding log for app not installed? That way we can easily compare ratio of installed to not-installed.

src/native/popup/popup.js
mnicpt: Added catch from `isAndroidPayPalAppInstalled` and for the Venmo one where the logger was available.

src/payment-flows/native.js
mnicpt: Good catch!

src/payment-flows/native.js
```diff
@@ -54,7 +54,7 @@ const NATIVE_DOMAIN = {
 // Popup domain needs to be different than native domain for app switch to work on iOS
 const NATIVE_POPUP_DOMAIN = {
     [ ENV.TEST ]:       'https://history.paypal.com',
-    [ ENV.LOCAL ]:      'https://history.paypal.com',
+    [ ENV.LOCAL ]:      'http://localhost:8001',
```
bluepnume, line 57: Can we do `[ ENV.LOCAL ]:     getDomain(),` so it's agnostic to whatever port you happen to be using?

bluepnume:
Still need `<link rel="manifest" href="/.well-known/manifest.webmanifest">` in the popup template. Once we have that, I think this PR is looking pretty much ready to go. Lemme know when you're good to merge.

PR #111, opened 2021-02-23
feat: Enable personalization call for Upfront Presentment

akshay-kr:
This PR enables personalization call for Upfront Presentment tagline.


server/service/personalization.jsx
```diff
@@ -127,8 +130,14 @@ function contentToJSX(content : string) : ComponentFunctionType<PersonalizationC
 }
 
 export async function resolvePersonalization(req : ExpressRequest, gqlBatch : GraphQLBatchCall, personalizationOptions : PersonalizationOptions) : Promise<Personalization> {
+    // $FlowFixMe
+    const personalizationEnabled = req.app && req.app.kraken && req.app.kraken.get('ui:enablePersonalization');
```
gregjopa, line 134: @akshay-kr Let's move this logic over to the parent app and pass this down as an argument to the mounted app.

akshay-kr: Done. Moved the logic to parent app.

akshay-kr:
@gregjopa Addressed review comment. Please review again.

server/service/personalization.jsx
```diff
@@ -137,12 +145,13 @@ export async function resolvePersonalization(req : ExpressRequest, gqlBatch : Gr
     intent = intent ? intent.toUpperCase() : intent;
     label = label ? label.toUpperCase() : label;
 
+    const taglineEnabled = tagline === true || tagline === 'true';
```
bluepnume, line 148: Can we get a 100ms timeout for this call? Want to make sure it doesn't block button render if GQL latency spikes for any reason.

akshay-kr: Done. Added timeout of 100ms.

gregjopa (approved):
Looks good!

PR #112, opened 2021-02-23
Filter cookied users

mnicpt:
Filter out cookied users with either of the following cookies from being native eligible: 
```
const AUTH_COOKIES = [
    'session_token=',
    'ulat=',
    'access_token=',
    'refresh_token='
];
```

https://github.paypal.com/Checkout/smartcomponentnodeweb/blob/develop/lib/tracking.js

### DIFF
--- a/src/native/popup/popup.js
+++ b/src/native/popup/popup.js
@@ -44,7 +44,7 @@ type AndroidApp = {|
 |};
 
 function isAlreadyAuthorizedUser() : boolean {
-    return window.document.cookie && window.document.cookie.length ? AUTH_COOKIES.some(cookie => window.document.cookie.includes(cookie)) : false;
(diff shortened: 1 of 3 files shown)


src/native/popup/popup.js
```diff
@@ -36,6 +43,10 @@ type AndroidApp = {|
     version? : string
 |};
 
+function isAlreadyAuthorizedUser() : boolean {
+    return window.document.cookie && window.document.cookie.length ? AUTH_COOKIES.some(cookie => window.document.cookie.includes(cookie)) : false;
```
gregjopa, line 47: Can we use `String.prototype.indexOf` instead of `String.prototype.includes`? IE 11 does not support string includes() and I don't think we have a polyfill for it.
```suggestion
    return window.document.cookie && window.document.cookie.length ? AUTH_COOKIES.some(cookie => window.document.cookie.indexOf(cookie) !== -1) : false;
```

mnicpt: Bummer.  No polypill.  `indexOf` it is.

PR #113, opened 2021-02-24
Update venmo universal link to v2

gregjopa:
This PR updates SPB to use the forked universal link.


gregjopa:
Closing this for now. We will revisit later with a new plan that works for iOS as well as Android.

PR #116, opened 2021-02-25
fix/shippingCallback

mnicpt:
`onShippingChange` is undefined when SPB is initialized and setup through gap.com.  We find that it is set after full load in `xprops`.  

Is there a reason why it's undefined for this merchant?
Do we want to use this fix as a short term solution until we find the root cause of why it is undefined?


bluepnume:
We should make the fix directly in v4 xo-buttonjs

PR #123, opened 2021-03-04
DTCHKINT-565 Web Add Contxt Id And Button_Session_Id To RTDB Request And Response Messages

rondwilliams:
- Add additional props to close native
https://engineering.paypalcorp.com/jira/browse/DTCHKINT-565


src/payment-flows/native.js
```diff
@@ -83,6 +83,7 @@ const PARTIAL_ENCODING_CLIENT = [
 let clean;
 let initialPageUrl;
 let nativeEligibility : NativeEligibility;
+let orderId;
```
mnicpt, line 86: We don't want to maintain this at this level.  If you need `orderId` and `buttonSessionID` for logging, then use `getSDKProps()`.  I believe it's memoized.

src/payment-flows/native.js
```diff
@@ -564,6 +565,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
 
     const getSDKProps = memoize(() : ZalgoPromise<NativeSDKProps> => {
         return createOrder().then(orderID => {
+            orderId = orderID
```
mnicpt, line 568: Please remove.

PR #127, opened 2021-03-05
DTCHKINT-565 web add contxt id and button session id to rtdb request and response messages

rondwilliams:
- Adds button session id to send message for closeNative


mnicpt (approved):
This looks great, @rondwilliams !  Thanks for your help!

PR #129, opened 2021-03-07
[Native] - iOS Back to Browser

ZkHaider:
## PR Details

This pull request adds handling in the case where we have app switched over to Native for iOS, encountered an issue and fall back to Safari Browser.

#### Mechanisms

**onFallbackListener**

We needed to define another `#` hash listener which will listen to a url containing the `#fallback` fragment. This will then reuse our existing `popupWindow` and redirect to a checkout experience in Safari browser.

```javascript
const onFallbackListener = listen(popupWin, getNativePopupDomain(), POST_MESSAGE.ON_FALLBACK, () => {
    getLogger().info(`native_message_onfallback`)
        .track({
            [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_ON_FALLBACK
        }).flush();
    fallbackToWebCheckout(popupWin);
});
```

The above is called from `popup.js` where we listen to a redirect.

**popup.js**

```javascript
case HASH.ON_FALLBACK: {
    sendToParent(MESSAGE.ON_FALLBACK);
    break;
}
```

## Testing

Tested locally with an end to end experience.

### DIFF
--- a/src/native/popup/constants.js
+++ b/src/native/popup/constants.js
@@ -5,6 +5,7 @@ export const MESSAGE = {
     DETECT_APP_SWITCH:  'detectAppSwitch',
     ON_APPROVE:         'onApprove',
     ON_CANCEL:          'onCancel',
+    ON_FALLBACK:        'onFallback',
     ON_COMPLETE:        'onComplete',
     ON_ERROR:           'onError'
 };
--- a/src/native/popup/constants.js
+++ b/src/native/popup/constants.js
@@ -24,6 +24,7 @@ export const HASH = {
     FALLBACK:           'fallback',
     ON_APPROVE:         'onApprove',
     ON_CANCEL:          'onCancel',
+    ON_FALLBACK:        'fallback',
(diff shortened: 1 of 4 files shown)


mnicpt:
@ZkHaider Where is the code to handle the hash in the popup?

ZkHaider:
> @ZkHaider Where is the code to handle the hash in the popup?

Ah right :-) copied from work laptop to here. Missing implementation in `popup.js` let me update

mnicpt:
Can we add a test to `test/client/nativePopup.js`?  Looking great, @ZkHaider

ZkHaider:
> Can we add a test to `test/client/nativePopup.js`? Looking great, @ZkHaider

Thank you for the feedback @mnicpt ! I went ahead and added a test here.

mnicpt:
@ZkHaider Looks great!  Thank you for doing this.  Awesome work!

mnicpt (approved):

PR #130, opened 2021-03-08
Update sfvc logging

mnicpt:


src/native/popup/popup.js
```diff
@@ -161,6 +158,19 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
...
     const opener = window.opener;
     if (!opener) {
+        if (isIOSSafari()) {
+            const height = window.innerHeight;
+            const scale = Math.round(window.screen.width / window.innerWidth * 100) / 100;
+            const computedHeight = Math.round(height * scale);
+            
+            const log = `${ FPTI_TRANSITION.NATIVE_POPUP_NO_OPENER }_hash_${ getRawHash() }_${ logMessage }_${ computedHeight }`;
+            logger
+                .info(log)
```
bluepnume, line 168: Let's avoid sending non-unique event names. This has gotten us in hot water in the past, notably with amplitude.

bluepnume: Can we send width and height as discrete FPTI parameters in `process_button_load` event?

bluepnume: Recommend syncing up with @ppartheeban to agree on the correct fpti params to send.

mnicpt: @bluepnume Good point on unique event names.  Regarding `process_button_load` event, won't this be inside the iframe and not give us the height and width we're looking for?  Maybe I keep the `transition_name` unique and add height to `info_msg`?

bluepnume: `outerWidth` and `outerHeight` will give the full dimensions of the page, even from inside the iframe

mnicpt: @bluepnume But we need innerHeight of the outer container to check the computedHeight which determines SFVC.

PR #132, opened 2021-03-10
remove confirm call from props

andyfen:


mnicpt (approved):
LGTM.  Would like Daniel to confirm.

PR #136, opened 2021-03-15
Add Message Data For Firebase Callbacks

rondwilliams:
Passed in data for socket.on callbacks

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -626,14 +626,14 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
             onCancel(),
             close()
         ]).then(() => {
-            return { buttonSessionID };
+            return {buttonSessionID};
         });
     };
 
-    const onErrorCallback = ({ data : { message } }: {| data : {| message : string |} |}) => {
+    const onErrorCallback = ({ data : { message }} : {| data : {| message: string |} |}) => {
         getLogger().info(`native_message_onerror`, {err: message})
             .track({
-                [FPTI_KEY.TRANSITION]:  FPTI_TRANSITION.NATIVE_ON_ERROR,
+                [FPTI_KEY.TRANSITION]: FPTI_TRANSITION.NATIVE_ON_ERROR,
                 [FPTI_CUSTOM_KEY.INFO_MSG]: `Error message: ${message}`
             }).flush();
         return ZalgoPromise.all([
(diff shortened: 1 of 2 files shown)


src/payment-flows/native.js
```diff
@@ -705,16 +705,16 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
             });
         });
 
-        const getPropsListener = socket.on(SOCKET_MESSAGE.GET_PROPS, () : ZalgoPromise<NativeSDKProps> => {
+        const getPropsListener = socket.on(SOCKET_MESSAGE.GET_PROPS, { buttonSessionID }, () : ZalgoPromise<NativeSDKProps> => {
```
mnicpt, line 708: You should be getting a lint error here since we don't use `buttonSessionID`.  This is a listener, it's not sending a message.

src/payment-flows/native.js
```diff
@@ -705,16 +705,16 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
             getLogger().info(`native_message_getprops`).flush();
             return getSDKProps();
         });
 
-        const onShippingChangeListener = socket.on(SOCKET_MESSAGE.ON_SHIPPING_CHANGE, onShippingChangeCallback);
-        const onApproveListener = socket.on(SOCKET_MESSAGE.ON_APPROVE, onApproveCallback);
-        const onCancelListener = socket.on(SOCKET_MESSAGE.ON_CANCEL, onCancelCallback);
-        const onErrorListener = socket.on(SOCKET_MESSAGE.ON_ERROR, onErrorCallback);
-        const onFallbackListener = socket.on(SOCKET_MESSAGE.ON_FALLBACK, onFallbackCallback);
+        const onShippingChangeListener = socket.on(SOCKET_MESSAGE.ON_SHIPPING_CHANGE, { buttonSessionID }, onShippingChangeCallback);
```
mnicpt, line 713: Same with all of these.  These are listeners, not sending messages.

mnicpt:
@rondwilliams Why is this needed?  Can you provide an use case?

rondwilliams:
@mnicpt From my understanding, it's required that we send the `buttonSessionID` with every Firebase event. From what I understand, this allows us to track transactions across RTDB FPTI events. If there is a better way of doing this, I am all ears.

bluepnume (approved):

PR #137, opened 2021-03-17
Added elmo for branded standalone button

JianyiGao:


bluepnume:
Let's use the existing `branded` prop here too please.

server/components/buttons/middleware.js
```diff
@@ -42,13 +46,14 @@ type ButtonMiddlewareOptions = {|
     },
     tracking : (ExpressRequest) => void,
     getPersonalizationEnabled : (ExpressRequest) => boolean,
-    cdn? : boolean
+    cdn? : boolean,
+    getApmStandaloneButtonExperiment? : (req : ExpressRequest, params : ApmStandaloneButtonElmoParam) => Promise<boolean>
```
mnicpt, line 50: @bluepnume This seems really specific in this area.  Is there a way to abstract this out into it's own `personalization` middleware?

JianyiGao: I see `getInlineGuestExperiment` is also in this middleware, this change is quite similar to that so I try to keep it in the same place. Let me know if it makes more sense to move it out

bluepnume: Yes I would like to make this a generic `isFundingSourceBranded` function instead. See https://github.paypal.com/Checkout/smartcomponentnodeweb/pull/64#discussion_r2910090

Would rather not tightly couple this with APMs, the same principle should be applicable to any button.

server/components/buttons/middleware.js
```diff
@@ -27,6 +27,10 @@ type InlineGuestElmoParams = {|
     buyerCountry : $Values<typeof COUNTRY>
 |};
 
+type BrandedFundingSourceElmoParam = {|
+    clientID : string
```
bluepnume, line 31: Please pass `fundingSource` too, so this function can make a decision based on the rendered button.

server/components/buttons/middleware.js
```diff
@@ -121,7 +126,8 @@ export function getButtonMiddleware({
             const isCardFieldsExperimentEnabled = await isCardFieldsExperimentEnabledPromise;
             const wallet = await walletPromise;
             const personalization = await personalizationPromise;
-
+            const brandedFundingSource = await isFundingSourceBranded(req, { clientID });
```
bluepnume, line 129:
```
const brandedFundingSource = fundingSource
    ? await isFundingSourceBranded(req, { clientID, fundingSource })
    : true;
```

src/button/props.js
```diff
@@ -228,6 +228,7 @@ export function getProps({ facilitatorAccessToken } : {| facilitatorAccessToken
 
     enableFunding = enableFunding || [];
     disableFunding = disableFunding || [];
+    branded = branded || brandedFundingSource || false;
```
bluepnume, line 231: I think this should be `branded = branded ?? brandedFundingSource`.

- Don't want to do `||` since this will trigger if branded is passed as false
- Don't want to default to `false` since paypal/venmo etc are branded by default.

server/components/buttons/middleware.js
```diff
@@ -146,7 +152,8 @@ export function getButtonMiddleware({
 
             const setupParams = {
                 fundingEligibility, buyerCountry, cspNonce, merchantID, sdkMeta, wallet, correlationID,
-                firebaseConfig, facilitatorAccessToken, eligibility, content, cookies, personalization
+                firebaseConfig, facilitatorAccessToken, eligibility, content, cookies, personalization,
+                brandedFundingSource
```
bluepnume, line 156: Let's call this `brandedDefault`, since this is the default value for `branded` which can be overridden by `props.branded`

src/button/button.js
```diff
@@ -97,6 +99,11 @@ export function setupButton(opts : ButtonOpts) : ZalgoPromise<void> {
                     }
                     return;
                 }
+            } else {
+                if (!brandedFundingSource) {
```
bluepnume, line 103: Let's use `props.branded` here so we get the final derived flag.

src/button/props.js
```diff
@@ -285,7 +286,7 @@ export function getProps({ facilitatorAccessToken } : {| facilitatorAccessToken
     const createOrder = getCreateOrder({ createOrder: xprops.createOrder, currency, intent, merchantID, partnerAttributionID }, { facilitatorAccessToken, createBillingAgreement, createSubscription });
 
     const onError = getOnError({ onError: xprops.onError });
-    const onApprove = getOnApprove({ onApprove: xprops.onApprove, intent, onError, partnerAttributionID, upgradeLSAT, clientAccessToken, vault }, { facilitatorAccessToken, createOrder });
+    const onApprove = getOnApprove({ onApprove: xprops.onApprove, intent, onError, partnerAttributionID, upgradeLSAT, clientAccessToken, vault }, { facilitatorAccessToken, brandedFundingSource, createOrder });
```
bluepnume, line 289: Let's use `props.branded` here so we get the final derived flag.

src/props/onApprove.js
```diff
@@ -263,9 +263,8 @@ export function getOnApprove({ intent, onApprove = getDefaultOnApprove(intent),
                 }).flush();
 
             if (!billingToken && !subscriptionID && !clientAccessToken && !vault) {
-                if (!payerID) {
-                    getLogger().error('onapprove_payerid_not_present', { orderID }).flush();
-                    // throw new Error(`payerID not present in onApprove call`);
+                if (!payerID && brandedFundingSource) {
```
bluepnume, line 266: Let's use `props.branded` here so we get the final derived flag.

src/button/button.js
```diff
@@ -97,6 +99,11 @@ export function setupButton(opts : ButtonOpts) : ZalgoPromise<void> {
                     }
                     return;
                 }
+            } else {
+                if (!brandedDefault) {
```
bluepnume, line 103: Let's use `props.branded` here please

src/props/onApprove.js
```diff
@@ -243,7 +243,7 @@ type OnApproveXProps = {|
     vault : boolean
 |};
 
-export function getOnApprove({ intent, onApprove = getDefaultOnApprove(intent), partnerAttributionID, onError, clientAccessToken, vault, upgradeLSAT = false } : OnApproveXProps, { facilitatorAccessToken, createOrder } : {| facilitatorAccessToken : string, createOrder : CreateOrder |}) : OnApprove {
+export function getOnApprove({ intent, onApprove = getDefaultOnApprove(intent), partnerAttributionID, onError, clientAccessToken, vault, upgradeLSAT = false } : OnApproveXProps, { facilitatorAccessToken, branded, createOrder } : {| facilitatorAccessToken : string, branded? : boolean, createOrder : CreateOrder |}) : OnApprove {
```
bluepnume, line 246: Let's make this non-optional type `branded : boolean`

src/types.js
```diff
@@ -58,7 +58,8 @@ export type CheckoutProps = {|
     clientMetadataID : ?string,
     enableFunding : ?$ReadOnlyArray<$Values<typeof FUNDING>>,
     standaloneFundingSource : ?$Values<typeof FUNDING>,
-    amplitude? : boolean
+    amplitude? : boolean,
+    branded? : boolean
```
bluepnume, line 62: Let's make this non-optional type `branded : boolean`

JianyiGao:
Not sure why the builds are not being triggered?

server/components/buttons/middleware.js
```diff
@@ -42,13 +47,14 @@ type ButtonMiddlewareOptions = {|
     },
     tracking : (ExpressRequest) => void,
     getPersonalizationEnabled : (ExpressRequest) => boolean,
-    cdn? : boolean
+    cdn? : boolean,
+    isFundingSourceBranded? : (req : ExpressRequest, params : BrandedFundingSourceElmoParam) => Promise<boolean>
```
bluepnume, line 51: I think this should not be an optional type, right? `isFundingSourceBranded` will always be passed. Then we can also remove the default `() => Promise.resolve(false)`.

server/components/buttons/params.js
```diff
@@ -43,7 +43,8 @@ type ButtonInputParams = {|
     riskData? : string,
     platform : ?$Values<typeof PLATFORM>,
     paymentMethodNonce? : ?string,
-    branded? : boolean
+    branded? : boolean,
+    fundingSource : string
```
bluepnume, line 47: Please use `$Values<typeof FUNDING>`

server/components/buttons/params.js
```diff
@@ -81,7 +82,8 @@ type ButtonParams = {|
     platform : $Values<typeof PLATFORM>,
     cookies : string,
     paymentMethodNonce : ?string,
-    branded : ?boolean
+    branded : ?boolean,
+    fundingSource : string
```
bluepnume, line 86: Please use `$Values<typeof FUNDING>`

bluepnume:
Not sure why the job isn't running. Feel free to either open a new PR, or let me know if `npm t` passes locally.

server/components/buttons/middleware.js
JianyiGao: I think i cant remove the default `() => Promise.resolve(false)`, or `isFundingSourceBranded` will need to be added to server/devServer as a required param
https://github.com/paypal/paypal-smart-payment-buttons/blob/master/server/devServer.js#L254

JianyiGao:
Tests pass
Screen Shot 2021-04-02 at 5 16 50 PM
 in local, builds seem to fail because of some new changes.

PR #138, opened 2021-03-17
Add response on approve message response

rondwilliams:
- This PR returns a response for the new `response_onApprove` message coming from Firebase


bluepnume:
Hey Rondale! 

Why exactly is this needed? Native sends onApprove to SPB, SPB sends a response to Native, why do we need a response to the response?

rondwilliams:
@bluepnume I'm getting clarification on this right now give me one second.

PR #139, opened 2021-03-18
format error details message

andyfen:
throws the error details message if there is an error on an api call.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -34,7 +34,7 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } : 
         json:    data
     }).then(({ status, body, headers: responseHeaders }) : T => {
         if (status >= 300) {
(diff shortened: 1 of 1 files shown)


src/api/api.js
```diff
@@ -34,6 +34,12 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
         json:    data
     }).then(({ status, body, headers: responseHeaders }) : T => {
         if (status >= 300) {
+            if (body.details && Array.isArray(body.details)) {
+                body.details.forEach(({ issue, description }) => {
+                    // eslint-disable-next-line no-console
+                    console.log(`%cERROR \n%c${ issue }: %c${ description }`, 'color: #d93025; font-size: 16px; font-weight: bold', 'color: inherit; font-weight: bold',  'color: inherit');
```
bluepnume, line 40: Instead of console.log can you include these details in the error message when doing `throw new Error(message)`? This will enable merchants to capture the error and log what's going wrong.

andyfen: can we do both ? the idea is to make it more readable than debug the error messages

PR #140, opened 2021-03-23
update error details logging

andyfen:


src/api/api.js
```diff
@@ -34,7 +34,7 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
         json:    data
     }).then(({ status, body, headers: responseHeaders }) : T => {
         if (status >= 300) {
-            throw new Error(`${ body.detail[0].issue }: ${ body.detail[0].description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
+            throw new Error(`${ body.details[0].issue }: ${ body.details[0].description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
```
bluepnume, line 37: Actually could you null check `body.details && body.details[0]`, else log a generic message? Some REST apis may not follow this response format.

src/api/api.js
```diff
@@ -34,7 +34,7 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
         json:    data
     }).then(({ status, body, headers: responseHeaders }) : T => {
         if (status >= 300) {
-            throw new Error(`${ body.detail[0].issue }: ${ body.detail[0].description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
+            throw new Error(`${ body.details && body.details[0].issue }: ${ body.details && body.details[0].description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
```
mnicpt, line 37: @andyfen If `body.details` is an empty array, `body.details[0]` will be undefined.

mnicpt: I would pull the above condition out of the Error and then have a generic Error message if it's not available.

andyfen: how's that?

mnicpt: @andyfen Because when you check `body.details` with an empty array, it returns true and then `body.details[0]` is undefined causing the access to `issue` to error. https://codepen.io/mnicpt/pen/KKNYjex

src/api/api.js
```diff
@@ -34,7 +34,8 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
         json:    data
     }).then(({ status, body, headers: responseHeaders }) : T => {
         if (status >= 300) {
-            throw new Error(`${ body.detail[0].issue }: ${ body.detail[0].description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
+            const hasDetails = Boolean(body.details) && Boolean(body.details.length);
```
mnicpt, line 37: Don't need to construct Boolean's from these but not a blocker.  Could just have `body.details && body.details.length` and you would be fine.  LGTM.

PR #105, opened 2021-02-10
Providing alternative lookup for intent

sinhanurag:
This PR adds an alternative lookup for the initiation intent passed by merchants that have integrated with the Orders.V2 APIs for an AS2 transaction.

PR #143, opened 2021-03-25
Provide alternative lookup for intent v2

gregjopa:
This PR adds back the alternative lookup for intent and fixes the bug with the graphql query by moving it under cart.

Previous PR: https://github.com/paypal/paypal-smart-payment-buttons/pull/105

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -130,7 +130,7 @@ export function authorizeOrder(orderID : string, { facilitatorAccessToken, buyer
 }
 
 type PatchData = {|
-    
+
 |};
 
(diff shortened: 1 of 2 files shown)


nbierdeman (approved):
Nice work! 🍻

src/api/order.js
```diff
@@ -487,6 +490,9 @@ export const getSupplementalOrderInfo : GetSupplementalOrderInfo = memoize(order
                                 currencyValue
                             }
                         }
+                        supplementary {
```
gregjopa, line 493: Previously `supplementary` was not nested under `cart`. This PR fixes that issue.

PR #147, opened 2021-03-30
[WIP] Honey checkout

amyegan:
* Include `currency` in exports
* Include `upgradeFacilitatorAccessToken` in exports
* Pass `userExperienceFlow` from url to checkout flow client config

### DIFF
--- a/src/button/exports.js
+++ b/src/button/exports.js
@@ -11,11 +11,10 @@ import type { ButtonProps } from './props';
 
 type ExportsProps = {|
     props : ButtonProps,
-    isEnabled : () => boolean,
-    facilitatorAccessToken : string
+    isEnabled : () => boolean
 |};
 
-export function setupExports({ props, isEnabled, facilitatorAccessToken } : ExportsProps)  {
+export function setupExports({ props, isEnabled } : ExportsProps)  {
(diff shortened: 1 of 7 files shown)


src/button/exports.js
```diff
@@ -62,7 +64,10 @@ export function setupExports({ props, isEnabled } : ExportsProps)  {
                     return onApprove(data, actions);
                 },
                 onCancel,
-                onError
+                onError,
+                upgradeFacilitatorAccessToken: (buyerAccessToken, orderID) => {
```
bluepnume, line 68: One quick thing -- this function will need to take `facilitatorAccessToken` and `orderID` (both from the merchant spb). We can't expose `buyerAccessToken` outside of this iframe, so we'll need to store it somewhere in memory until this function is called.

bluepnume: We can store it in any variable in memory in the iframe; maybe we can spin up some new helpers in `session.js`, something like:

```javascript
let buyerAccessToken : ?string;

export function getBuyerAccessToken() : ?string {
    return buyerAccessToken;
}

export function storeBuyerAccessToken(val : string) {
    buyerAccessToken = val;
}
```

(then invoke these from nonce.js and exports.js when needed)

PR #149, opened 2021-04-05
Fix firebase issue

mnicpt:
Ensures that the popup window is not closed until `onApprove` is called.  When the popup window closes prior to the `onApprove` message being received from native, the wrong tab is opened that is not the merchants page.

### DIFF
--- a/src/native/popup/popup.js
+++ b/src/native/popup/popup.js
@@ -256,7 +256,7 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
             break;
         }
         case HASH.CLOSE: {
-            sendToParent(MESSAGE.ON_COMPLETE);
+            sendToParent(MESSAGE.ON_COMPLETE).finally(closeWindow);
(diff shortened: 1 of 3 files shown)


src/native/popup/popup.js
```diff
@@ -282,7 +282,7 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
     appInstalledPromise.then(app => {
         sendToParent(MESSAGE.AWAIT_REDIRECT, { app, pageUrl, sfvc, stickinessID }).then(({ redirect = true, redirectUrl, appSwitch = true }) => {
             if (!redirect) {
-                return;
+                closeWindow();
```
bluepnume, line 285: Can't close here. This means parent wants to use the popup to show fallback experience.

bluepnume:
LGTM other than that one line!

bluepnume (approved):

gregjopa (approved):

PR #151, opened 2021-04-19
Fix a11y iframe landmark

elizabethmv:
This PR adds `role="main` to the buttons container `<div>` to resolve an axe-core error being thrown for html landmarks. It also adds an `aria-label` so that the landmarks distinguishable from the parent.

Screen Shot 2021-04-19 at 2 41 38 PM


mnicpt:
LGTM!  I've re-run the builds to see if they fix themselves.

gregjopa (approved):

PR #156, opened 2021-04-29
Add channel param to universal/app link for Venmo Desktop

nbierdeman:
### Purpose

This PR adds the new `channel` param to the universal/app link. The two possible values for `channel` are `mobile-web` and `desktop-web`.

### DIFF
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -86,8 +86,6 @@ const CHANNEL = {
     MOBILE:  'mobile-web'
 };
 
-const channel = isDevice() ? CHANNEL.MOBILE : CHANNEL.DESKTOP;
-
 let clean;
 let initialPageUrl;
 let nativeEligibility : NativeEligibility;
@@ -442,6 +440,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
 
     const shippingCallbackEnabled = Boolean(onShippingChange);
     sdkMeta = sdkMeta.replace(/[=]+$/, '');
+    const channel = isDevice() ? CHANNEL.MOBILE : CHANNEL.DESKTOP;
 
     if (!firebaseConfig) {
         throw new Error(`Can not run native flow without firebase config`);
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -432,7 +432,8 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     let { facilitatorAccessToken, sdkMeta, buyerCountry, merchantID, cookies } = serviceData;
     const { fundingSource } = payment;
     const { sdkVersion, firebase: firebaseConfig } = config;
-
+    const channel : string = isDevice() ? 'mobile-web' : 'desktop-web';
+    
     const shippingCallbackEnabled = Boolean(onShippingChange);
     sdkMeta = sdkMeta.replace(/[=]+$/, '');
 
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -2,7 +2,7 @@
 /* eslint max-lines: off, max-nested-callbacks: off */
 
 import { extendUrl, uniqueID, getUserAgent, supportsPopups, memoize, stringifyError,
-    stringifyErrorMessage, cleanup, once, noop, inlineMemoize } from 'belter/src';
+    stringifyErrorMessage, cleanup, once, noop, inlineMemoize, isDevice } from 'belter/src';
 import { ZalgoPromise } from 'zalgo-promise/src';
 import { PLATFORM, ENV, FPTI_KEY, FUNDING } from '@paypal/sdk-constants/src';
 import { type CrossDomainWindowType, isWindowClosed, onCloseWindow, getDomain } from 'cross-domain-utils/src';
--- a/src/payment-flows/native.js
+++ b/src/payment-flows/native.js
@@ -504,7 +504,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const getDirectNativeUrl = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
         return conditionalExtendUrl(`${ getNativeDomain() }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
             query: {
-                sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
+                ...(fundingSource === 'venmo' ? { channel } : {}), sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
                 enableFunding:  enableFunding.join(','),
                 domain:         merchantDomain,
                 rtdbInstanceID: firebaseConfig.databaseURL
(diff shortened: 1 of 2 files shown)


mstuart:
Looks great! :D

src/payment-flows/native.js
```diff
@@ -503,7 +504,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const getDirectNativeUrl = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
         return conditionalExtendUrl(`${ getNativeDomain() }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
             query: {
-                sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
+                ...(fundingSource === 'venmo' ? { channel } : {}), sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
```
mnicpt, line 507: Great work so far, @nbierdeman !  Can we refactor this expression to be on line 435 like this and just add `...channel` here and in the `getDelayedNativeUrlQueryParams`?  
```
const channel = fundingSource === FUNDING.VENMO ? { channel: isDevice() ? 'mobile-web' : 'desktop-web'  } : {};
```
If you want to split up the line above that's fine, just was looking to store the repeated expression.

src/payment-flows/native.js
```diff
@@ -432,7 +432,8 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     let { facilitatorAccessToken, sdkMeta, buyerCountry, merchantID, cookies } = serviceData;
     const { fundingSource } = payment;
     const { sdkVersion, firebase: firebaseConfig } = config;
-
+    const channel : string = isDevice() ? 'mobile-web' : 'desktop-web';
```
gregjopa, line 435: Looking good! We also have `xprops.supportedNativeBrowser` which returns a boolean value which indicates if the mobile device browser can app switch or not. Should we use this instead of `isDevice()`? For example, currently Chrome on iOS can not app switch so what should happen when the buyer clicks Venmo in that use case? Should it fallback to this desktop experience?

mnicpt: @gregjopa @nbierdeman Actually, if we have gotten into the `init` method, then we are on device.  `initNative` will only be called if the payment flow is eligible which means `supportedNativeBrowser` is true.  So no condition needed here.  Only `mobile-web`.

mnicpt: I'll update my code example above.

gregjopa: Ahh great point @mnicpt! This native flow only gets triggered when it's a native eligible browser so it's always `mobile-web`. Makes sense now 👍

nbierdeman: Thanks for the feedback! I'll submit some revisions soon.

src/payment-flows/native.js
```diff
@@ -432,7 +432,8 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     let { facilitatorAccessToken, sdkMeta, buyerCountry, merchantID, cookies } = serviceData;
     const { fundingSource } = payment;
     const { sdkVersion, firebase: firebaseConfig } = config;
-
+    const channel = fundingSource === FUNDING.VENMO ? { channel: 'mobile-web' } : {};
```
bluepnume, line 435: Why limit this to venmo? Can we not just make this a generic param for all FIs for both mobile-web and desktop-web?

Can we also define a `CHANNEL` enum please?

src/payment-flows/native.js
```diff
@@ -503,7 +504,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const getDirectNativeUrl = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
         return conditionalExtendUrl(`${ getNativeDomain() }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
             query: {
-                sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
+                ...channel, sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
```
bluepnume, line 507: Let's just pass the real value rather than doing a spread.

src/payment-flows/native.js
```diff
@@ -517,6 +518,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         const forceEligible = isNativeOptedIn({ props });
 
         return {
+            ...channel,
```
bluepnume, line 521: Same

src/payment-flows/native.js
nbierdeman: Thanks for the feedback, @bluepnume!

> Why limit this to venmo? Can we not just make this a generic param for all FIs for both mobile-web and desktop-web?

I was wondering the same thing. I'll go ahead and make `channel` a generic param for all FI's.

> Can we also define a CHANNEL enum please?

Yes, will do.

src/payment-flows/native.js
nbierdeman: @mnicpt, since `initNative` is only called when we are on a device and `supportedNativeBrowser` is `true`, when and where will `channel=desktop-web` get passed to Venmo?

src/payment-flows/native.js
```diff
@@ -81,6 +81,13 @@ const PARTIAL_ENCODING_CLIENT = [
     'AeG7a0wQ2s97hNLb6yWzDqYTsuD-4AaxDHjz4I2EWMKN6vktKYqKJhtGqmH2cNj_JyjHR4Xj9Jt6ORHs'
 ];
 
+const CHANNEL = {
+    DESKTOP: 'desktop-web',
+    MOBILE:  'mobile-web'
+};
+
+const channel = isDevice() ? CHANNEL.MOBILE : CHANNEL.DESKTOP;
```
bluepnume, line 89: Let's do this inside one of the functions. Generally helps with tree-shaking if we don't do function calls at the top level of the module.

src/payment-flows/native.js
```diff
@@ -503,7 +510,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const getDirectNativeUrl = memoize(({ pageUrl = initialPageUrl, sessionUID } = {}) : string => {
         return conditionalExtendUrl(`${ getNativeDomain() }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
             query: {
-                sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
+                channel, sdkMeta, fundingSource, sessionUID, buttonSessionID, pageUrl, clientID, stickinessID:   defaultStickinessID,
```
bluepnume, line 513: Looks awesome. Can we get a test for iOS/Android to make sure this param is sent correctly?

bluepnume: FWIW, it may be useful to define this `query` object as a flow type at some point also, so we can explicitly call out mandatory and optional params and avoid this contract breaking in future.

src/payment-flows/native.js
nbierdeman: @bluepnume I moved `channel` inside of `initNative`. At the moment, it will never be `desktop-web`. I've let @mstuart and @BriceShatzer know this as well.

src/payment-flows/native.js
mnicpt: @nbierdeman In checkout flow on browser.

src/payment-flows/native.js
mnicpt: So we really just need to set the channel and not even store in a variable: `{ channel: 'mobile-web' }`

PR #158, opened 2021-05-04
adds QRCode UI component

BriceShatzer:
Adds a component to render a modal with a QR Code that will be used in [Venmo Desktop Pay project](https://engineering.paypalcorp.com/confluence/pages/viewpage.action?pageId=585593422)

**Testing:**
```
npm start 

// -> http://localhost.paypal.com:8003/smart/qrcode?demo=true&qrPath=string_to_be_encoded_in_qr_code

```

**Designs:** [Figma](https://www.figma.com/file/g9Pg0vQJK0EVeiBLiuMoC2/Universal-Venmo-Pay?node-id=0%3A1)


server/components/qrcode/middleware.js
```diff
@@ -0,0 +1,62 @@
...
+                    href="https://www.paypalobjects.com/paypal-ui/web/fonts-and-normalize/1-1-0/fonts-and-normalize.min.css"
+                />
+            </head>
+            <body data-nonce="${ cspNonce }" data-client-version="${ client.version }">
+                ${ meta.getSDKLoader({ nonce: cspNonce }) }
+                <script nonce="${ cspNonce }">${ client.script }</script>
+                <script nonce="${ cspNonce }">
+        spb.renderQRCode(${ safeJSON({
+        cspNonce,
+        qrPath,
```
bluepnume, line 51: If we're doing client side rendering, do we need to pass this `qrPath` via the server? Can't we just use `window.xprops.qrPath`?

server/components/qrcode/middleware.js
```diff
@@ -0,0 +1,62 @@
...
+        demo
```
bluepnume, line 52: Same question

server/components/qrcode/middleware.js
```diff
@@ -0,0 +1,62 @@
...
+    return sdkMiddleware({ logger, cache }, {
+        app: async ({ req, res, params, meta, logBuffer }) => {
+            logger.info(req, EVENT.RENDER);
+
+
+            const { cspNonce, qrPath, demo, debug } = getParams(params, req, res);
+
+            const client = await getSmartQRCodeClientScript({ debug, logBuffer, cache, useLocal });
+
+            logger.info(req, `menu_client_version_${ client.version }`);
```
bluepnume, line 30: s/menu/qrcode

src/qrcode/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+```
+// from something like this:
+var QRCode = function(t){"use strict";var r,e=function(){return"function"=...
+
+// to export a const like this:
+export const QRCode = function(t){"use strict";var r,e=function(){return"function"=...
+```
+
+4. Copy the modified contents of `build/qrcode.js` into the `/src/qrcode/node-qrcode.js` file
```
bluepnume, line 21: Weird. If this is forked, can we just publish `node-qrcode-braintree` or something, so we can just pull this in from npm rather than following these steps?

src/qrcode/components.jsx
```diff
@@ -0,0 +1,324 @@
...
+import { html } from 'jsx-pragmatic';
+import { h, Fragment, Node } from 'preact';
+import { VenmoLogo, LOGO_COLOR } from '@paypal/sdk-logos/src';
+
+export type NodeType = typeof Node;
+
+export function Logo() : NodeType {
+    return (<span innerHTML={ `
+        ${ VenmoLogo({ logoColor: LOGO_COLOR.DEFAULT }).render(html()) }
+    ` } />);
```
bluepnume, line 13: Can we just do:

```
return <span>
  <VenmoLogo />
</span>
```

This should work the same way but be cleaner than setting innerHTML

bluepnume: Aaaaah because it's preact, got it. In that case you could do:

```
return VenmoLogo().render(preact({ Preact: { h } }))
```

src/qrcode/qrcard.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+type QRPath = string;
+type QRCardProps = {|
+    cspNonce : ?string,
+    qrPath : QRPath,
+    demo : boolean
+|};
+
+class QRCodeElement extends Component<{|qrPath : QRPath|}, {|dataURL : string|}> {
+    async componentDidMount() {
```
bluepnume, line 20: Would it make sense to use hooks for consistency with other components?

webpack.config.js
```diff
@@ -92,6 +92,30 @@ export const WEBPACK_CONFIG_MENU_DEBUG : WebpackConfig = getSmartWebpackConfig({
     minify:     false,
     vars:       globals
 });
+export const WEBPACK_CONFIG_QRCODE : WebpackConfig = getSmartWebpackConfig({
+    modulename: 'spb',
```
bluepnume, line 96: Can we do `spbQRCode`

src/qrcode/qrcard.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+function ErrorMessage({
+    message,
+    resetFunc
+} : {|
+    message? : string,
+    resetFunc : () => void
+|}) : NodeType {
+    return (
+        <div id="venmo-error-view">
```
bluepnume, line 51: Can we avoid hard-coding `venmo` in any of the ids, class names, etc? Would like to make this easy to adapt to other use cases as we go forward.

src/qrcode/components.jsx
BriceShatzer: Yeah I wasn't super familiar w/ getting the preact & jsx-pragmatic stuff to play nice together, and when I got that to work I figured ¯\_(ツ)_/¯

server/components/qrcode/middleware.js
BriceShatzer: It's been expressed that there is a desire to have a sort of "test page" that the other teams can work against. Hitting the route & including a path to be encoded into the QR Code element seemed to make the most sense to me. Is there a way to do this that is better/makes more sense?

src/qrcode/README.md
BriceShatzer: What we're doing here is basically duplicating what [braintree is doing](https://github.com/braintree/braintree-web/tree/master/src/venmo/internal/vendor) in their integration. 

I think the most desirable solution would be to get braintree's [changes](https://github.com/braintree/node-qrcode) to the [original](https://github.com/soldair/node-qrcode) library merged in, followed by publishing `node-qrcode-braintree` like you mentioned. 

You're not wrong, but at this point I feel like getting this sorted as a published module could turn into a time-sink with very little upside.

src/qrcode/qrcard.jsx
BriceShatzer: I've messed around w/ trying to get this to work in a rational way for far longer than I'd care to admit lol

_I think_ the main issue has to do with the fact that it [seems like](https://github.com/soldair/node-qrcode/issues/175) the `toDataURL` method is reliant on the browser's canvas api and is implemented in a way that is asynchronous and needing the async/await behavior w/ `componentDidMount()`, which I couldn't figure out how to do w/ a hook? I guess? lol 

Honestly, I tried a bunch of different ways to implement this component and this was the only way I could get it to work. I'm certainly open to ideas thought.

src/qrcode/README.md
bluepnume: Fair. Second question though -- would it make sense to include this on the server instead and spitting out html, to avoid loading a new library on the client every time?

src/qrcode/qrcard.jsx
bluepnume: Fair. Not a huge deal, if this works I'm not opposed to it.

src/qrcode/qrcard.jsx
```diff
@@ -0,0 +1,111 @@
...
+            <style nonce={ cspNonce }> { cardStyle } </style>
+            <div id="view-boxes" className={ processState }>
+                { isError() ?
+                    <ErrorMessage message={ errorMessage } resetFunc={ () => setState_default() } /> :
+                    <div id="front-view" className="card">
+                        <QRCodeElement svgString={ svgString } />
+                        <Logo />
+                        <div id="instructions">
+                            <InstructionIcon stylingClass="instruction-icon" />
+                            To scan QT code, Open your Venmo App
```
mstuart, line 63: I can't believe this is my first and only comment (so far) on this PR... I'm losing it :)

QT -> QR 😃

bluepnume (approved):
Looking pretty great to me! Thanks Brice

PR #160, opened 2021-05-13
Fallback to smartAPI

mnicpt:
Attempt to fix onApprove capturing and instrument fallback scenario for insights into failure.

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -79,6 +79,7 @@ const handleSmartResponse = (response, orderID : string, restAPIError, action : 
     const { headers } = response;
     const corrID = headers[HEADERS.PAYPAL_DEBUG_ID];
 
+    // $FlowFixMe
     const { apiHeaders } = restAPIError.response;
     const apiCorrID = apiHeaders[HEADERS.PAYPAL_DEBUG_ID];
     
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -88,8 +88,8 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
 }
 
 export function captureOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    return forceRestAPI
-        ? callRestAPI({
+    if (forceRestAPI) {
+        return callRestAPI({
             accessToken: facilitatorAccessToken,
             method:      `post`,
             url:         `${ ORDERS_API_URL }/${ orderID }/capture`,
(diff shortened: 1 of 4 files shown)


mstuart:
lgtm! @bluepnume @abarco ?

src/props/onApprove.js
```diff
@@ -61,21 +61,25 @@ type ActionOptions = {|
...
-        if (isProcessorDecline) {
-            return restart().then(unresolvedPromise);
-        }
+    if (isProcessorDecline) {
+        return restart().then(unresolvedPromise);
+    }
 
-        throw new Error('Order could not be captured');
-    };
+    throw new Error(`Order could not be captured for action: ${ action }`);
```
bluepnume, line 79: Probably this message should not hard-code the word 'capture' -- could we actually just do `throw err;` and re-throw?

src/api/order.js
```diff
@@ -88,45 +88,75 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
...
+        }).catch(() => {
+            return callSmartAPI({
+                accessToken: buyerAccessToken,
+                method:      'post',
+                url:         `${ SMART_API_URI.ORDER }/${ orderID }/capture`,
+                headers:     {
+                    [HEADERS.CLIENT_CONTEXT]: orderID
+                }
+            }).then(smartResponse => {
+                getLogger().info(`lsat_uprade_shadow_success_capture`, { orderID });
```
bluepnume, line 109: For this message let's add:

- The correlation id from the REST call
- The correlation id from the Smart call

Then we can easily share samples with the api team.

mnicpt: @bluepnume Do the correlation id's come from the response?  Also, this log only is in response to the smartAPI call.

mnicpt: Not sure where to get the corridor from REST call. I see that we log that already if it errors.

src/api/order.js
```diff
@@ -68,65 +68,112 @@ export function createOrderID(order : OrderCreateRequest, { facilitatorAccessTok
     });
 }
 
+const handleSmartResponse = (response, orderID : string, action : string) => {
+    const { headers } = response;
+    const corrID = headers[HEADERS.PAYPAL_DEBUG_ID];
+    
+    getLogger().info(`lsat_uprade_shadow_success_get_${ action }`, { corrID, orderID });
```
bluepnume, line 75: Could we also get the corrID for the failed REST call?

bluepnume: Sorry to nitpick :( but can we get the REST corrid in this specific log too? Just so we don't have to search CAL a bunch for two different logs with the same button session id, we'll have both corr ids in one place to compare.

bluepnume:
One small suggested change, plus let's ramp lsat upgrade to 100, then get this merged 👍

src/api/order.js
mnicpt: No need to be sorry.  Appreciate the nitpick.

bluepnume (approved):
Gonna make up for beating you in poker by approving this one.

PR #161, opened 2021-05-14
Make button render errors generate clientErrorResponse

agnjunio:
When the render method for paypal-checkout-components throws an Error, it should be considered an clientErrorResponse and not a serverErrorResponse, thus returning response code 400.

### DIFF
--- a/server/components/buttons/middleware.js
+++ b/server/components/buttons/middleware.js
@@ -65,13 +65,13 @@ export function getButtonMiddleware({
             for (const name of Object.keys(req.cookies || {})) {
                 logger.info(req, `smart_buttons_cookie_${ name || 'unknown' }`);
             }
-            
+
             tracking(req);
 
             const { env, clientID, buttonSessionID, cspNonce, debug, buyerCountry, disableFunding, disableCard, userIDToken, amount,
                 merchantID: sdkMerchantID, currency, intent, commit, vault, clientAccessToken, basicFundingEligibility, locale,
                 clientMetadataID, pageSessionID, correlationID, cookies, enableFunding, style, paymentMethodNonce, branded, fundingSource } = getButtonParams(params, req, res);
-            
+
             const { label, period, tagline } = style;
             logger.info(req, `button_params`, { params: JSON.stringify(params) });
 
@@ -138,7 +138,7 @@ export function getButtonMiddleware({
             const wallet = await walletPromise;
             const personalization = await personalizationPromise;
             const brandedDefault = await isFundingSourceBranded(req, { clientID, fundingSource, wallet });
-            
+
             const eligibility = {
                 cardFields: isCardFieldsExperimentEnabled
             };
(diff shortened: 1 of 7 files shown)


agnjunio:
Hey @gregjopa , can you please review this change?

server/components/buttons/middleware.js
```diff
@@ -145,15 +145,16 @@ export function getButtonMiddleware({
                 fundingEligibility, content, wallet, personalization
             };
 
+            let buttonHTML;
             try {
                 if (render.button.validateButtonProps) {
                     render.button.validateButtonProps(buttonProps);
                 }
+
+                buttonHTML = render.button.Buttons(buttonProps).render(html());
```
gregjopa, line 154: @agnjunio this change looks good! Could you add a unit test for it? It would be great to test that an error is caught and the expected error status code is returned.

agnjunio:
@gregjopa When I was writing the tests, I found out that the lack of eligible funds causes the validateButonProps to pass. but render to fail. I wonder if `render.button.validateButtonProps(buttonProps)` should validate only the data type or if it should validate business logic as well.

This would make impossible to render to throw any errors, since they would be already validated on `render.button.validateButtonProps(buttonProps)`

bluepnume:
I think we should try to keep 400s to `validateButtonProps` and validate anything there that would cause an error thrown from render.

Catching all errors from buttons and treating as a 4xx I'm not sure is correct. This could mask real bugs.

We could also throw a specific validation error type from the render function and send 4xx only for that case? Like set `err.code = 'VALIDATION_ERROR'` and check for that?

agnjunio:
Trying to catch all 400s in `validateButtonProps` is the best approach for this.

We should make an effort so this never happens: https://github.com/paypal/paypal-checkout-components/blob/master/src/ui/buttons/buttons.jsx#L105

I can implement the same validation logic in `validateButtonProps`. What do you guys think?

gregjopa:
> Trying to catch all 400s in `validateButtonProps` is the best approach for this.
> 
> We should make an effort so this never happens: https://github.com/paypal/paypal-checkout-components/blob/master/src/ui/buttons/buttons.jsx#L105
> 
> I can implement the same validation logic in `validateButtonProps`. What do you guys think?

Hi @agnjunio I agree. Let's move this logic to `validateButtonProps` instead. That way we avoid masking real bugs.

agnjunio:
@gregjopa @bluepnume Now that checkout-components is throwing a ValidationError `err.code = ERROR_CODE.VALIDATION_ERROR` , this PR has changed its focus.
First, we [have moved](https://github.com/paypal/paypal-sdk-constants/pull/59) `ERROR_CODE` to sdk-constants, so we need to update SPB reference.
Second, this pr also adds a test to verify whether the Button render is returning a 400 error (clientErrorResponse), instead of the previous 500 (serverErrorResponse).

What do you guys think? Is this good to go?

PR #162, opened 2021-05-18
Enable LSAT upgrade in exports

amyegan:


dist/smart-menu.js
```diff
@@ -1056,13 +1056,13 @@ window.spb = function(modules) {
     }
```
bluepnume, line 1056: Could you back out dist changes? This will be auto-genned when we publish. Thanks!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/smart-payment-buttons",
-  "version": "5.0.23",
+  "version": "5.0.24-2",
```
bluepnume, line 3: Same here

src/api/order.js
```diff
@@ -391,6 +394,9 @@ export function approveOrder({ orderID, planID, buyerAccessToken } : ApproveOrde
             [ HEADERS.CLIENT_CONTEXT ]: orderID
         }
     }).then(({ approvePayment }) => {
+        // eslint-disable-next-line no-console
+        console.log('@@@ approve payment', approvePayment);
+        setBuyerAccessToken(approvePayment.buyer.auth.accessToken);
```
bluepnume, line 399: Looks good! Can we also use this set/get in `payment-flows/checkout.js` so we have a single place to store the AT for any flow?

src/button/exports.js
```diff
@@ -62,7 +66,19 @@ export function setupExports({ props, isEnabled } : ExportsProps)  {
                     return onApprove(data, actions);
                 },
                 onCancel,
-                onError
+                onError,
+                // eslint-disable-next-line no-shadow
+                upgradeFacilitatorAccessToken: (facilitatorAccessToken, orderID) => {
+                    const buyerAccessToken = getBuyerAccessToken() || '';
```
bluepnume, line 72: If buyer access token is not present we should throw an error here. Good for caller to know that the operation couldn't go ahead.

bluepnume: (plus a CAL log would be helpful too so we can track on our side)

src/lib/session.js
```diff
@@ -32,3 +32,11 @@ export function getSessionState<T>(handler : (state : Object) => T) : T {
 export function getStickinessID() : string {
     return (hashStr(getBrowserFingerprint().toString()) % 100).toString();
 }
+
+let buyerAccessToken = 'defaultTokenValue';
```
bluepnume, line 36: Probably can just default this to undefined. May cause confusion we we try to use this default value for anything real.

bluepnume:
This is looking great! Few minor comments.

src/api/order.js
```diff
@@ -542,6 +545,7 @@ export function approveOrder({ orderID, planID, buyerAccessToken } : ApproveOrde
             [ HEADERS.CLIENT_CONTEXT ]: orderID
         }
     }).then(({ approvePayment }) => {
+        setBuyerAccessToken(approvePayment.buyer.auth.accessToken);
```
shrutikapoor08, line 548: to prevent `undefined` property access, you can do checks here such as 
`if(approvePayment.buyer)` or `approvePayment?.buyer?.auth?.accessToken`

src/api/order.js
```diff
@@ -828,7 +835,8 @@ export function payWithPaymentMethodToken({ orderID, paymentMethodToken, clientI
             [ HEADERS.CLIENT_METADATA_ID ]: clientMetadataID
         }
     }).then(({ approvePaymentWithNonce }) => {
-        getLogger().info('pay_with_paymentMethodToken', JSON.stringify(approvePaymentWithNonce));
+        getLogger().info('pay_with_paymentMethodNonce', approvePaymentWithNonce.buyer.userId);
```
shrutikapoor08, line 838: same comment about undefined property access here and next line

src/button/exports.js
```diff
@@ -50,7 +53,8 @@ export function setupExports({ props, isEnabled } : ExportsProps)  {
                 },
                 onApprove: (merchantData) => {
                     const data = {
-                        payerID: merchantData.payerID
+                        payerID:      merchantData.payerID,
+                        forceRestAPI: merchantData.forceRestAPI || false
```
shrutikapoor08, line 57: curious to learn - what does forceRestAPI do?

shrutikapoor08 (commented):
minor comments. looks good otherwise.

src/api/order.js
amyegan: Good call. Fixed!

src/button/exports.js
amyegan: If `forceRestAPI` is true, it will try to use the upgraded access token with the rest api so we don't need the `buyerAccessToken` which the merchant button won't have with HXO. If `forceRestAPI` is false, it will use the smart api and client-side integrations would not work with HXO in that case because of the missing `buyerAccessToken`

PR #163, opened 2021-05-18
remove the branded check

andyfen:


mnicpt (approved):

bluepnume (changes requested):
Recommend we don't merge this until we have a good idea what the problem is with the Elmo. Would prefer to fix on the elmo side.

PR #165, opened 2021-05-19
WIP: Remove sdkMeta from universal link

nbierdeman:
### Purpose

This PR removes `sdkMeta` from the universal link when `channel=desktop-web`.

### Notes

- Need to add a test in desktop mode checking that `sdkMeta` has been removed, once there's a web path to this code.

### DIFF
--- a/src/payment-flows/native/url.js
+++ b/src/payment-flows/native/url.js
@@ -141,7 +141,7 @@ function getNativeUrlQueryParams({ props, serviceData, config, fundingSource, se
 export function getNativeUrl({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID } : GetNativeUrlOptions) : string {
     const queryParams = getNativeUrlQueryParams({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID });
     
-    if (fundingSource === FUNDING.VENMO) {
+    if (queryParams.channel === 'desktop-web') {
         delete queryParams.sdkMeta;
     }
 
--- a/src/payment-flows/native/url.js
+++ b/src/payment-flows/native/url.js
@@ -141,6 +141,10 @@ function getNativeUrlQueryParams({ props, serviceData, config, fundingSource, se
 export function getNativeUrl({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID } : GetNativeUrlOptions) : string {
     const queryParams = getNativeUrlQueryParams({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID });
     
+    if (fundingSource === FUNDING.VENMO) {
+        delete queryParams.sdkMeta;
+    }
+
     return extendUrl(`${ getNativeDomain({ props }) }${ NATIVE_CHECKOUT_URI[fundingSource] }`, {
         // $FlowFixMe
         query: queryParams
--- a/src/payment-flows/native/url.js
+++ b/src/payment-flows/native/url.js
@@ -130,7 +130,7 @@ function getNativeUrlQueryParams({ props, serviceData, config, fundingSource, se
         buyerCountry,
         sdkVersion
     };
-
+    // shorten universal link to make the QR Code easier to read
     if (queryParams.channel === CHANNEL.DESKTOP) {
         delete queryParams.sdkMeta;
     }
--- a/src/payment-flows/native/url.js
+++ b/src/payment-flows/native/url.js
@@ -141,10 +141,6 @@ function getNativeUrlQueryParams({ props, serviceData, config, fundingSource, se
 export function getNativeUrl({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID } : GetNativeUrlOptions) : string {
     const queryParams = getNativeUrlQueryParams({ props, serviceData, config, fundingSource, sessionUID, pageUrl, orderID, stickinessID });
     
-    if (queryParams.channel === CHANNEL.DESKTOP) {
-        delete queryParams.sdkMeta;
-    }
-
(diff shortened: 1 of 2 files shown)


src/payment-flows/native/url.js
```diff
@@ -124,6 +124,10 @@ function getNativeUrlQueryParams({ props, serviceData, fundingSource, sessionUID
 export function getNativeUrl({ props, serviceData, fundingSource, firebaseConfig, sessionUID, pageUrl, orderID, stickinessID } : GetNativeUrlOptions) : string {
     const queryParams = getNativeUrlQueryParams({ props, serviceData, fundingSource, sessionUID, firebaseConfig, pageUrl, orderID, stickinessID });
     
+    if (fundingSource === FUNDING.VENMO) {
+        delete queryParams.sdkMeta;
```
gregjopa, line 128: Does the native sdk use `sdkMeta` to get the client-id of the transaction? Or do they get that another way? client-id gets passed to the native eligibility call from both the web and native and I want to make sure that's not impacted by this change. This PR is probably fine, just wanted to double check on this one.

nbierdeman: @gregjopa, good questions. I just updated this to a WIP PR, as we're still waiting for the final word from the Venmo team on whether or not it's okay to remove `sdkMeta`. We've been told "yes" before, but the last Venmo meeting we had, not everyone was so certain. I'll update this PR as soon as I know.

nbierdeman: @bluepnume, great point. Maybe it should be something like "if platform === desktop && fundingSource === venmo"?

nbierdeman: Another idea is to use the soon-to-be-added `channel` param, like "if channel === desktop-web", instead of platform, since we were going to make `channel` a generic param for all FI's. Not sure if it matters though.

mnicpt: > I think this check should be "if platform === desktop", I don't think we should couple to venmo. Venmo on ios/android could have a web fallback at some point so we shouldn't remove sdkMeta entirely.

This shouldn't ever be desktop in the native flow.

mnicpt: > Does the native sdk use `sdkMeta` to get the client-id of the transaction? Or do they get that another way? client-id gets passed to the native eligibility call from both the web and native and I want to make sure that's not impacted by this change. This PR is probably fine, just wanted to double check on this one.

I'm pretty sure we pass client-if separately. Also, I would be confirming with the Venice teams for both iOS and Android and not just Venmo on this change.

bluepnume: > Maybe it should be something like "if platform === desktop && fundingSource === venmo"?

I would say don't put Venmo in there for now. We can make it more specific later if we need. For now we can assume any desktop QR code integrations will not need sdkMeta 

Using channel would be fine with me.

> This shouldn't ever be desktop in the native flow.

It will be for Venmo. We'll need to launch a QR code and spin up firebase.

mnicpt: @bluepnume We're launching the QR code on the desktop from the native flow?

mnicpt: I'm pretty sure native doesn't use this so we may just be able to remove it.  @nbierdeman Reach out to Justin Warmkessel on the iOS team and Shyam Bhatt on the Android team about whether they use anything from the `sdkMeta` prop sent to them.

bluepnume: Native doesn't use it but we need it if the app isn't installed.

bluepnume: And yeah QR code is a native flow right? Same eligibility, same URL, same firebase messaging.

nbierdeman: @mnicpt, thanks! There's a thread in Slack w/ @mstuart and several other people on this issue. I'll see if the people you mentioned are already weighing in on it.

nbierdeman: @bluepnume, just updated this to `if (queryParams.channel === 'desktop-web')`, pending [PR 166](https://github.com/paypal/paypal-smart-payment-buttons/pull/166).

bluepnume: Looks good! Can we use the constant though? Sorry minor nitpick.

test/client/native.js
```diff
@@ -1762,10 +1762,6 @@ describe('native ios cases', () => {
 
                         const redirectQuery = parseQuery(res.redirectUrl.split('?')[1]);
 
-                        if (!redirectQuery.sdkMeta) {
```
bluepnume, line 1765: We need to keep checking if it's being sent in mobile mode

src/payment-flows/native/url.js
bluepnume: Also can we do this check in the `getNativeUrlQueryParams` function itself? So anyone using that reusable function gets the same behavior.

nbierdeman: @bluepnume, thanks for the catches. I just pushed these revisions after temporarily copying in the required code from [PR 166](https://github.com/paypal/paypal-smart-payment-buttons/pull/166) and testing.

src/payment-flows/native/url.js
```diff
@@ -127,6 +127,12 @@ function getNativeUrlQueryParams({ props, serviceData, fundingSource, sessionUID
         domain:         merchantDomain,
         rtdbInstanceID: firebaseConfig.databaseURL
     };
+
+    if (queryParams.channel === CHANNEL.DESKTOP) {
+        delete queryParams.sdkMeta;
```
gregjopa, line 132: Please add a comment here explaining why `queryParams.sdkMeta` needs to be deleted for QR Pay. It's unclear to me why this is needed.

nbierdeman: There was interest in shortening the universal link to make the QR code less dense and possibly easier to scan.

gregjopa: ah that makes sense. Can you add that as a code comment?

PR #170, opened 2021-05-24
Hide overlay after onApprove promise(s) get resolved

leogedler:
- Modified sequence on the onApprove method to first wait for all promises to get resolved before closing the popup and hiding the checkout overlay.

### DIFF
--- a/src/payment-flows/checkout.js
+++ b/src/payment-flows/checkout.js
@@ -327,14 +327,13 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
 
             onApprove: ({ payerID, paymentID, billingToken, subscriptionID, authCode }) => {
                 approved = true;
-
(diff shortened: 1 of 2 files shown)


bluepnume (approved):
This works 👍 -- in the long term we should still think about extracting the overlay to a separate component though, as this will only help with checkout and not other flows. It will also leave the popup open while the merchant is doing any work in onApprove.

bluepnume:
Can we also get a test for this case, to make sure `close()` is not called before the onApprove promise resolves?

leogedler:
> Can we also get a test for this case, to make sure `close()` is not called before the onApprove promise resolves?
 
@bluepnume  Sure, working on it.

leogedler:
@bluepnume Added test to verify that `close` get called only after `onApprove` and any possible inner promises get resolved.

gregjopa (approved):

PR #171, opened 2021-05-24
Fix LSAT Logging

mnicpt:


src/api/api.js
```diff
@@ -82,7 +82,7 @@ export function callSmartAPI({ accessToken, url, method = 'get', headers: reqHea
                 throw new Error(`Api: ${ url } returned ack: ${ body.ack } (Corr ID: ${ headers[HEADERS.PAYPAL_DEBUG_ID] })`);
             }
 
-            return body.data;
+            return { data: body.data, headers };
```
bluepnume, line 85: LGTM, can we reflect the same in the type def?

src/api/order.js
```diff
@@ -69,14 +69,22 @@ export function createOrderID(order : OrderCreateRequest, { facilitatorAccessTok
     });
 }
 
-const handleRestAPIResponse = (err, orderID : string, action : string) => {
-    getLogger().info(`call_rest_api_failure_${ action }`, { err: stringifyError(err), orderID });
+const handleRestAPIResponse = (err, orderID : string, action : string) : string => {
+    // $FlowFixMe
+    const { headers } = err.response;
```
bluepnume, line 74: Can we do `const { headers = {} } = err.response || {};` for safety?

src/api/api.js
mnicpt: @bluepnume Not so sure we should do this now.  This breaks compatibility on `actions.subcription.get()`. Maybe I can not alter that signature and still provide type for `callSmartAPI`

mnicpt: `getSubscription` and `activateSubscription` just return the result from `callSmartAPI`

mnicpt: I'll use `then(` to convert to just the body data being returned.

mnicpt:
Working on tests.

gregjopa (approved):

PR #175, opened 2021-05-26
QRCode component build

BriceShatzer:
adds the script files for the QRCode component to the dist folder


mstuart (approved):

bluepnume:
Should only be necessary to add to `webpack.config.js`. Dist will be autogenerated during build/release.

PR #179, opened 2021-05-28
Log when the latest dist tag has been downloaded

gregjopa:
This PR adds logging so we know when the `latest` dist tag has been downloaded with grabthar `poll()`.

### DIFF
--- a/server/components/buttons/script.js
+++ b/server/components/buttons/script.js
@@ -8,7 +8,7 @@ import { ENV } from '@paypal/sdk-constants';
 
 import type { CacheType } from '../../types';
 import { BUTTON_RENDER_JS, BUTTON_CLIENT_JS, SMART_BUTTONS_MODULE, CHECKOUT_COMPONENTS_MODULE,
-    BUTTON_CLIENT_MIN_JS, WEBPACK_CONFIG, ACTIVE_TAG, LATEST_TAG } from '../../config';
+    BUTTON_CLIENT_MIN_JS, WEBPACK_CONFIG, ACTIVE_TAG } from '../../config';
 import { isLocalOrTest, compileWebpack, babelRequire, evalRequireScript, resolveScript,
     dynamicRequire, type LoggerBufferType } from '../../lib';
 import { getPayPalSDKWatcher, getPayPalSmartPaymentButtonsWatcher } from '../../watchers';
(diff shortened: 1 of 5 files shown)


server/components/buttons/script.js
```diff
@@ -103,5 +105,9 @@ export async function getSmartPaymentButtonsClientScript({ logBuffer, cache, deb
     const { version } = await watcher.get(ACTIVE_TAG);
     const script = await watcher.read(debug ? BUTTON_CLIENT_JS : BUTTON_CLIENT_MIN_JS, ACTIVE_TAG);
 
+    watcher.get(LATEST_TAG).then(({ version: deployVersion }) => {
```
gregjopa, line 108: Note that we do not want this promise to be blocking. That's why it's not using `await`. We only want to fire off the log when it's done so we now the deploy version (clientsdknodeweb takes the same approach).

server/components/buttons/script.js
```diff
@@ -103,5 +105,9 @@ export async function getSmartPaymentButtonsClientScript({ logBuffer, cache, deb
...
+        logBuffer.info(`deploy_client_version_${ deployVersion.replace(/[^0-9]+/g, '_') }`, { deployVersion });
```
bluepnume, line 109: Can we:

- Log the active version too
- Do this in the other places we use grabthar too, to help diagnose future issues

server/watchers.js
```diff
@@ -10,17 +10,42 @@ import { SDK_RELEASE_MODULE, SMART_BUTTONS_MODULE, MODULE_POLL_INTERVAL, SDK_CDN
...
+        |}
+    }
+|};
+
 type Watcher = {|
-    get : (tag? : string) => Promise<{|
-        version : string
-    |}>,
+    get : (tag? : string) => Promise<ModuleDetails>,
+    getTag : () => Promise<ModuleDetails>,
```
gregjopa, line 28: I noticed that clientsdknodeweb returns `getTag()` and `getDeployTag()` as convenience functions. I'm thinking we can do the same here and also attach logging to it.

https://github.paypal.com/Checkout/clientsdknodeweb/blob/master/server/sdk/watcher.js#L45-L46

server/components/buttons/script.js
gregjopa: @bluepnume thanks for the feedback! I think I got the logging added everywhere for both active and deploy tags. Please review again.

bluepnume (approved):
LGTM!

PR #180, opened 2021-06-03
Update references to sdk-constants to reduce bundle size

agnjunio:
Signed-off-by: Agnaldo Junior <agoliveira@paypal.com>


agnjunio:
Tests can't work with `src` in import. Also, there's no evidence that bundle size is actually lower.

PR #182, opened 2021-06-07
Update LSAT Logic

mnicpt:


src/api/order.js
```diff
@@ -88,7 +88,7 @@ const handleSmartResponse = (data : Object, headers : {| [$Values<typeof HEADERS
 };
 
 export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    if (forceRestAPI) {
+    if (forceRestAPI && window[LSAT_UPGRADE_RESULT_KEY]) {
```
gregjopa, line 91: I think we may need to pass this in as a param instead since we memoize these functions based on the args: https://github.com/paypal/paypal-smart-payment-buttons/blob/eb1dd972bb6c88728656ac2b1c5c8aae85641f95/src/props/onApprove.js#L77-L80

bluepnume, line 91: We have to default to `true` since SPB does not always do the upgrade. This should be more like:

`if (forceRestAPI && !window.lsatUpgradeFailed) {`

-- so we only skip REST if we know for sure the upgrade failed

mnicpt: Or returned false, right?

mnicpt: This is set to true by default and can only be set to false if the call returns false or null or throws an error.

src/props/onAuth.js
```diff
@@ -19,31 +18,32 @@ export type OnAuth = (params : XOnAuthDataType) => ZalgoPromise<string | void>;
...
-                upgradeLSATExperiment.logStart();
-
                 if (upgradeLSAT) {
                     return createOrder()
                         .then(orderID => upgradeFacilitatorAccessToken(facilitatorAccessToken, { buyerAccessToken: accessToken, orderID }))
-                        .then(() => {
+                        .then(data => {
                             getLogger().info('upgrade_lsat_success');
+                            
+                            if (data && !data.upgradeLowScopeAccessToken) {
```
bluepnume, line 38: Why do we need this check? If the call doesn't throw an error we can assume it succeeded.

src/api/auth.js
```diff
@@ -92,7 +92,9 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
             }
         `,
         variables: { facilitatorAccessToken, buyerAccessToken, orderID }
-    }).then(noop);
+    }).then(data => {
+        return data;
```
bluepnume, line 96: I don't think this is needed. The api just returns a success/failure, it doesn't return any meaningful data.

src/button/props.js
```diff
@@ -178,6 +178,7 @@ export type ButtonProps = {|
 
 // eslint-disable-next-line complexity
 export function getProps({ facilitatorAccessToken, brandedDefault } : {| facilitatorAccessToken : string, brandedDefault : boolean | null |}) : ButtonProps {
+    window[LSAT_UPGRADE_RESULT_KEY] = true;
```
bluepnume, line 181: Nitpick, but instead of having to manually default this to `true` here, why not just make the param negative like: `window.lsatUpgradeFailed`? Then the default behavior will always be "call REST" and the exceptional case will be "skip REST and go straight to SMART"?

Little safer if we ever refactor and forget to default this to true, or if `getProps()` is ever called again (getProps doesn't sound like a function that should have any side-effects like writing a global)

src/api/order.js
```diff
@@ -161,7 +163,7 @@ export function captureOrder(orderID : string, { facilitatorAccessToken, buyerAc
 }
 
 export function authorizeOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    if (forceRestAPI) {
+    if (forceRestAPI && window[LSAT_UPGRADE_RESULT_KEY]) {
```
bluepnume, line 166: I'm also thinking `if (forceRestAPI && !window.lsatUpgradeFailed) {` makes it a little bit more obvious what this condition means; `window[LSAT_UPGRADE_RESULT_KEY]` is a little more mental overhead to understand without context.

test/client/actions.js
```diff
@@ -101,11 +101,65 @@ describe('actions cases', () => {
...
+                            throw new Error(`We haven't received the buyer's access token`);
+                        }
+
+                        if (!data.variables.orderID) {
+                            throw new Error(`We haven't received the orderID`);
+                        }
+
+                        return {
+                            data: {
+                                upgradeLowScopeAccessToken: false
```
bluepnume, line 153: Naw let's simulate a real failure and return a 500 error. Would be good to match what the GQL api actually does when it fails.

src/props/onAuth.js
```diff
@@ -19,31 +18,38 @@ export type OnAuth = (params : XOnAuthDataType) => ZalgoPromise<string | void>;
...
+                            if (window.xprops.createSubscription) {
+                                return null;
+                            }
+
+                            return upgradeFacilitatorAccessToken(facilitatorAccessToken, { buyerAccessToken: accessToken, orderID });
+                        })
+                        .then(data => {
+                            getLogger().info(`upgrade_lsat_success_${ data ? JSON.stringify(data) : 'null' }`);
+
+                            if (!data || (data && !data.upgradeLowScopeAccessToken)) {
```
gregjopa, line 44: You could use optional chaining here:
```suggestion
                            if (!data || !data?.upgradeLowScopeAccessToken) {
```

src/api/order.js
mnicpt: How about changing the name to read better?  The double negative seems worse than a positive for the positive condition, no?

src/props/onAuth.js
```diff
@@ -19,31 +18,34 @@ export type OnAuth = (params : XOnAuthDataType) => ZalgoPromise<string | void>;
...
 
         return ZalgoPromise.try(() => {
             if (accessToken) {
-                upgradeLSATExperiment.logStart();
-
                 if (upgradeLSAT) {
                     return createOrder()
-                        .then(orderID => upgradeFacilitatorAccessToken(facilitatorAccessToken, { buyerAccessToken: accessToken, orderID }))
+                        .then(orderID => {
+                            if (window?.xprops?.createSubscription) {
```
bluepnume, line 35: Can you pass in `createSubscription` as a param rather than needing to pull from `window`?

mnicpt: Yes

test/client/actions.js
mnicpt: It fails the test here if I throw Error.

bluepnume (approved):

PR #183, opened 2021-06-09
Native opt-out fallback

leogedler:
**Reasoning**
As part of a better onboarding native payment flow will be implementing an opt-out feature that gives the user a choice to use web payment flow instead native. Once a user indicates they do not wish to try native they should fall back to the web flow. In this process, JS SDK will store an opt-out lifetime indicator in localStorage on the device so that SPB no longer app switches to native until the date has expired.

**Implementation**
Now the JS SDK will look for a data type `native_opt_out` on the onFallBack event coming from firebase*. If this type exists and if it is the opt-out type the JS SDK will store the property `nativeOptOutLifetime` on the current `__smart_payment_buttons_storage__` object on the localStorage setting the lifetime to 1 week*.

There is a new method called `isNativeOptedIn` that will be used to validate whether the native flow was opt-out or not on the eligibility logic for the native payment flow.

**Considerations**
- As it stands this implementation only handle the Android flow using firebase and listening to the `onFallBack` event.
- The one week opt-out time it is just tentative time span.

### DIFF
--- a/src/payment-flows/native/native.js
+++ b/src/payment-flows/native/native.js
@@ -24,6 +24,20 @@ function setupNative({ props, serviceData } : SetupOptions) : ZalgoPromise<void>
     return prefetchNativeEligibility({ props, serviceData }).then(noop);
 }
 
+function setNativeOptOut(data? : {| type? : string,  win? : CrossDomainWindowType |}) : boolean {
+    let optOut = false;
+    if (data && data.type === FPTI_TRANSITION.NATIVE_OPT_OUT) {
+        // Opt-out 1 week from native experience
+        const OPT_OUT_TIME = 7 * 24 * 60 * 60 * 1000;
+        const now = Date.now();
+        getStorageState(state => {
+            state.nativeOptOutLifetime = now + OPT_OUT_TIME;
+        });
+        optOut = true;
+    }
+    return optOut;
+}
+
 function initNative({ props, components, config, payment, serviceData } : InitOptions) : PaymentFlowInstance {
     const { onApprove, onCancel, onError,
         buttonSessionID, onShippingChange } = props;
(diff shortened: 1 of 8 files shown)


src/payment-flows/native/native.js
```diff
@@ -149,13 +149,28 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         });
     };
 
-    const onFallbackCallback = () => {
+    const onFallbackCallback = ({ data } : {| data? : {| type? : string,  win? : CrossDomainWindowType |} |}) => {
+        
         return ZalgoPromise.try(() => {
+            let optOut = false;
```
bluepnume, line 155: let's extract this block into a separate function, like `setNativeOptOut`

src/payment-flows/native/native.js
```diff
@@ -149,13 +149,28 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
...
+                getStorageState(state => {
+                    state.nativeOptOutLifetime = now + OPT_OUT_TIME;
+                });
+                optOut = true;
+            }
+
             getLogger().info(`native_message_onfallback`)
                 .track({
-                    [FPTI_KEY.TRANSITION]:  FPTI_TRANSITION.NATIVE_ON_FALLBACK
+                    [FPTI_KEY.TRANSITION]:  !optOut ? FPTI_TRANSITION.NATIVE_ON_FALLBACK : FPTI_TRANSITION.NATIVE_OPT_OUT
```
bluepnume, line 168: I would say let's have a different key, or an extra transition -- since this is still also a fallback case.

test/client/native.js
```diff
@@ -4320,4 +4320,189 @@ describe('native chrome cases', () => {
...
+
+            await clickButton(FUNDING.PAYPAL);
+            await window.xprops.onApprove.await();
+
+            await mockWebSocketServer.done();
+            postRobotMock.done();
+            gqlMock.done();
+            mockWindow.done();
+
+            getStorageState(state => {
```
bluepnume, line 4494: I think this is too much testing an implementation detail. We should run another button click and make sure the flow goes immediately to web as expected.

src/payment-flows/native/native.js
leogedler: I have created the recommended method.

src/payment-flows/native/native.js
leogedler: I have created an extra transition key called `TRANSITION_TYPE` to log the type of the transition.

test/client/native.js
leogedler: Got it, I have added an extra click that continues with the web flow.

leogedler:
@bluepnume Thank you for your feedback!. I have addressed your comments.

I'm still a bit confused with the iOS path. 

I have added an extra logic to expect a `type` param from the URL on this [line](https://github.com/paypal/paypal-smart-payment-buttons/pull/183/files#diff-122aff48da71baeec4a7106ef209b1c6beaafcfa0e170ad168097c78cafa2296R249).

That should get passed to the parent and an `onFallback` event should be trigged [here](https://github.com/paypal/paypal-smart-payment-buttons/pull/183/files#diff-a28f426627409909aa0875d9bb7eb1e8f43a8c052d87d1dffec92c89bb11c4e2R320) .

Then the flow should continue the same as if the event were triggered by firebase.

I'd like to know your thoughts and/or recommendations.

bluepnume (approved):

mnicpt (approved):

test/client/mocks.js
```diff
@@ -1410,6 +1411,26 @@ export function getNativeFirebaseMock({ getSessionUID, extraHandler } : {| getSe
         waitingForResponse.push(onApproveRequestID);
     };
 
+    const onFallbackOptOut = () => {
+        fallbackRequestID = `${ uniqueID()  }_onApprove`;
```
mnicpt, line 1415: Why does this have `onApprove` in it?

leogedler: @mnicpt was a typo, thanks for noticing this.

I have changed it to `fallback`.

mnicpt:
@leogedler Do you know why `server/watchers.js` is down 37% in coverage?

leogedler:
> @leogedler Do you know why `server/watchers.js` is down 37% in coverage?

@mnicpt This wasn't the case before yesterday when I rebased master, I'm rebasing again to re-check.

PR #184, opened 2021-06-09
[WIP] fix(spinner): Removing button spinner for Native and ApplePay

TNuzzi:
Fixes an issue when the user returns back from the native (PayPal or Venmo) app and the button still has the spinner applied.  This is problematic when the user cancels the payment (in the native app) but can't proceed because the spinner deactivates the payment button..

### DIFF
--- a/src/payment-flows/applepay/applepay.js
+++ b/src/payment-flows/applepay/applepay.js
@@ -491,5 +491,5 @@ export const applepay : PaymentFlow = {
     isEligible:        isApplePayEligible,
     isPaymentEligible: isApplePayPaymentEligible,
     init:              initApplePay,
-    spinner:           true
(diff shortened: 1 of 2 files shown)


TNuzzi:
I was able to verify that this fix removes the spinner however the button remains disabled.  I am looking into an additional update to this fix.

mnicpt:
@TNuzzi Please don't remove spinner for Apple Pay.

TNuzzi:
@mnicpt thanks for letting me know.  I am closing this PR.

TNuzzi:
@bluepnume I am opening another PR.  We seeing an issue where the `resolve` is not being called in `initPopUpAppSwitch` for IOS only.  I will tag you for review.

PR #185, opened 2021-06-10
fix(iOS): initPopupAppSwitch promise not resolving.

TNuzzi:
Problem: on iOS any funding button clicked will remain inactive with a spinner.  This happens on all scenarios. 

We are seeing a scenario where the promise `resolve` is not being called on iOS only.  Android uses a different mechanism (`setTimeout`, close after a second and call `detectAppSwitch`) where as iOS used the `onPostMessage` to trigger `onDetectAppSwitch`.

Could not determine the exact reason why for iOS this did not work.  As a solution, `resolve` the promise at the bottom of `initPopupAppSwitch`.

### DIFF
--- a/src/payment-flows/native/native.js
+++ b/src/payment-flows/native/native.js
@@ -271,6 +271,8 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
             });
 
             clean.register(nativePopup.cancel);
+
+            resolve();
         });
     };
 


TNuzzi:
CC @bluepnume

src/payment-flows/native/native.js
```diff
@@ -271,6 +271,8 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
             });
 
             clean.register(nativePopup.cancel);
+
+            resolve();
```
gregjopa, line 275: I would expect `onDetectAppSwitch` to resolve the promise here for this native app switch use case. Any idea why it's not?

TNuzzi: @gregjopa Both @mnicpt and I looked into it and we could not see the exact reason why.  It was curious that `Android` was fine but is called sooner in the process.  I wonder if it is a threading/backgrounding issue?  For `iOS` the user is already navigated away for the browser and in the Venmo/PayPal app.

gregjopa: > I wonder if it is a threading/backgrounding issue?
It sounds like it. Definitely a tricky issue. I'm fine with proceeding with this PR.

bluepnume: Yeah I think it would be better to investigate and find the reason `onDetectAppSwitch` is not resolving. For one thing, the tests right now rely on this promise only resolving after that happens.

TNuzzi: @bluepnume Thanks for helping be debug this issue and finding the root cause.  The JS SDK is attempting to pass the props to the native SDK but not getting a response.  The JS SDK logs `native_message_setprops` but the promise never resolves and logs `native_response_setprops`.  The issue is only happening on iOS.  Here is the line that is not resolving:

https://github.com/paypal/paypal-smart-payment-buttons/blob/0fe4c1deab3d67018a7b90404dea4ac51d5fa590/src/payment-flows/native/native.js#L208

TNuzzi:
Closing is PR.

PR #187, opened 2021-06-21
Prevent onCancel if onClick Fails

mnicpt:
https://engineering.paypalcorp.com/jira/browse/DTCHKINT-867

### DIFF
--- a/src/payment-flows/checkout.js
+++ b/src/payment-flows/checkout.js
@@ -241,7 +241,6 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
     const connectEligible = isConnectEligible({ connect, createBillingAgreement, createSubscription, vault, fundingSource });
 
     let approved = false;
-    let orderCreated = false;
     let forceClosed = false;
 
     const init = () => {
@@ -321,7 +320,6 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
                                 clientID, merchantID, buyerCountry, currency, commit, intent, disableFunding, disableCard, userIDToken });
                         }
                     }).then(() => {
-                        orderCreated = true;
                         return orderID;
                     });
                 });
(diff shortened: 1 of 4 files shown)


src/payment-flows/native/native.js
```diff
@@ -274,10 +274,9 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
     const onCloseCallback = () => {
         return ZalgoPromise.delay(1000).then(() => {
             if (!approved && !cancelled && !didFallback && !isAndroidChrome()) {
-                return ZalgoPromise.all([
-                    onCancel(),
-                    destroy()
-                ]);
+                return ZalgoPromise.try(() => {
```
gregjopa, line 277: Why doesn't native need to call `onCancel` when an orderID is already created?

gregjopa (approved):
Looks good to me! I left one question about the native flow.

src/payment-flows/native/native.js
mnicpt: This callback occurs when popup is closed either manually or from false being returned from `onClick`.  In each of these cases we don't want to call `onCancel`.  Closing manually follows same pattern as checkout flow.

src/payment-flows/checkout.js
```diff
@@ -241,6 +241,7 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
     const connectEligible = isConnectEligible({ connect, createBillingAgreement, createSubscription, vault, fundingSource });
 
     let approved = false;
+    let orderCreated = false;
```
bluepnume, line 244: I think we don't need these changes in `checkout.js`. `start()` should not even be called if `onClick()` returns false so this should never happen.

bluepnume (approved):

PR #188, opened 2021-06-21
Use merchantID instead of sdkMerchantID for personalization

ssono:
Currently adding logic in personalization that requires the merchantID. There is no special logic in the graphql layer that requires the sdkMerchatID so this change should be fine.

https://engineering.paypalcorp.com/jira/browse/DTNAVYA-1102

### DIFF
--- a/test/server/render.button.test.js
+++ b/test/server/render.button.test.js
@@ -216,3 +216,34 @@ test('should render empty personalization when config is disabled', async () => 
         throw new Error(`Expected personalization to be empty, got: ${ JSON.stringify(setupButtonParams.personalization) }`);
     }
 });
+
+test('should render filled out tagline when config is enabled', async () => {
+    const buttonMiddleware = getButtonMiddleware({
+        graphQL,
+        getAccessToken,
+        getMerchantID,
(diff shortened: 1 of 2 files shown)


mnicpt:
@ssono Can you paste your JIRA ticket in the description?

mnicpt (approved):
LGTM

mnicpt:
@bluepnume Do you see any issue with this?

mnicpt:
@ssono Looks like we need better test coverage.

test/server/render.button.test.js
```diff
@@ -216,3 +216,34 @@ test('should render empty personalization when config is disabled', async () =>
...
+        cache,
+        logger,
+        tracking,
+        getPersonalizationEnabled: () => true,
+        isFundingSourceBranded
+    });
+
+    const req = mockReq({
+        query: {
+            clientID:                     'xyz'
```
gregjopa, line 235: nit: I don't think this white space is required
```suggestion
            clientID: 'xyz'
```

gregjopa (approved):

PR #190, opened 2021-06-21
[DTCHKINT-594] Adding test cases for SPB nonce flow

shrutikapoor08:
- Adding positive cases and negative cases
- 
Screen Shot 2021-06-21 at 2 40 13 PM

Screen Shot 2021-06-21 at 2 39 56 PM


test/client/nonce.js
```diff
@@ -0,0 +1,616 @@
...
+                                                stringValue: 'xyz-us-b1@paypal.com'
+                                            }
+                                        }
+                                    ]
+                                }
+                            }
+                        };
+                    }
+
+                    if (data.query.includes('mutation ApprovePaymentWithNonce')) {
```
bluepnume, line 139: Can we throw an error if `ApprovePaymentWithNonce` is not called? Want to make sure the test is validating this path is hit

shrutikapoor08: We have got `.expectCalls` so I am expecting the test case will fail if the mutation is not called. Is that correct assumption?

bluepnume: That will only guarantee that GQL is called (which could be for many different queries) -- it will not assert that specifically `ApprovePaymentWithNonce` was hit.

shrutikapoor08: Done and updated!

bluepnume (approved):

gregjopa (approved):

gregjopa:
@shrutikapoor08 can you resolve the conflict with constants.js? Let's merge it after that.

shrutikapoor08:
@gregjopa resolved conflicts👍🏼

gregjopa:
Thanks @shrutikapoor08! Let's merge after CI passes

gregjopa:
It looks like CI is failing due to this test:
```
1) should render a button, click the button, and render checkout with an explicit email merchantID and multiple correct payees with the same email, and succeed
     payee cases
     Error: Expected intent from order api call to be authorize, got capture. Please ensure you are passing intent=capture to the sdk url. https://developer.paypal.com/docs/checkout/reference/customize-sdk/
    at triggerIntegrationError (test/index.js:80061:27)
    at test/index.js:86067:65
```

bluepnume:
BTW @shrutikapoor08 we're renaming `paymentMethodNonce` to `paymentMethodToken` to reflect new Braintree best-practices. I've made this change already and I've also renamed `nonce.js` to `branded-vault-card.js` for now, since it's not the only payment method file which will deal with PMTs. Would be good to update these tests to reflect the new naming. Thanks!

mrbaguvix (commented):
@shrutikapoor08 Can we ensure that these checks pass.

PR #196, opened 2021-06-24
Native flow opt-out duration

leogedler:
Added logic to receive a prop called `duration` from the native flow opt-out fallback.

- If this prop is passed in and it a valid number representing `milliseconds` it will be used to set the lifetime duration of the native flow opt-out.
- If the duration prop is not passed in, the JS SDK will set a default duration of a `1 week` for the native flow opt-out.

### DIFF
--- a/src/payment-flows/native/native.js
+++ b/src/payment-flows/native/native.js
@@ -24,15 +24,15 @@ function setupNative({ props, serviceData } : SetupOptions) : ZalgoPromise<void>
     return prefetchNativeEligibility({ props, serviceData }).then(noop);
 }
 
-function setNativeOptOut(data? : {| type? : string, duration? : number,  win? : CrossDomainWindowType |}) : boolean {
+function setNativeOptOut(data? : {| type? : string, skip_native_duration? : number,  win? : CrossDomainWindowType |}) : boolean {
     let optOut = false;
     if (data && data.type === FPTI_TRANSITION.NATIVE_OPT_OUT) {
-        const { duration } = data;
+        const { skip_native_duration } = data;
 
-        // Opt-out 1 week from native experience as default
-        let OPT_OUT_TIME = 7 * 24 * 60 * 60 * 1000;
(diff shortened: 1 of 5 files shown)


src/payment-flows/native/native.js
```diff
@@ -24,11 +24,17 @@ function setupNative({ props, serviceData } : SetupOptions) : ZalgoPromise<void>
     return prefetchNativeEligibility({ props, serviceData }).then(noop);
 }
 
-function setNativeOptOut(data? : {| type? : string,  win? : CrossDomainWindowType |}) : boolean {
+function setNativeOptOut(data? : {| type? : string, skip_native_duration? : number,  win? : CrossDomainWindowType |}) : boolean {
     let optOut = false;
     if (data && data.type === FPTI_TRANSITION.NATIVE_OPT_OUT) {
-        // Opt-out 1 week from native experience
-        const OPT_OUT_TIME = 7 * 24 * 60 * 60 * 1000;
+        const { skip_native_duration } = data;
```
mnicpt, line 30: Seems like this may be a problem if `skip-native_duration` is not defined since it is optional.  Should we move this into the conditional below?

src/payment-flows/native/native.js
```diff
@@ -169,7 +175,7 @@ function initNative({ props, components, config, payment, serviceData } : InitOp
         });
     };
 
-    const onFallbackCallback = ({ data } : {| data? : {| type? : string,  win? : CrossDomainWindowType |} |}) => {
+    const onFallbackCallback = ({ data } : {| data? : {| type? : string, skip_native_duration? : number,  win? : CrossDomainWindowType |} |}) => {
```
mnicpt, line 178: How are we using this for fallback?  Fallback should only be web.

mnicpt: Is this for native to override?

mnicpt (approved):
LGTM! 🎉

mstuart:
Looks great!

PR #198, opened 2021-06-29
Fix LSAT Logic

mnicpt:
**ISSUE:**
`window[LSAT_UPGRADE_FAILED] && !window[LSAT_UPGRADE_FAILED]` returned true if it didn't exist on the Window object.  Also, no boolean if successful so didn't know if undefined meant successful or onAuth was called or not.

**FIX:**
`const lsatUpgraded =window[LSAT_UPGRADE_FAILED] === false;`

LSAT_UPGRADE_FAILED exists AND true  => false
LSAT_UPGRADE_FAILED exists AND false => true
LSAT_UPGRADE_FAILED NOT exists => false

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -88,7 +88,7 @@ const handleSmartResponse = (data : Object, headers : {| [$Values<typeof HEADERS
 };
 
 export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    const lsatUpgraded = LSAT_UPGRADE_FAILED in window && !window[LSAT_UPGRADE_FAILED];
+    const lsatUpgraded = window[LSAT_UPGRADE_FAILED] === false;
     
     if (forceRestAPI && lsatUpgraded) {
         return callRestAPI({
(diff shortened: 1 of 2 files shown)


src/api/order.js
```diff
@@ -88,7 +88,7 @@ const handleSmartResponse = (data : Object, headers : {| [$Values<typeof HEADERS
 };
 
 export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    if (forceRestAPI && !window[LSAT_UPGRADE_FAILED]) {
+    if (forceRestAPI && (window[LSAT_UPGRADE_FAILED] && !window[LSAT_UPGRADE_FAILED])) {
```
gregjopa, line 91: Could we check for the exact boolean value like this?
```suggestion
    if (forceRestAPI && window[LSAT_UPGRADE_FAILED] === false) {
```

gabrielo91 (approved):

src/api/order.js
```diff
@@ -88,7 +88,9 @@ const handleSmartResponse = (data : Object, headers : {| [$Values<typeof HEADERS
 };
 
 export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    if (forceRestAPI && !window[LSAT_UPGRADE_FAILED]) {
+    const lsatUpgraded = LSAT_UPGRADE_FAILED in window && !window[LSAT_UPGRADE_FAILED];
```
gregjopa, line 91: I think this would work as well and eliminate the need to check for the existence of the key on window.
```suggestion
    const lsatUpgraded = window[LSAT_UPGRADE_FAILED] === false;
```

mnicpt: Perfect!

mnicpt:
Looks like I have a broken test.

PR #199, opened 2021-06-29
Add support for testing the sdk cdn in stage

gregjopa:
The JS SDK supports the `cdn-registry` and `version` query params in lower environments (https://github.com/paypal/paypal-sdk-constants/pull/54). This PR updates SPB to respect these params for the second render so it uses the same JS SDK bundle when rendering server-side.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "grabthar": "^4",
     "jsx-pragmatic": "^2",
     "module-alias": "^2.2.2",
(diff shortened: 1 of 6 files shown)


gregjopa:
An alternative solution is to move this logic to smartcomponentnodeweb and pass down a helper function to the middleware.

gregjopa:
Closing in favor of moving this logic upstream

PR #176, opened 2021-05-27
Qrcode functionality

BriceShatzer:
[Setting up local](https://gist.github.com/BriceShatzer/7568d93fcf99bd9e5ccb4e1484207318#file-local-setup-notes-md)

What is happening:  
- Venmo button is displayed on desktop 
- When it's clicked native.js [knows were on desktop](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/payment-flows/native/eligibility.js#L90-L95) and: 
  - [generates a QRCode component](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/payment-flows/native/native.js#L271) 
  - sets up [listeners](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/qrcode/qrcard.jsx#L78) for events [fired](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/payment-flows/native/native.js#L298) in native.js's various "on___Callback" code (which triggers the different UI [state changes](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/qrcode/qrcard.jsx#L38))
  - since it's utilizing the existing callback code, it continues the native.js payment-flow like a regular payment on [approval](https://github.com/BriceShatzer/paypal-smart-payment-buttons/blob/74e0310c651607b6014717ce5b3eff2fefcf6904/src/payment-flows/native/native.js#L304).

PR #200, opened 2021-06-30
Stop including UserAgent in getNativeUrl

BriceShatzer:
**This PR removes the inclusion of `userAgent` when constructing the native url via `getNativeUrl`**

When we call `getNativeUrl` to generate the URL that powers both the native app switch and is converted into a QRCode that is scanned in our native desktop flow, one of the query parameters that is included is an encoded string that provides the `UserAgent` of the web browser where native payment flow was initiated.  [Per folks on both the android & ios teams](https://github.com/paypal/paypal-smart-payment-buttons/pull/200#discussion_r673317688), this information is no longer used by either application. 

Providing vestigial data in the native url had no downside when it was solely as a means to power the native app switch. With that URL now being converted into a QRCode, that additional data is forcing the rendered QRCode Image to be far more complex (and visually less appealing). That additional data further adds to the complexity in that it is requiring the least compressible "[mode](https://github.com/soldair/node-qrcode#encoding-modes)", further exacerbating the problem. 

[Jira Ticket](https://engineering.paypalcorp.com/jira/browse/DTCHKINT-885)

*Branch originally forked off of https://github.com/paypal/paypal-smart-payment-buttons/pull/176/ which has since been merged*


src/payment-flows/native/url.js
```diff
@@ -122,7 +115,6 @@ function getNativeUrlQueryParams({ props, serviceData, config, fundingSource, se
         commit:         String(commit),
         webCheckoutUrl: isIOSSafari() ? webCheckoutUrl : '',
         stickinessID,
-        userAgent,
```
gregjopa, line 125: I think we need to keep `userAgent`. It's a param passed to graphql in the native eligibility call. That call is made by SPB and also the native SDK. so we need to keep it

BriceShatzer: so `getNativeUrlQueryParams` is used in `getNativeUrl` & `getNativeFallbackUrl`

`getNativeUrl` is used 


[payment-flows/native/popup.js](https://github.com/paypal/paypal-smart-payment-buttons/blob/cb7826ab1152dc05eb74bf735610029340146ad9/src/payment-flows/native/popup.js#L214) => `awaitRedirectListener`  
uses both `getNative` & `getNativeFallbackUrl`

& 

[payment-flows/native/native.js](https://github.com/paypal/paypal-smart-payment-buttons/blob/cb7826ab1152dc05eb74bf735610029340146ad9/src/payment-flows/native/native.js#L319) => `initQRCode`  
uses just `getNative`

So, assuming that the native apps don't have a use `userAgent` as a query param in their universal links, it should be safe to remove from that perspective.  I'm tracking down some mobile folks to confirm this is the case. 

---

@mstuart also mentioned that [xobuynernodeserv](https://github.paypal.com/Checkout-R/xobuyernodeserv/blob/develop/src/graphql/data/SDKEligibility/venmo/eligibility.js#L28) has kind of a "use it if it's there, but doesn't actually need it" approach to user agent as query param. 

As for references to userAgent for eligibility calls within SPB, user agent is pulled out of the query params in:
- [`resolveFundingEligibility`](https://github.com/paypal/paypal-smart-payment-buttons/blob/cb7826ab1152dc05eb74bf735610029340146ad9/server/service/fundingEligibility.js#L164)
- [`resolvePersonalization`](https://github.com/paypal/paypal-smart-payment-buttons/blob/cb7826ab1152dc05eb74bf735610029340146ad9/server/service/personalization.jsx#L134)
- [`resolveWallet`](https://github.com/paypal/paypal-smart-payment-buttons/blob/cb7826ab1152dc05eb74bf735610029340146ad9/server/service/wallet.js#L284)

all of which are used in service of the [getButtonMiddleware](https://github.com/paypal/paypal-smart-payment-buttons/blob/03c9ac029cff127929b7181c31327730bb6c4251/server/components/buttons/middleware.js#L54). I can't concieve of a situation where `getNativeUrl` or `getNativeFallbackUrl` would ever interact with or create a url that would need to be interpreted by the button middleware.   

The other references to user agent are 
- [src/api/eligibility.js => `getNativeEligibility`](https://github.com/paypal/paypal-smart-payment-buttons/blob/19c89e3fe226836f8f98860382bb336149aa9f56/src/api/eligibility.js#L101)
- [src/payment-flows/native/socket.js => `getSDKSocketProps`](https://github.com/paypal/paypal-smart-payment-buttons/blob/19c89e3fe226836f8f98860382bb336149aa9f56/src/payment-flows/native/socket.js#L90) 

which both use belter's [`getUserAgent`](https://github.com/krakenjs/belter/blob/8cf35443d53cb81fcb7bce0bdb385aeaffc7423c/src/device.js#L4) which pulls off the window.

BriceShatzer: @vladfilyakov & @ghbhatt

vladfilyakov: Just confirmed: iOS Checkout SDK is reading `userAgent` from UL but not using it for anything - so free to remove.

BriceShatzer: Sounds good from @ghbhatt's side on android as well: 
Screen Shot 2021-07-20 at 12 08 07 PM

BriceShatzer: Sounds good from @ghbhatt's side on android as well: 
Screen Shot 2021-07-20 at 12 08 07 PM

PR #202, opened 2021-06-30
Vault setup for popup-bridge flow

leogedler:
**Reasoning**

Currently for the popup-bridge payment flow there is not check for vault eligibility nor vault setup logic.

To check if there is vault eligibility the JS SDK need to do a call to graphQL `GetFundingEligibility` and after checking possible eligibility it should enable it doing a call to graphQL `EnableVault`.

**Changes:**
- Abstracted vault check-setup logic from the checkout flow to be used on the `popup-bridge flow` as well.

_Example of the button now showing the pay and agree to vaulting_ 

Screen Shot 2021-07-06 at 10 55 09 AM


mnicpt:
@leogedler Is there any Eligibility changes needed in xobuyer for vaulting for this flow?

src/payment-flows/vault-setup.js
```diff
@@ -0,0 +1,130 @@
...
+
+    return ZalgoPromise.try(() => {
+        getLogger()
+            .info(`vault_auto_setup_vault_${ vault.toString() }_id_token_${ userIDToken ? 'present' : 'not_present' }`)
+            .flush();
+
+        return isVaultAutoSetupEligible({ vault, clientAccessToken, createBillingAgreement, createSubscription, fundingSource,
+            clientID, merchantID, buyerCountry, currency, commit, intent, disableFunding, disableCard });
+
+    }).then(eligible => {
```
mnicpt, line 121: How do we handle `eligible === false`?

leogedler:
> @leogedler Is there any Eligibility changes needed in xobuyer for vaulting for this flow?

@mnicpt not sure about that. I think we need some feedback from @bluepnume on this.
I have assumed that the eligibility is the same as for the checkout flow.

src/payment-flows/vault-setup.js
```diff
@@ -0,0 +1,130 @@
...
+
+        return getFundingEligibility(`
+                ${ fundingSource } {
+                    vaultable
+                }
+            `, {
+            accessToken, clientID, merchantID, buyerCountry, currency, commit, vault,
+            intent, disableFunding, disableCard
+        }).then(newFundingEligibility => {
+            if (newFundingEligibility[fundingSource] && newFundingEligibility[fundingSource].vaultable) {
```
mnicpt, line 76: Can't we just: `return newFundingEligibility[fundingSource] && newFundingEligibility[fundingSource].vaultable;`?

src/payment-flows/popup-bridge.js
```diff
@@ -48,27 +49,42 @@ function isPopupBridgePaymentEligible({ payment } : IsPaymentEligibleOptions) :
...
-                    redirect_uri: parentPopupBridge.nativeUrl,
-                    native_xo:    '1'
+
+            return ZalgoPromise.try(() => {
+                if (clientID && buyerIntent === BUYER_INTENT.PAY) {
+                    return enableVaultSetup({ orderID, vault, clientAccessToken, fundingEligibility, fundingSource, createBillingAgreement, createSubscription,
+                        clientID, merchantID, buyerCountry, currency, commit, intent, disableFunding, disableCard, userIDToken });
+                }
+            }).then(() => {
+                if (!parentPopupBridge) {
```
mnicpt, line 71: Is this what we do when `enableVaultSetup`is false?

src/payment-flows/vault-setup.js
leogedler: @mnicpt if `eligible === false` the promise `enableVaultSetup` gets resolved and we don't call `enableVault`.

`enableVault` is the GraphQL mutation that enables the vaulting option for the order.

```
export function enableVault({ orderID, clientAccessToken } : {| orderID : string, clientAccessToken : string |}) : ZalgoPromise<mixed> {
    return callGraphQL({
        name:  'EnableVault',
        query: `
            mutation EnableVault(
                $orderID : String!
            ) {
                enableVault(
                    token: $orderID
                )
            }
        `,
        variables: {
            orderID
        },
        headers: {
            [ HEADERS.ACCESS_TOKEN ]:   clientAccessToken,
            [ HEADERS.CLIENT_CONTEXT ]: orderID
        }
    });
}
```

src/payment-flows/vault-setup.js
leogedler: @mnicpt I tried your suggested approach but I got the following type error:

_**Cannot call `ZalgoPromise.try` with function bound to `method` because  `ZalgoPromise` [1] is incompatible with  boolean [2] in the return value.**_

src/payment-flows/popup-bridge.js
leogedler: @mnicpt no, actually this is coming from [here](https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/payment-flows/popup-bridge.js#L16) with is called to setup the popup-bridge flow.

This was already there before this vault implementation and it wasn't altered by it.

src/payment-flows/vault-setup.js
```diff
@@ -0,0 +1,130 @@
+/* @flow */
```
bluepnume, line 1: Let's not put this in `payment-flows/` -- payment flows are supposed to be full end-to-end handlers which approve a payment. This is more of a utility/middleware that applies to all of the existing payment flows.

src/payment-flows/popup-bridge.js
```diff
@@ -48,27 +49,42 @@ function isPopupBridgePaymentEligible({ payment } : IsPaymentEligibleOptions) :
...
-                query: {
-                    fundingSource,
-                    token:        orderID,
-                    useraction:   commit ? USER_ACTION.COMMIT : USER_ACTION.CONTINUE,
-                    redirect_uri: parentPopupBridge.nativeUrl,
-                    native_xo:    '1'
+
+            return ZalgoPromise.try(() => {
+                if (clientID && buyerIntent === BUYER_INTENT.PAY) {
+                    return enableVaultSetup({ orderID, vault, clientAccessToken, fundingEligibility, fundingSource, createBillingAgreement, createSubscription,
```
bluepnume, line 67: Rather than adding this to popup-bridge.js, we should really just make this call for all payment methods. We can probably do it in a common place like pay.js: https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/button/pay.js#L127

src/payment-flows/vault-setup.js
leogedler: @bluepnume I have moved this logic to a new `middleware` folder.

src/payment-flows/popup-bridge.js
leogedler: @bluepnume I have moved the logic to pay.js as suggested, Now it is being applied to all the payments flows.

src/button/pay.js
```diff
@@ -153,6 +162,7 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
 
             return ZalgoPromise.all([
                 clickPromise,
+                vaultPromise,
```
bluepnume, line 165: Could you actually sequence this before `validateOrderPromise` and `startPromise`? Want to make sure this is complete before we launch into any payment flows.

src/middleware/vault-setup.js
```diff
@@ -0,0 +1,130 @@
...
+
+        return getFundingEligibility(`
+                ${ fundingSource } {
+                    vaultable
+                }
+            `, {
+            accessToken, clientID, merchantID, buyerCountry, currency, commit, vault,
+            intent, disableFunding, disableCard
+        }).then(newFundingEligibility => {
+            if (newFundingEligibility[fundingSource] && newFundingEligibility[fundingSource].vaultable) {
```
mnicpt, line 76: @leogedler Do `return ZalgoPromise.resolve(newFundingEligibility[fundingSource] && newFundingEligibility[fundingSource].vaultable);`

src/payment-flows/vault-setup.js
mnicpt: @leogedler `enableVaultSetup` gets resolved either way unless there is an error.  So if it's false we just don't call `enableVault`.

src/middleware/vault-setup.js
```diff
@@ -0,0 +1,130 @@
...
+|};
+
+
+function isFundingSourceVaultable({ accessToken, fundingSource, clientID, merchantID, buyerCountry, currency, commit, vault, intent, disableFunding, disableCard } : IsFundingSourceVaultableOptions) : ZalgoPromise<boolean> {
+    return ZalgoPromise.try(() => {
+        if (fundingSource !== FUNDING.PAYPAL) {
+            return false;
+        }
+
+        return getFundingEligibility(`
```
mnicpt, line 68: @leogedler We should have access to fundingEligibility already.  Don't think we need to call this again.

src/button/pay.js
```diff
@@ -120,7 +122,14 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                     });
                 }).catch(noop);
 
-            const { intent, currency } = props;
+            const vaultPromise = createOrder().then(orderID => {
+                return ZalgoPromise.try(() => {
+                    if (clientID && buyerIntent === BUYER_INTENT.PAY) {
+                        return enableVaultSetup({ orderID, vault, clientAccessToken, fundingEligibility, fundingSource, createBillingAgreement, createSubscription,
```
mnicpt, line 128: @leogedler We pass in `fundingEligibility` here but don't accept it in `enableVaultSetup`.  If we do that, we don't need to call graphQL.

leogedler: As per @bluepnume comment, we still need to do the `graphQL` call.

src/middleware/vault-setup.js
leogedler: We still need to call with `accessToken` and re-check eligibility.

src/middleware/vault-setup.js
leogedler: I also tried this approach but still getting some flow errors

leogedler (commented):
@bluepnume I have modified the sequence to call `vaultPromise` before `validateOrderPromise` and `startPromise`.

src/button/pay.js
leogedler: @bluepnume I have added a sequence to call first `vaultPromise` then `validateOrderPromise` -> `startPromise`.

src/payment-flows/vault-setup.js
leogedler: Correct, if the `eligible` is false we don't enable the vaulting.

bluepnume (approved):

PR #203, opened 2021-07-01
Filter out Unsure SFVC for Native Flow

mnicpt:
Currently, if we detect popup in Safari View Controller (SFVC), then we flag this as native ineligible.  We also want to do this for form factors that we don't know (crossover data between Safari and SFVC).

### DIFF
--- a/src/native/popup/popup.js
+++ b/src/native/popup/popup.js
@@ -89,7 +89,7 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
         [FPTI_CUSTOM_KEY.INFO_MSG]: base64encode(window.location.href)
     }).flush();
 
-    const sfvc = isSFVC();
+    let sfvc = isSFVC();
     const sfvcOrSafari = !sfvc ? isSFVCorSafari() : false;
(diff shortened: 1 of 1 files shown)


src/native/popup/popup.js
```diff
@@ -287,6 +287,7 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
     const pageUrl = `${ window.location.href.split('#')[0] }#${  HASH.CLOSE }`;
 
     appInstalledPromise.then(app => {
+        sfvc = !sfvc ? sfvcOrSafari === true : true;
```
gabrielo91, line 290: just a nonblocking suggestion, using "positive logic" is easier to read, so, my recommendation:

`sfvc = sfvc? true : sfvcOrSafari === true;`

gabrielo91 (approved):

remotevision (approved):
👍

src/native/popup/popup.js
mnicpt: Agree. Initially did that but did the negative logic on purpose to give it focus as to my intent.

PR #204, opened 2021-07-01
Add generic API call logs

elizabethmv:
This PR adds generic instrumentation to the main `callRestAPI`, `callSmartAPI`, and `callGraphQL` functions handling the API requests.

Screen Shot 2021-07-02 at 10 28 19 AM

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -38,12 +38,12 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } : 
             const hasDetails = body.details && body.details.length;
             const issue = (hasDetails && body.details[0].issue) ? body.details[0].issue : 'Generic Error';
             const description = (hasDetails && body.details[0].description) ? body.details[0].description : 'no description';
-            
+
             const error = new Error(`${ issue }: ${ description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
             // $FlowFixMe
             error.response = { status, headers: responseHeaders };
 
-            getLogger().warn(`rest_api${ url.replace(/\//g, '_') }_error`);
+            getLogger().warn(`rest_api${ url.replace(/\//g, '_') }_error`, { err: issue });
             throw error;
         }
 
@@ -115,7 +115,7 @@ export function callGraphQL<T>({ name, query, variables = {}, headers = {} } : {
         if (errors.length) {
             const message = errors[0].message || JSON.stringify(errors[0]);
 
-            getLogger().warn(`graphql_${ name }_error`);
+            getLogger().warn(`graphql_${ name }_error`, { err: message });
             throw new Error(message);
         }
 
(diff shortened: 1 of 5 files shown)


mstuart:
Looks good! But, for the event names, could we take the endpoint (Ex: `/foo/bar`) and include it in the event name like `foo_bar_error`? That'll allow us to do some quick event counts in our dashboards without having to peek into the "payload" part to figure out what API it was.

src/api/api.js
```diff
@@ -37,11 +38,12 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
             const hasDetails = body.details && body.details.length;
             const issue = (hasDetails && body.details[0].issue) ? body.details[0].issue : 'Generic Error';
             const description = (hasDetails && body.details[0].description) ? body.details[0].description : 'no description';
-            
+
             const error = new Error(`${ issue }: ${ description } (Corr ID: ${ responseHeaders[HEADERS.PAYPAL_DEBUG_ID] }`);
             // $FlowFixMe
             error.response = { status, headers: responseHeaders };
 
+            getLogger().warn(`rest_api${ url.replace(/\//g, '_') }_error`, { err: issue });
```
mnicpt, line 46: @elizabethmv Wondering if we should create a helper function in logger.js in /lib to do `url.replace(/\//g, '_')`?

mnicpt: Maybe have a `formatSmartApiLog`, `fomatRestApiLog` and `formatGraphQlLog` that all pass in there specifics to a private generic one.

mnicpt: Actually remove `format` from those names.

elizabethmv: Yeah, I can pull this into a helper 👍 

It's just taking `/some/url` and doing `_some_url` to include it into the event name. Something like `slashToUnderscore(url)`?

elizabethmv: Updated! @mnicpt please re-review when you get a chance

elizabethmv:
hi team, anyone free to review? thanks! @paypal/checkout-sdk

src/lib/logger.js
```diff
@@ -93,3 +93,7 @@ export function setupLogger({ env, sessionID, clientID, sdkCorrelationID, buyerC
         logger.flush().catch(noop);
     });
 }
+
+export function slashToUnderscore(endpoint : string) : string {
```
gregjopa, line 97: Could we move this to `src/lib/util.js`? Seems more like a generic utility function.

gregjopa (approved):
Looks great! Nice work @elizabethmv 💯🥇👍

mstuart:
lgtm!

mstuart (approved):

src/lib/logger.js
elizabethmv: Great suggestion, updated 👍

PR #205, opened 2021-07-06
Test out mobile.paypal.com, behind a client-side experiment

andregardi:
### Test out mobile.paypal.com, behind a client-side experiment
Ticket: [DTCHKINT-884](https://engineering.paypalcorp.com/jira/browse/DTCHKINT-884)

For Native Checkout, today we open up history.paypal.com in the popup, but we want to test out mobile.paypal.com instead. We're not sure how mobile.paypal.com performs, so we need to wrap the change in a client-side experiment using createExperiment(), then we'll ramp mobile.paypal.com to 1%.


mstuart (approved):

src/payment-flows/native/url.js
```diff
@@ -33,7 +33,18 @@ export function getNativePopupDomain({ props } : {| props : ButtonProps |}) : st
...
+    };
+    const mobile_native_popup_domain_experiment = createExperiment('enable_mobile_native_popup_domain', experimentOptions);
+
+    const is_mobile_native_popup_domain_enabled = mobile_native_popup_domain_experiment.isEnabled();
+
+    const native_popup_domain = is_mobile_native_popup_domain_enabled ?
+        HISTORY_NATIVE_POPUP_DOMAIN :
+        MOBILE_NATIVE_POPUP_DOMAIN;
+
+    return native_popup_domain[env];
```
mstuart, line 47: Can you convert these to camelCase? Just so it follows the code style in the rest of the file.

andregardi: Sure, you are right! It is done!

src/payment-flows/native/url.js
```diff
@@ -33,7 +33,16 @@ export function getNativePopupDomain({ props } : {| props : ButtonProps |}) : st
         return 'https://www.sandbox.paypal.com';
     }
 
-    return NATIVE_POPUP_DOMAIN[env];
+    const experimentOptions = {
+        sample: 1
```
gregjopa, line 37: @mstuart @andregardi what do you think about having this at 0% to start? If we do 0% than it reduces risk while we verify it end to end in prod using the local storage opt-in:

```js
window.localStorage.setItem('enable_mobile_native_popup_domain', true);
```

Just thinking about migrating risk incase there's a problem with this logic.

```suggestion
        sample: 0
```

mstuart: I think that's a great idea. Especially because how sensitive this is :o

mstuart: We can quickly verify it's all working, then do a 1% release afterwards.

andregardi: Great suggestion @gregjopa! 
I just changed it to 0% now.

gregjopa (approved):
Looks good!

mstuart:
lgtm!

mstuart (approved):

PR #208, opened 2021-07-15
Add Logging for SPB Rate Limiting

andregardi:
### Problem
Buyers who repeatedly tap a Smart Payment Button are blocked via rate limiting
This policy isn't necessarily a problem but we chased our tail for days hunting this problem down due to limited logging.

### Solution
Add logs for this rate limiting error so native engineers and product analysts can track incidences of rate limiting


src/api/api.js
```diff
@@ -44,6 +45,14 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers } :
             error.response = { status, headers: responseHeaders };
 
             getLogger().warn(`rest_api${ slashToUnderscore(url) }_error`, { err: issue });
+
+            if (status === STATUS_CODES.TOO_MANY_REQUESTS) {
+                getLogger().track({
+                    [FPTI_KEY.TRANSITION]:      FPTI_TRANSITION.CALL_REST_API,
+                    [FPTI_CUSTOM_KEY.ERR_DESC]: `Error: ${ status } - ${ body }`,
```
mstuart, line 52: Is body an object? Wonder if it'd look like `[object Object]`. Maybe JSON.stringify it? Or we could maybe go without logging the body since it'll be logged in CAL.

andregardi: @mstuart ,
It seems that body is not an object in the particular case:
https://github.paypal.com/Security-R/ratelimiter/blob/d79542fe3cc82a47834d5c89a075049f1cddd1d1/lib/utils.js#L245

The `ratelimiter` middleware, when facing **429** errors, responds with a string `Too many requests ` instead of an object.

mstuart:
@andregardi tested this locally and confirmed that the FPTI calls are being made. nice work! 👍

PR #212, opened 2021-07-29
Enforce eventName param for API logs

elizabethmv:
This PR adds generic instrumentation to the main `callRestAPI`, `callSmartAPI`, and `callGraphQL` functions handling the API requests by enforcing an eventName param.

**Before:**

This was originally introduced in #204 but was reverted as the generic catch all using `url` was creating too many unique event names such as `rest_api_v2_checkout_orders_ABC123_capture_error` which introduces issues with event counts and such.

**After:**

Screen Shot 2021-08-02 at 3 13 20 PM


src/api/order.js
```diff
@@ -78,6 +79,7 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
         return callRestAPI({
             accessToken: facilitatorAccessToken,
             url:         `${ ORDERS_API_URL }/${ orderID }`,
+            eventName:   `v2_checkout_orders_get`,
```
gregjopa, line 82: I noticed we are doing this else where in this file, but I think we should stick with single quotes and only use string interpolation when we have variables in the string.
```suggestion
            eventName:   'v2_checkout_orders_get',
```

src/api/api.js
```diff
@@ -52,15 +55,16 @@ type SmartAPIRequest = {|
     url : string,
     method? : string,
     json? : $ReadOnlyArray<mixed> | Object,
-    headers? : { [string] : string }
+    headers? : { [string] : string },
+    eventName : string
```
gregjopa, line 59: I like that eventName is a required string. This approach looks good to me!

elizabethmv: I like this approach too, it was a great suggestion from @bluepnume 😄

gregjopa (approved):

PR #215, opened 2021-08-03
Fix Venmo Desktop Eligibility

mnicpt:
Venmo Desktop requires part of native eligibility check except for User Agent checks w.r.t. being on a mobile device.


TNuzzi (approved):
Good call on keeping the eligibility.  I suspected we would need to bring it back at some point.  We will need to create another eligibility specifically for desktop in the future.

PR #217, opened 2021-08-10
QR Code Logic Fix

mnicpt:
* Current QR code logic would let everything through on second eligibility call.  Only want to let `isUserAgentEligible` and `isBrowserMobileAndroid` through.

* Enable Amplitude in popup.

### DIFF
--- a/src/payment-flows/native/qrcode.js
+++ b/src/payment-flows/native/qrcode.js
@@ -45,15 +45,16 @@ function getEligibility({ fundingSource, props, serviceData, validatePromise } :
                 skipElmo:   true
             }).then(eligibility => {
                 // ignore isUserAgentEligible and isBrowserMobileAndroid for Venmo Desktop as they don't apply
-                if (
-                    !eligibility &&
-                    !eligibility[fundingSource] &&
-                    !eligibility[fundingSource].eligibility &&
-                    eligibility[fundingSource].ineligibilityReason &&
-                    eligibility[fundingSource].ineligibilityReason.length &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isUserAgentEligible') === -1 &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isBrowserMobileAndroid') === -1
+                const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
+                const ineligibleReasons = eligibility && eligibility[fundingSource]?.ineligibilityReason?.split(',');
+                const ineligible = ineligibleReasons?.reduce((acc, reason) => {
+                    acc = acc || eligibleReasons.indexOf(reason) === -1;
+                    return acc;
+                }, false);
 
+                if (
+                    ineligibleReasons &&
+                    ineligible
                 ) {
                     getLogger().info(`native_appswitch_ineligible`, { orderID })
                         .track({
--- a/src/payment-flows/native/qrcode.js
+++ b/src/payment-flows/native/qrcode.js
@@ -48,7 +48,9 @@ function getEligibility({ fundingSource, props, serviceData, validatePromise } :
                 const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
                 const ineligibleReasons = eligibility && eligibility[fundingSource]?.ineligibilityReason?.split(',');
 
-                const ineligible = ineligibleReasons?.reduce((acc, reason) => {
+                const ineligible = ineligibleReasons?.every(reason => {
+                    return eligibleReasons.indexOf(reason) === -1;
+                });
                     acc = acc || eligibleReasons.indexOf(reason) === -1;
                     return acc;
                 }, false);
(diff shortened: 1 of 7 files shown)


src/payment-flows/native/qrcode.js
```diff
@@ -45,15 +45,17 @@ function getEligibility({ fundingSource, props, serviceData, validatePromise } :
...
-                    !eligibility[fundingSource] &&
-                    !eligibility[fundingSource].eligibility &&
-                    eligibility[fundingSource].ineligibilityReason &&
-                    eligibility[fundingSource].ineligibilityReason.length &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isUserAgentEligible') === -1 &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isBrowserMobileAndroid') === -1
+                const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
+                const ineligibleReasons = eligibility && eligibility[fundingSource]?.ineligibilityReason?.split(',');
+
+                const ineligible = ineligibleReasons?.reduce((acc, reason) => {
```
gregjopa, line 51: What do you think about using Array.prototype.every instead?
```suggestion
                const ineligible = ineligibleReasons?.every(reason => {
                    return eligibleReasons.indexOf(reason) === -1;
                });
```

test/client/native-qrcode.js
```diff
@@ -149,4 +147,191 @@ describe('native qrcode cases', () => {
...
     });
+
+    it('should render a button with createOrder, click the button, and render checkout via qrcode path with eligibility disregard for isUserAgentEligible or isBrowserMobileAndroid', async () => {
+        return await wrapPromise(async ({ expect, avoid }) => {
+            window.xprops.platform = PLATFORM.MOBILE;
+            delete window.xprops.onClick;
+
+            const sessionToken = uniqueID();
+
+            const gqlMock = getGraphQLApiMock({
```
gregjopa, line 158: Thanks for adding this test case! It really helps with understanding the desktop eligibility logic.

test/client/native-qrcode.js
```diff
@@ -149,4 +147,191 @@ describe('native qrcode cases', () => {
...
+                extraHandler: expect('extraHandler', ({ message_name, message_type }) => {
+                    if (message_name === 'onInit' && message_type === 'request') {
+                        ZalgoPromise.delay(50).then(onApprove);
+                    }
+                })
+            });
+
+            const mockWebSocketServer = expectSocket();
+
+            const orderID = generateOrderID();
```
gregjopa, line 222: It looks like the `generateOrderID` mock function needs to be imported in.

mnicpt: Interesting.  All of the tests were passing locally.  Checking.

src/payment-flows/native/qrcode.js
```diff
@@ -45,15 +45,19 @@ function getEligibility({ fundingSource, props, serviceData, validatePromise } :
...
                 // ignore isUserAgentEligible and isBrowserMobileAndroid for Venmo Desktop as they don't apply
-                if (
-                    !eligibility &&
-                    !eligibility[fundingSource] &&
-                    !eligibility[fundingSource].eligibility &&
-                    eligibility[fundingSource].ineligibilityReason &&
-                    eligibility[fundingSource].ineligibilityReason.length &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isUserAgentEligible') === -1 &&
-                    eligibility[fundingSource].ineligibilityReason.indexOf('isBrowserMobileAndroid') === -1
+                const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
```
bluepnume, line 48: This totally makes sense, and thanks for catching this while I was out!

That said -- can we plan to incorporate this fix into the GQL layer itself? My concern is we will be logging one decision on the server/gql, and making another decision on the client. Would be better to keep as much of the eligibility as possible in one place.

mnicpt: @bluepnume I see this as temporary and really think we should create a separate Funding Source for Venmo Desktop so that when we make changes to native we don't accidentally break desktop or need to have these kinds of if-conditions.

bluepnume: You mean a separate funding source in the GQL eligibility api?

mnicpt: Yes.

test/client/native-qrcode.js
mnicpt: @gregjopa Also noticed that it is being imported.  This is really odd.

src/payment-flows/native/qrcode.js
bluepnume: OK, let's discuss. I prefer to use parameters wherever possible, otherwise you end up with an insane number of enums :) e.g. `venmo` evolves to `venmo_desktop` evolves to `venmo_desktop_us_loggedinuser` etc.

PR #224, opened 2021-08-23
fix(analyics): Add ineligible reasons to app_switch_ineligible tracking

TNuzzi:


src/payment-flows/native/popup.js
```diff
@@ -85,11 +85,14 @@ function getEligibility({ fundingSource, props, serviceData, sfvc, validatePromi
                 merchantID:   merchantID[0],
                 domain:       merchantDomain
             }).then(eligibility => {
+                const ineligibilityReason = eligibility && eligibility[fundingSource] ? eligibility[fundingSource].ineligibilityReason : '';
```
mnicpt, line 88:
```suggestion
                const ineligibilityReason = eligibility && eligibility[fundingSource] && eligibility[fundingSource].ineligibilityReason ? eligibility[fundingSource].ineligibilityReason : '';
```

mnicpt (approved):

mnicpt (changes requested):
Need to condition eligibility on Venmo for the ineligibleReasons filtering.  So you want to have a case for PayPal, which would just be the result of `nativeEligibilityResults[fundingSource]?.eligble` and Venmo being what you have.

src/payment-flows/native/native.js
```diff
@@ -26,10 +27,14 @@ function setupNative({ props, serviceData } : SetupOptions) : ZalgoPromise<void>
...
 function initNative({ props, components, config, payment, serviceData } : InitOptions) : PaymentFlowInstance {
-    const { onApprove, onCancel, onError,
+    const { env, clientID, sdkCorrelationID, locale, sessionID, onApprove, onCancel, onError,
         buttonSessionID, onShippingChange } = props;
     const { fundingSource } = payment;
+    const { buyerCountry } = serviceData;
     const { firebase: firebaseConfig } = config;
+    const sdkVersion = getSDKVersion();
+
+    setupLogger({ env, sessionID, clientID, sdkCorrelationID, locale, sdkVersion, buyerCountry, fundingSource });
```
bluepnume, line 37: Why do we need to setup logger again? This should be happening when the buttons first render right?

src/payment-flows/native/eligibility.js
```diff
@@ -184,6 +188,20 @@ export function isNativePaymentEligible({ payment } : IsPaymentEligibleOptions)
         return false;
     }
 
+    const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
+    const ineligibleReasons = nativeEligibilityResults && nativeEligibilityResults[fundingSource]?.ineligibilityReason?.split(',');
+
+    const eligible = ineligibleReasons?.every(reason => {
```
bluepnume, line 194: Should we limit this check to only desktop mode, for now?

src/payment-flows/native/native.js
TNuzzi: @bluepnume We found an instance where `fundingSource` was not being set when we wanted to log it.  It was in the `app_switch_ineligible`.  @mnicpt traced it back to trying to log it before the `fundingSource` was set.

@mnicpt Did I get that right?

src/payment-flows/native/eligibility.js
TNuzzi: @mnicpt has the details on this.  I pointed out to him that we were seeing `isBrowserMobileAndroid` and `isUserAgentEligible` on an iPad.

TNuzzi: I think this code is temporary until @mnicpt did some refactoring in xobuyer.

mnicpt: @TNuzzi Let's do what @bluepnume suggested.  This will keep this contained to desktop which is what we really want anyways.

src/payment-flows/native/native.js
mnicpt: @TNuzzi Instead of calling `setupLogger` again, let's add a tracking builder after `click`.  Spoke with @bluepnume  last night and he suggested.  Seems like the best option for handling this case.

src/payment-flows/native/eligibility.js
```diff
@@ -184,6 +189,23 @@ export function isNativePaymentEligible({ payment } : IsPaymentEligibleOptions)
         return false;
     }
 
+    // For Venmo desktop, ignore failing eligibility if the given reasons are returned from NativeEligibility
+    if (platform && platform === PLATFORM.DESKTOP) {
+        const eligibleReasons = [ 'isUserAgentEligible', 'isBrowserMobileAndroid' ];
+        const ineligibleReasons = nativeEligibilityResults && nativeEligibilityResults[fundingSource]?.ineligibilityReason?.split(',');
```
mnicpt, line 195:
```suggestion
        const ineligibleReasons = nativeEligibilityResults && nativeEligibilityResults[fundingSource]?.ineligibilityReason ? nativeEligibilityResults[fundingSource].ineligibilityReason.split(',') : '';
```

mnicpt: If `nativeEligibilityResults[fundingSource]?.ineligibilityReason` is empty, then eligible will be null which will be false and return false for eligibility.

src/payment-flows/native/eligibility.js
```diff
@@ -184,6 +189,23 @@ export function isNativePaymentEligible({ payment } : IsPaymentEligibleOptions)
...
+
+        const eligible = ineligibleReasons?.every(reason => {
+            return eligibleReasons?.indexOf(reason) !== -1;
```
mnicpt, line 198:
```suggestion
            return reason ? eligibleReasons?.indexOf(reason) !== -1 : true;
```

mnicpt: Please do this to the other one too.

mnicpt (approved):

PR #226, opened 2021-08-23
Personalization analytic time out

erick2014:


server/service/personalization.jsx
```diff
@@ -172,6 +173,12 @@ export async function resolvePersonalization(req : ExpressRequest, gqlBatch : Gr
         return personalization;
 
     } catch (err) {
+        if (err.toString().includes('Timed out after')) {
```
gregjopa, line 176: I worry about checking for this hard-coded error message. Could we move it to a constant and share it with promiseTimeout()? https://github.com/paypal/paypal-smart-payment-buttons/blob/1af8a7444acb6dc0ac911c869c9ff035061b37c9/server/lib/util.js#L240

erick2014: sounds good!

erick2014: @gregjopa Hey I just added the constant but I had to assign a function to it, in order to receive the time value

server/config/constants.js
```diff
@@ -35,3 +35,5 @@ export const EVENT = {
 export const AUTH_ERROR_CODE = {
     INVALID_CLIENT: 'invalid_client'
 };
+
+export const TIMEOUT_ERROR_MESSAGE = (time : number) : string => `Timed out after ${ time }ms`;
```
gregjopa, line 39: Could we move this function into server/lib/util.js and have it be exported from there? Also since this is a function, we should not uppercase the name since it's not a constant static value.

ssono: My question is whether this needs to be a constant in the first place? It is being used twice and since it isn't formatted correctly without creating a function it feels like an unnecessary abstraction to me

server/service/personalization.jsx
```diff
@@ -172,6 +173,12 @@ export async function resolvePersonalization(req : ExpressRequest, gqlBatch : Gr
         return personalization;
 
     } catch (err) {
+        if (err.toString().includes(TIMEOUT_ERROR_MESSAGE(PERSONALIZATION_TIMEOUT))) {
```
ssono, line 176: Is there a better way to check if this was a timeout failure than converting the error to a string? Might be worth looking at the other attributes of the err object to see if there is something cleaner. 

If there isn't then I wonder if we need to match the PERSONALIZATION_TIMEOUT value for the timeout. Unless we can 100% guarantee that that will be the only possible timeout message. Could we check to see if it includes `Timed out after` or regex /Timed out after [0-9]+ms/

ssono (commented):
Left some comments. Also, you should check with Partheeban Prabhakaran from analytics to see if this timeout will be sufficient for tracking.

server/config/constants.js
erick2014: Hey I just changed this constant to just hold the value `Timed out after`

server/service/personalization.jsx
erick2014: I changed this as well

server/config/constants.js
gregjopa:
```js
export const TIMEOUT_ERROR_MESSAGE = `Timed out after`;
```

I like this approach. Looks good! 👍

server/config/constants.js
```diff
@@ -35,3 +35,5 @@ export const EVENT = {
 export const AUTH_ERROR_CODE = {
     INVALID_CLIENT: 'invalid_client'
 };
+
+export const TIMEOUT_ERROR_MESSAGE = `Timed out after`;
```
gregjopa, line 39: Nit: let's use single quotes

```suggestion
export const TIMEOUT_ERROR_MESSAGE = 'Timed out after';
```

erick2014: pushed!

ssono (approved):
LGTM 🐙

server/types.js
```diff
@@ -22,7 +22,8 @@ export type LoggerType = {|
     debug : (req : ExpressRequest, event : string, payload : ?LoggerPayload) => void,
     info : (req : ExpressRequest, event : string, payload : ?LoggerPayload) => void,
     warn : (req : ExpressRequest, event : string, payload : ?LoggerPayload) => void,
-    error : (req : ExpressRequest, event : string, payload : ?LoggerPayload) => void
+    error : (req : ExpressRequest, event : string, payload : ?LoggerPayload) => void,
+    track : (req : ExpressRequest, event : Object, payload : ?LoggerPayload) => void
```
sioked, line 26: Was just digging into this recently. Technically the function signature for `track` is:

```javascript
track : (req: ExpressRequest, payload: LoggerPayload, meta: ?Object) => void
```

Fundamentally it works exactly the same as what you have here, but in all other logging methods the event is a string where the argument for tracking is a complex object (the payload). I think the mismatch of types could be somewhat confusing for another developer in the future to use this method. Just a nit though!

PR #230, opened 2021-08-30
Increase experiment and wallet timeouts for non-prod environments

rjzheng:
Need to increase Wallet and Experiment (Elmo) timeouts for non-prod environments


server/config/config.js
```diff
@@ -42,8 +42,8 @@ export const SDK_CDN_NAMESPACE = 'js-sdk-release';
 export const SMART_BUTTONS_CDN_NAMESPACE = 'smart-payment-buttons';
 
 export const FUNDING_ELIGIBILITY_TIMEOUT = 200;
-export const WALLET_TIMEOUT = 2000;
-export const EXPERIMENT_TIMEOUT = 100;
+export const WALLET_TIMEOUT : number = getEnv() === 'production' ? 2000 : 200000;
```
bluepnume, line 45: Can we make this an object mapping of `{ env: timeout }`? And can we also use the `ENV` constant rather than hard-coding `'production'` as a string? Thanks!

Also -- 200,000 means we will wait for more than 3 minutes for the button to render before we time out. Recommend going up to 10,000 for now. Else this will be un-usable on stage/sandbox if there is some stalled service.

server/config/config.js
```diff
@@ -42,8 +42,20 @@ export const SDK_CDN_NAMESPACE = 'js-sdk-release';
 export const SMART_BUTTONS_CDN_NAMESPACE = 'smart-payment-buttons';
 
 export const FUNDING_ELIGIBILITY_TIMEOUT = 200;
-export const WALLET_TIMEOUT : number = getEnv() === 'production' ? 2000 : 200000;
-export const EXPERIMENT_TIMEOUT : number = getEnv() === 'production' ? 1000 : 100000;
+export const WALLET_TIMEOUT : number = {
+    production: 2000,
```
bluepnume, line 46: please use `ENV` const :)

rjzheng: Ah whoops, updated!

PR #234, opened 2021-09-02
Allow for variants to be carried over from initial fundingEligibility

ssono:
Related to https://github.com/paypal/paypal-checkout-components/pull/1725 and https://github.paypal.com/Checkout-R/clientsdknodeweb/pull/285

In order for the variant based eligibility to carry over it has to be preserved by SPB middleware

### DIFF
--- a/server/components/buttons/params.js
+++ b/server/components/buttons/params.js
@@ -184,6 +184,11 @@ function getFundingEligibilityParam(req : ExpressRequest) : FundingEligibilityTy
                             productEligibility.eligible = productEligibilityInput.eligible;
                             productsEligibility[product] = productEligibility;
                         }
+
+                        if (typeof productEligibilityInput.variant === 'string') {
+                            productEligibility.variant = productEligibilityInput.variant;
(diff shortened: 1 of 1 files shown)


server/components/buttons/params.js
```diff
@@ -184,6 +184,11 @@ function getFundingEligibilityParam(req : ExpressRequest) : FundingEligibilityTy
                             productEligibility.eligible = productEligibilityInput.eligible;
                             productsEligibility[product] = productEligibility;
                         }
+
+                        if (typeof productEligibilityInput.variant === 'string') {
```
westeezy, line 188: Excuse my ignorance to this, but what are the possible types for variant? I am wondering if this should just be a truthiness check vs `typeof string`.

ssono: Yes it is an optional string in type FundingProductEligibility

ssono:
Tests pass on local
Screen Shot 2021-09-07 at 1 54 00 PM

gregjopa (approved):

PR #235, opened 2021-09-02
Fix bug on the iOS opt-out flow.

leogedler:
**Reasoning**
After an initial test from the iOS team we found that the opt-out logic wasn't working as expected for the iOS Venice app.

**Changes**
In this PR I'm fixing an issue with the payload data object when passing this payload received from the `fallback` event back to its parent window via `onPostMessage` (postRobot).

I also did a refactor on the existing native opt-out logic.

### DIFF
--- a/src/payment-flows/native/eligibility.js
+++ b/src/payment-flows/native/eligibility.js
@@ -6,7 +6,7 @@ import { supportsPopups, isIos, isAndroid } from 'belter/src';
 
 import { type NativeEligibility, getNativeEligibility } from '../../api';
 import { enableAmplitude, getStorageState, isIOSSafari, isAndroidChrome } from '../../lib';
-import { LSAT_UPGRADE_EXCLUDED_MERCHANTS, FPTI_TRANSITION } from '../../constants';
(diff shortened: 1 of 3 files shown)


src/constants.js
```diff
@@ -72,7 +72,8 @@ export const USER_EXPERIENCE_FLOW = {
 };
 
 export const PRODUCT_FLOW = {
-    SMART_PAYMENT_BUTTONS: 'SMART_PAYMENT_BUTTONS'
+    SMART_PAYMENT_BUTTONS: 'SMART_PAYMENT_BUTTONS',
+    NATIVE_OPT_OUT:        'native_opt_out'
```
mnicpt, line 76: This will not be used anywhere except in the native flow.  Can you move this constant to be local to the file where this is used?

mnicpt (changes requested):
LGTM except for small change.

src/constants.js
leogedler: @mnicpt sure!, I'll move it.

mnicpt (approved):

PR #238, opened 2021-09-08
Feature/qr venmo escape path

fedperezp:
Adding escape path for Venmo QR on Desktop
https://venmoinc.atlassian.net/browse/VMORECOMBO-173

Linked PRs
https://github.com/paypal/paypal-checkout-components/pull/1734


src/qrcode/qrcard.jsx
```diff
@@ -54,6 +55,11 @@ function QRCard({
         return state === QRCODE_STATE.ERROR;
     };
 
+    const handleClick = (event : string) => {
+        window.xprops.hide();
+        window.xprops.onEscapePath(event);
```
mnicpt, line 60: Why are you calling this `event`?  Looks like you're passing in funding instrument.

src/qrcode/qrcard.jsx
```diff
@@ -54,6 +55,11 @@ function QRCard({
         return state === QRCODE_STATE.ERROR;
     };
 
+    const handleClick = (event : string) => {
+        window.xprops.hide();
```
mnicpt, line 59: Are we wanting to `hide` or `close` here?  Do we plan on re-showing?

fedperezp: we don't plan on re-showing, I need to reuse the payment so I was worried that xprops.close would cancel the payment. I'm going to test with close() if you think is better in this case

fedperezp: Hey Steve, `close` is going the cancel the switch I'm making to the other payment flow once I go into an escape path so I think `hide` is the only alternative for now

mnicpt: @fedperezp So why not have the switch to the other payment return a Promise and then close?

PR #240, opened 2021-09-09
feat: log and track onError checkout calls

jscheinhorn:
Track onError calls to determine flow interaction of proposed onError changes:

`onError: (err) => {
                // eslint-disable-next-line no-use-before-define
                return close().then(() => {
                    return onError(err);
                });
            },`

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -113,6 +113,7 @@ export const FPTI_TRANSITION = {
     CHECKOUT_SHIPPING_CHANGE: ('process_checkout_shipping_change' : 'process_checkout_shipping_change'),
     CHECKOUT_APPROVE:         ('process_checkout_approve' : 'process_checkout_approve'),
     CHECKOUT_CANCEL:          ('process_checkout_cancel' : 'process_checkout_cancel'),
+    CHECKOUT_ERROR:           ('process_checkout_error' : 'process_checkout_error'),
 
     TOKENIZE_APPROVE:         ('process_tokenize_approve' : 'process_tokenize_approve'),
 
--- a/src/payment-flows/checkout.js
+++ b/src/payment-flows/checkout.js
@@ -265,18 +265,13 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
             },
 
             onError: (err) => {
-                return createOrder().then(orderID => {
-                    getLogger()
-                        .info('on_error_tracking')
-                        .track({
-                            [FPTI_KEY.TRANSITION]:   FPTI_TRANSITION.CHECKOUT_ERROR,
-                            [FPTI_KEY.CONTEXT_TYPE]: FPTI_CONTEXT_TYPE.ORDER_ID,
-                            [FPTI_KEY.TOKEN]:        orderID,
-                            [FPTI_KEY.CONTEXT_ID]:   orderID,
-                            [FPTI_KEY.ERROR_DESC]:   JSON.stringify(err)
-                        }).flush();
-                    return onError(err);
-                });
+                getLogger()
+                    .info(`checkout_flow_error `, { err: stringifyError(err) })
+                    .track({
+                        [FPTI_KEY.TRANSITION]:   FPTI_TRANSITION.CHECKOUT_ERROR,
+                        [FPTI_KEY.ERROR_DESC]:   stringifyError(err)
+                    }).flush();
+                return onError(err);
             },
 
             dimensions: getDimensions(fundingSource),
--- a/src/payment-flows/checkout.js
+++ b/src/payment-flows/checkout.js
@@ -264,7 +264,21 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
                 }
             },
 
-            onError,
+            onError: (err) => {
+                return createOrder().then(orderID => {
+                    getLogger()
+                        .info('on_error_tracking')
+                        .track({
+                            [FPTI_KEY.TRANSITION]:   FPTI_TRANSITION.CHECKOUT_ERROR,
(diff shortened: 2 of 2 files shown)


jscheinhorn:
Example tracking payload for onError call for sample order here: https://fpti-5x-3.pp-devcos-searchserv.us-central1.gcp.dev.paypalinc.com:5601/goto/53dcb4cd3d90c71da378b76ed5210539 


Using unique identifier for onError calls:
`transition_name`:`process_checkout_error `

westeezy (approved):
lgtm. I was going to mention maybe adding a test but given it'd be mostly mocks and we have a few passing tests [here](https://github.com/paypal/paypal-smart-payment-buttons/blob/1af8a7444acb6dc0ac911c869c9ff035061b37c9/test/client/error.js) I think this should be fine.

mnicpt (approved):

mnicpt:
@jscheinhorn I think this callback can be called prior to creating an order. If so, we don't want to call `createOrder`.

src/payment-flows/checkout.js
```diff
@@ -240,7 +240,20 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
                 }
             },
 
-            onError,
+            onError: (err) => {
+                return createOrder().then(orderID => {
+                    getLogger()
+                        .info('on_error_tracking')
```
bluepnume, line 246: Let's call this event `checkout_flow_error` `{ err: stringifyError(err) }` just to be a bit more specific.

src/payment-flows/checkout.js
```diff
@@ -240,7 +240,20 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
...
+                        .track({
+                            [FPTI_KEY.TRANSITION]:   FPTI_TRANSITION.CHECKOUT_ERROR,
+                            [FPTI_KEY.CONTEXT_TYPE]: FPTI_CONTEXT_TYPE.ORDER_ID,
+                            [FPTI_KEY.TOKEN]:        orderID,
+                            [FPTI_KEY.CONTEXT_ID]:   orderID,
+                            [FPTI_KEY.ERROR_DESC]:   JSON.stringify(err)
```
bluepnume, line 252: Please use `stringifyError` instead of `JSON.stringify`

mnicpt (commented):
Left a comment...

mnicpt (changes requested):
Left a comment...

src/payment-flows/checkout.js
jscheinhorn: should we forego attempting to get the order ID (currently from createOrder) because of the above scenario (onError called in the event createOrder fails)?

src/constants.js
```diff
@@ -111,6 +111,7 @@ export const FPTI_TRANSITION = {
     CHECKOUT_SHIPPING_CHANGE: ('process_checkout_shipping_change' : 'process_checkout_shipping_change'),
     CHECKOUT_APPROVE:         ('process_checkout_approve' : 'process_checkout_approve'),
     CHECKOUT_CANCEL:          ('process_checkout_cancel' : 'process_checkout_cancel'),
+    CHECKOUT_ERROR:           ('process_checkout_error' : 'process_checkout_error'),
```
jscheinhorn, line 114: change this too?
`CHECKOUT_ERROR:           ('checkout_flow_error' : 'checkout_flow_error'),`

src/payment-flows/checkout.js
gregjopa: What do you think about wrapping `createOrder` in `Promise.try()` so the Promise never rejects? Ex:

```js
ZalgoPromise.try(createOrder)
    .then((orderID = 'unknown') => {
        getLogger()
        .info('on_error_tracking')
            .track({
                [FPTI_KEY.TRANSITION]:   FPTI_TRANSITION.CHECKOUT_ERROR,
                [FPTI_KEY.CONTEXT_TYPE]: FPTI_CONTEXT_TYPE.ORDER_ID,
                [FPTI_KEY.TOKEN]:        orderID,
                [FPTI_KEY.CONTEXT_ID]:   orderID,
                [FPTI_KEY.ERROR_DESC]:   JSON.stringify(err)
            }).flush();
        
        return onError(err);
    })
```

bluepnume: > should we forego attempting to get the order ID (currently from createOrder) because of the above scenario (onError called in the event createOrder fails)?

Yeah I'd be happy to forego.

> What do you think about wrapping createOrder in Promise.try() so the Promise never rejects? Ex:

Quick note -- `Promise.try()` won't avoid a rejected promise. The only way to do that is to `.catch()` (and not re-throw)

jscheinhorn:
checkout_flow_error

src/constants.js
jscheinhorn: ^ @bluepnume @mnicpt

src/payment-flows/checkout.js
```diff
@@ -240,7 +240,15 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
                 }
             },
 
-            onError,
+            onError: (err) => {
+                getLogger()
+                    .info(`checkout_flow_error ${ { err: stringifyError(err) } }`)
```
bluepnume, line 245:
```
.info(`checkout_flow_error`, { err: stringifyError(err) }`)
```

bluepnume, line 245:
```suggestion
                    .info(`checkout_flow_error`, { err: stringifyError(err) }`)
```

src/constants.js
jscheinhorn: kept as `CHECKOUT_ERROR:           ('process_checkout_error' : 'process_checkout_error'),`

gregjopa (approved):

PR #241, opened 2021-09-09
Add `publish` GitHub Actions workflow

nbierdeman:
### Purpose

This PR adds a `publish` GitHub Actions workflow to `paypal-smart-payment-buttons`.


nbierdeman:
@gregjopa @westeezy, I'm not entirely sure if the `npm run reinstall` step in `publish.yml` will install `devDependencies`, and am therefore not sure if the `npm run build-commit` step in `npm run release` will fail. Do either of you know, or know how to test this prior to merging?

westeezy (approved):
lgtm to me thanks for adding this. Let's ensure we also update the internal release docs. cc @TNuzzi - I'd be happy to update those if this all looks good to you as well

.github/workflows/publish.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: ⎔ Setup node
+        # sets up the .npmrc file to publish to npm
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: 📥 Download deps
+        run: npm run reinstall
```
gregjopa, line 31: I don't think we need to use the `reinstall` task in SPB. Instead, let's try this:

```
      - name: 📥 Download deps
        uses: bahmutov/npm-install@v1
        with:
          useLockFile: false
```

gregjopa (approved):
Looks great! 💯

PR #242, opened 2021-09-09
WIP: POC card-fields

leogedler:
- Added support for prop `style` and `placeholder` for card field.
- Created some card-utils to handle masking and basic validations.
- Added logic to support custom styles for the inputs.


src/card/props.js
```diff
@@ -17,23 +18,21 @@ export type PrerenderDetailsType = {|
...
-|};
-
 export type CardExport = ({|
     submit : () => ZalgoPromise<void>
 |}) => ZalgoPromise<void>;
 
 export type OnChange = ({|
-    valid : boolean
+    valid : boolean,
+    errors : $ReadOnlyArray<string>
```
bluepnume, line 27: Let's define an enum for errors. Would be good to know some examples of the kinds of errors we are pushing up to the parent.

src/card/components/CardNumber.jsx
```diff
@@ -0,0 +1,133 @@
...
+    const [ isNumberValid, setIsNumberValid ] = useState({ isValid: false, isPossibleValid: true });
+
+
+    useEffect(() => {
+        setIsNumberValid(checkCardNumber(number, cardType));
+        if (typeof onValidityChange === 'function') {
+            onValidityChange(isNumberValid);
+            if (number && isNumberValid.isValid) {
+                // $FlowFixMe
+                if (!isSwitched && checkRef(nextRef)) {
```
bluepnume, line 51: Having the card number know about the "next" component seems like very tight coupling. Can we orchestrate this in the parent and send a `onCardInputComplete()` callback up or something?

src/card/components/CardNumber.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+        setCardType(detectCardType(value));
+
+        if (maskedValue.length !== maskedNumber.length && maskedValue.length === selectionStart + 1) {
+            startCursorPosition += 1;
+            endCursorPosition += 1;
+        }
+
+        moveCursor(event, startCursorPosition, endCursorPosition);
+        setCursorStart(startCursorPosition);
```
bluepnume, line 87: There is a lot of state changing here. Can we abstract this into a custom hook, perhaps? I'm assuming the objective here is formatting but keeping the current cursor position -- so maybe we could have a generic hook that works for different input fields like:

```javascript
const { ref, value, setValue } = useInputWithFormatting({
    formatter: (val) => formatValue(val)
})

return <input ref={ ref } value={ value } onChange={ e => setValue(e.target.value) } />
```

src/card/types.js
```diff
@@ -10,3 +10,94 @@ export type Card = {|
     cvv : string,
     expiry : string
 |};
+
+export type FieldStyle = {|
```
bluepnume, line 14: Do we have validation to make sure a merchant can pass only these values, and no more?

src/card/types.js
```diff
@@ -10,3 +10,94 @@ export type Card = {|
...
+    patterns : $ReadOnlyArray<number>,
+    type : string,
+    niceType : string,
+    code : {|
+        name : string,
+        size : number
+     |}
+|};
+
+export type CardNumberEvent = {|
```
bluepnume, line 82: Can we make these names more specific? I'm assuming this is a change event, so maybe `CardNumberChangeEvent`?

src/api/order.js
```diff
@@ -862,15 +864,45 @@ type ApproveCardPaymentOptions = {|
...
+                $clientId: String!
+                $card: CardInput!
+                $branded: Boolean!
+            ) {
+                processPayment(
+                    clientID: $clientId
+                    paymentMethod: { type: CREDIT_CARD, card: $card }
+                    branded: $branded
+                    token: $token
+                    buttonSessionID: "f7r7367r4"
```
bluepnume, line 892: @shrutikapoor08 -- There won't always be a button session id, since merchants may choose to use their own button. Can we make this optional?

We also have a `cardFieldsSessionID` -- so maybe we could do a tracking object like `tracking { buttonSessionID,  cardFieldsSessionID }`?

src/api/order.js
```diff
@@ -862,15 +864,45 @@ type ApproveCardPaymentOptions = {|
...
+        `,
+        variables: { token: orderID, clientId: 'alc_client1', card, branded }
+    }).then((gqlResult) => {
+        if (!gqlResult) {
+            throw new Error(`GraphQL GetApplePayPayment returned no applePayment object`);
+        }
+        return gqlResult;
     });
+
+    // return ZalgoPromise.try(() => {
```
bluepnume, line 904: Feel free to just remove this

src/card/components/CardCVV.jsx
```diff
@@ -0,0 +1,101 @@
...
+    navigation : CardNavigation,
+    onChange : (cvvEvent : CardCvvChangeEvent) => void,
+    onFocus : (event : InputEvent) => void,
+    onBlur : (event : InputEvent) => void,
+    allowNavigation : boolean,
+    onValidityChange? : (numberValidity : FieldValidity) => void
+|};
+
+
+export function CardCVV({ name = 'cvv', navigation = defaultNavigation, state, ref, type, className, placeholder, style, maxLength, onChange, onFocus, onBlur, onValidityChange, cardType, allowNavigation = false } : CardCvvProps) : mixed {
```
bluepnume, line 29: Nitpick but let's break long function signatures like this onto multiple lines.

src/card/components/fields.jsx
```diff
@@ -2,91 +2,165 @@
...
+        const errors = setErrors({ isNumberValid: numberValidity.isValid, isCvvValid: cvvValidity.isValid, isExpiryValid: expiryValidity.isValid });
+
+        onChange({ value: { number, cvv, expiry }, valid, errors });
+
+    }, [
+        number,
+        cvv,
+        expiry,
+        isValid,
+        JSON.stringify(numberValidity),
```
bluepnume, line 82: Do we need to stringify all of these? They should be fresh objects each time right, so React should pick that up?

src/card/constants.js
```diff
@@ -1,17 +1,144 @@
...
+
 export const CARD_FIELD_TYPE = {
     SINGLE: 'single',
     NUMBER: 'number',
     CVV:    'cvv',
     EXPIRY: 'expiry'
 };
 
+export const CARD_ERRORS = {
+    INVALID_NUMBER:  ('Invalid card number' : 'Invalid card number'),
```
bluepnume, line 17: Let's make the values just like `INVALID_NUMBER` rather than human-readable strings in english. Then the merchant can map these to errors to show the user.

src/card/constants.js
```diff
@@ -1,17 +1,144 @@
...
+    marginLeft:              'margin-left',
+    padding:                 'padding',
+    paddingTop:              'padding-top',
+    paddingRight:            'padding-right',
+    paddingBottom:           'padding-bottom',
+    paddingLeft:             'padding-left',
+    textAlign:               'text-align',
+    textShadow:              'text-shadow',
+    transition:              'transition',
+    MozAppearance:           '-moz-appearance',
```
bluepnume, line 74: Should we bother with vendor scoped CSS flags? I'm assuming things like box-shadow etc. work natively in all major browsers now. Maybe let's remove these and add them if requested?

src/card/components/fields.jsx
```diff
@@ -2,91 +2,165 @@
...
+   placeholder : {| number? : string, expiry? : string, cvv? : string  |}
 |};
 
-export function CardNumberField({ cspNonce, onChange } : CardNumberFieldProps) : mixed {
+export function CardNumberField({ cspNonce, onChange, styleObject = {}, placeholder = {} } : CardNumberFieldProps) : mixed {
     const [ number, setNumber ] = useState('');
+    const [ numberValidity, setNumberValidity ] = useState(true);
+    const [ generalStyles, inputStyles ] = getStyles(styleObject);
+
+    const compousedStyles = { ...{ input: DEFAULT_INPUT_STYLE },  ...generalStyles };
```
bluepnume, line 160: `composedStyles`

src/card/components/CardCVV.jsx
leogedler: Sure I'll do it.

src/card/components/fields.jsx
leogedler: You are right `onValidityChange` send a new object and I just tested it and is working as expected without the stringify.

src/card/constants.js
leogedler: Got it

src/card/constants.js
leogedler: I agree, I'll remove these.

src/card/types.js
leogedler: I have added a validation for the styles, now we will filter out unsupported styles.

nbierdeman:
@leogedler Nice work! I really like it so far. I just noticed when I tab between fields and hit "delete," I get "shift + tab" behavior, where the cursor moves to the previous field. Is that expected?
Screen Shot 2021-10-01 at 4 22 30 PM
Screen Shot 2021-10-01 at 4 22 47 PM

nbierdeman (commented):
@leogedler Nice work! I really like it so far. I just noticed when I tab between fields and hit "delete," I get "shift + tab" behavior, where the cursor moves to the previous field. Is that expected?
Screen Shot 2021-10-01 at 4 22 30 PM
Screen Shot 2021-10-01 at 4 22 47 PM

src/card/types.js
```diff
@@ -10,3 +10,130 @@ export type Card = {|
...
+    marginLeft? : string,
+    padding? : string,
+    paddingTop? : string,
+    paddingRight? : string,
+    paddingBottom? : string,
+    paddingLeft? : string,
+    textAlign? : string,
+    textShadow? : string,
+    transition? : string,
+    MozAppearance? : string,
```
bluepnume, line 59: We can remove here too right?

bluepnume:
This PR is looking awesome, and we should merge soon. Can you add a few test cases though? I don't think those cases need to be 100% comprehensive until we're pushing to go live, but it would at least be good to have a few example tests for the happy case of `createOrder -> enter fields -> submit -> onApprove`.

src/card/types.js
leogedler: Sure, I'll remove it.

leogedler:
@nbierdeman good catch!, hmm not that is not intended, I think we can improve that to make it more natural and remove the selected text.

leogedler:
@bluepnume got it!, I'll add some tests.

src/card/components/CardNumber.jsx
```diff
@@ -0,0 +1,191 @@
...
+    const onFocusEvent : mixed = (event : InputEvent) : mixed => {
+        if (typeof onFocus === 'function') {
+            onFocus(event);
+        }
+
+        const maskedValue = maskCard(inputValue);
+        const newState = { ...inputState, maskedInputValue: maskedValue };
+        
+        if (isValid) {
+            // Timeout needed to wait for the 4 digit mask replacement when the inout is valid
```
nbierdeman, line 137: Small nitpick:
```suggestion
            // Timeout needed to wait for the 4 digit mask replacement when the input is valid
```

nbierdeman:
Mobile testing update - it appears that the iOS issue above is related to BrowserStack (it works using an iOS mobile user agent in desktop Chrome and also using Xcode/Simulator).

bluepnume:
Thanks for adding the tests @leogedler! Will let @nbierdeman sign off too, but I'm good to get this merged in so we can start adding more incremental stuff like multi-field.

nbierdeman:
@leogedler, I just pulled the latest, and now when I go to update the expiration month, it deletes the entire date up to the character I just deleted. Is there a way to allow deleting a single character within the date without affecting the rest of it? Or maybe this is something we can work on after merging? I do think it's an improvement over having the cursor jump to the end of the date.  

Expiration-Date-Delete-Issue (1)

PR #244, opened 2021-09-10
Adding dimensions as a function and accept different values

aditya27marri:
By adding dimensions function to checkout we can try and pass different values into creating checkout window.


bluepnume:
Please back out /dist changes

src/constants.js
```diff
@@ -262,3 +263,22 @@ export const BUTTON_LABEL = {
 export const STATUS_CODES = {
     TOO_MANY_REQUESTS: 429
 };
+
+export const APM_LIST = {
```
bluepnume, line 267: Can this just be an array?

src/payment-flows/checkout.js
```diff
@@ -241,6 +241,16 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
             },
 
             onError,
+            dimensions: () => {
+                if (Object.values(APM_LIST).includes(fundingSource)) {
```
bluepnume, line 245: We can't use Object.values in IE11

src/payment-flows/checkout.js
```diff
@@ -241,6 +241,16 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
...
+                    getLogger().info(`dimensions_apmlist_fundingsource_${ fundingSource }`).flush();
```
bluepnume, line 246: Can we keep this event generic? Maybe just `popup_dimensions_${ fundingSource} { width, height }`

src/payment-flows/checkout.js
```diff
@@ -288,7 +298,18 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
         return ZalgoPromise.try(() => {
             if (!win && supportsPopups()) {
                 try {
-                    win = openPopup({ width: CHECKOUT_POPUP_DIMENSIONS.WIDTH, height: CHECKOUT_POPUP_DIMENSIONS.HEIGHT });
+                    const { width, height } = () => {
+                        if (Object.values(APM_LIST).includes(fundingSource)) {
```
bluepnume, line 302: Let's have a reusable function to calculate the dimensions rather than duplicating this logic.

src/payment-flows/checkout.js
```diff
@@ -241,6 +241,16 @@ function initCheckout({ props, components, serviceData, payment, config } : Init
             },
 
             onError,
+            dimensions: () => {
```
bluepnume, line 244: This should be an object type rather than a function, I think.

src/types.js
```diff
@@ -62,7 +62,11 @@ export type CheckoutProps = {|
     enableFunding : ?$ReadOnlyArray<FundingType>,
     standaloneFundingSource : ?FundingType,
     amplitude? : boolean,
-    branded : boolean | null
+    branded : boolean | null,
+    dimensions : () => {|
```
bluepnume, line 66: Let's use an object type

src/payment-flows/checkout.js
```diff
@@ -101,6 +101,16 @@ function getContext({ win, isClick } : {| win : ?(CrossDomainWindowType | ProxyW
     return CONTEXT.IFRAME;
 }
 
+function getDimensions(fundingSource : string) : Object {
+    if (APM_LIST.indexOf(fundingSource) !== -1) {
+        getLogger().info(`popup_dimensions_value_${ fundingSource }`).flush();
+        return { width: 1282, height: 720 };
```
westeezy, line 107: How were these values chosen? Curious the logic here and if it should potentially just use the defaults as well.

westeezy:
I went through and marked the previous conversations as resolved as I noticed the code changes were addressed as per @bluepnume's comments. My apologies if that is untrue but it makes it easier to review. Feel free to reopen those of course.

src/payment-flows/checkout.js
```diff
@@ -101,6 +101,16 @@ function getContext({ win, isClick } : {| win : ?(CrossDomainWindowType | ProxyW
     return CONTEXT.IFRAME;
 }
 
+function getDimensions(fundingSource : string) : Object {
```
gregjopa, line 104: Please avoid using the generic `Object` type.
```suggestion
function getDimensions(fundingSource : string) : {| width : number, height : number |} {
```

src/payment-flows/checkout.js
aditya27marri: that was our product requirement. I can give a link to the slack conversation regarding this resize.

src/payment-flows/checkout.js
aditya27marri: fixed it

aditya27marri:
retest this please

src/payment-flows/checkout.js
westeezy: Yes please share! I want to ensure we are providing the best experience possible via our packages so additional information is important to our review. I would like any custom dimensions to be passed in from the consumer so is that an option in your case? By that I mean `getDimmensions` will return the default unless you pass in explicit alternative dimmensions. To me that is more clear of an experience for other folks integrating to this package.

aditya27marri:
> I went through and marked the previous conversations as resolved as I noticed the code changes were addressed as per @bluepnume's comments. My apologies if that is untrue but it makes it easier to review. Feel free to reopen those of course.
here is the slack conversation regarding the resize . https://paypal.slack.com/archives/C01BKDP59C1/p1627410260007600 you can see the size values we wanted to change to inbetween.

src/payment-flows/checkout.js
aditya27marri: here is the conversation that we talked about resize https://paypal.slack.com/archives/C01BKDP59C1/p1627410260007600 
We came to the size conclusion after our product requirement it self. please check it.

aditya27marri:
Retest this please

gregjopa (approved):
Looks good to me!

src/payment-flows/checkout.js
bluepnume: Can you define a `CHECKOUT_APM_POPUP_DIMENSIONS` so this can be re-used where needed?

bluepnume:
Looks good to me! Please add a test to validate the dimensions are correct in both the apm and non-apm case.

PR #247, opened 2021-09-15
Venmo Desktop - Consumer Feedback collection

andregardi:
### Description
Feedback on Desktop XO would need to be collected at only one point on the QR code scanning page when the consumer hits the X option.  We do not want to collect the feedback post the QR scan because we are already collecting the feedback on the mobile experience.

#### Linked PR:
https://github.com/paypal/paypal-checkout-components/pull/1747

#### Changes
- Implement the new survey on QR code modal, asking for reason to not complete payment.
- Send a tracking event to FPTI to log the feedback reason collected on the survey.

#### Screenshot
image


src/constants.js
```diff
@@ -178,17 +179,18 @@ export const FPTI_BUTTON_TYPE = {
...
+    HONEY_SESSION_ID:           ('honey_session_id' : 'honey_session_id'),
+    INTEGRATION_ISSUE:          ('integration_issue' : 'integration_issue'),
+    INTEGRATION_WHITELIST:      ('whitelist' : 'whitelist'),
+    INFO_MSG:                   ('info_msg' : 'info_msg'),
+    PMT_TOKEN:                  ('pmt_token' : 'pmt_token'),
+    TRANSITION_TYPE:            ('transition_type' : 'transition_type'),
+    TRANSITION_REASON:          ('transition_reason' : 'transition_reason'),
+    SHIPPING_CALLBACK_PASSED:   ('shipping_callback_passed' : 'shipping_callback_passed'),
+    SHIPPING_CALLBACK_INVOKED:  ('shipping_callback_invoked' : 'shipping_callback_invoked'),
+    DESKTOP_EXIT_SURVEY_REASON: ('reason' : 'reason')
```
remotevision, line 193: @andregardi can we make `reason` more descriptive? maybe just keep it at `desktop_exit_survey_reason`?

andregardi: Sure we can @remotevision !
This `reason` name was suggested by the the instrumentation team, but I believe that using `desktop_exit_survey_reason` is a better solution. Thank you!

src/qrcode/qrcard.jsx
```diff
@@ -81,12 +103,22 @@ function QRCard({
             <Logo />
         </div>
     );
+    
+    const surveyElement = (
+        <Survey survey={ survey } onCloseClick={ onCloseClick } />
+    );
```
mrbaguvix, line 109: Don't think we need to reassign the entire component? Can use directly.

src/types.js
```diff
@@ -128,6 +128,7 @@ export type QRCodeProps = {|
     state? : $Values<typeof QRCODE_STATE>,
     errorText? : string,
     onClose? : () => ZalgoPromise<void>,
+    onSubmitFeedback? : (reason : string) => void,
```
mrbaguvix, line 131: Should we declare a map type?
```suggestion
    onSubmitFeedback? : (reason : Reasons[]) => void,
```

test/client/native-qrcode.js
```diff
@@ -728,5 +728,107 @@ describe('native qrcode cases', () => {
                 gqlMock.done();
             });
         });
+
+        it('should render a button with createOrder, click the button, and render checkout via qrcode that contains escape path with empty ineligibilityReason and onSubmitFeedback', async () => {
+            return await wrapPromise(async ({ expect, avoid }) => {
+                window.xprops.platform = PLATFORM.DESKTOP;
```
mrbaguvix, line 734: Would be better to break this unit tests into at least 2 assertions with _describe/it_ block

mrbaguvix (changes requested):

src/types.js
andregardi: Thank you for your suggestion David.
The parameter on the `onSubmitFeedback` is actually one single reason, on singular. That means we intent this to be a string.

src/qrcode/qrcard.jsx
andregardi: Yes, using it directly would also work. But keep it out of the returning JSX help us to keep a clean code on conditional rendering elements and allow us to move its position more easily in future changes.

src/payment-flows/native/qrcode.js
```diff
@@ -154,6 +155,15 @@ export function initNativeQRCode({ props, serviceData, config, components, fundi
                 });
             };
 
+            const onSubmitFeedback = (reason : string) => {
+                getLogger().info(`VenmoDesktopPay_qrcode_survey`).track({
```
bluepnume, line 159: Can we just do this logging inside the survey component? Better to avoid having the parent care about this survey at all right?

andregardi: I like that idea!
I was just following the pattern established on other events. But will be happy to move that to the component.

src/qrcode/components.jsx
```diff
@@ -74,6 +74,79 @@ export function AuthMark() : mixed {
     );
 }
 
+const radioSvg = (
```
mnicpt, line 77: Move all survey stuff into its own file.

andregardi: Yes @mnicpt, I will be happy to do that!

PR #248, opened 2021-09-15
Remove optional argument for `npm run release` from `workflow_dispatch`

nbierdeman:
### Purpose

This PR removes the optional argument for `npm run release` from the `workflow_dispatch` event, since we always do patch releases:

Screen Shot 2021-09-14 at 4 12 12 PM


westeezy (approved):
:fire:

PR #252, opened 2021-09-17
Convert skip native duration from a string to integer

leogedler:
### Description

On the iOS fallback the params are always string and for the iOS opt-out it is required to have the `skip_native_duration` as integer.

#### What is being changed from a technical perspective?

Added logic on the `setNativeOptOut` to parse to int the `skip_native_duration` on the fallback event.

#### Why are we making these changes? Include and reference tickets (Jira, Github Issue, etc)

- https://engineering.paypalcorp.com/jira/browse/DTPPSDK-345

When using the opt-out fallback event and passing `skip_native_duration` (in milliseconds) the JS SDK should handle this properly and set the `nativeOptOutLifetime` calculated as `Date.now()` + `skip_native_duration` (all in milliseconds, unix timestamp).

If the params `skip_native_duration` is not passed on the fallback event the default value should be: `let OPT_OUT_TIME = 6 * 7 * 24 * 60 * 60 * 1000;`

### DIFF
--- a/src/payment-flows/native/eligibility.js
+++ b/src/payment-flows/native/eligibility.js
@@ -228,8 +228,11 @@ export function setNativeOptOut(optOut : NativeOptOutOptions) : boolean {
 
         // Opt-out 6 weeks from native experience as default
         let OPT_OUT_TIME = 6 * 7 * 24 * 60 * 60 * 1000;
-        if (skip_native_duration && typeof parseInt(skip_native_duration, 10) === 'number') {
-            OPT_OUT_TIME = parseInt(skip_native_duration, 10);
+
+        const parsedSkipDuration = parseInt(skip_native_duration, 10);
+
+        if (parsedSkipDuration && typeof parsedSkipDuration === 'number') {
+            OPT_OUT_TIME = parsedSkipDuration;
         }
         
         const now = Date.now();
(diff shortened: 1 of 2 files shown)


src/payment-flows/native/eligibility.js
```diff
@@ -228,8 +228,8 @@ export function setNativeOptOut(optOut : NativeOptOutOptions) : boolean {
 
         // Opt-out 6 weeks from native experience as default
         let OPT_OUT_TIME = 6 * 7 * 24 * 60 * 60 * 1000;
-        if (skip_native_duration && typeof skip_native_duration === 'number') {
-            OPT_OUT_TIME = skip_native_duration;
+        if (skip_native_duration && typeof parseInt(skip_native_duration, 10) === 'number') {
```
westeezy, line 231: minor nitpick but can we set the `parseInt` result to a variable and check against that variable instead? Then we don't have to call parseInt twice and should be a little more readable

leogedler: @westeezy thanks for the feedback. I have pushed a change for this.

westeezy (approved):

mnicpt (approved):
Thanks, @leogedler, for tracking this down!

PR #260, opened 2021-09-24
[DO NOT MERGE] Turn off traffic for iOS15 in Native Flow

mnicpt:
### Description
Currently, when app switching on iOS 15, we don't return to original page.  This seems to be an issue with iOS.

❤️  Thank you!


src/payment-flows/native/eligibility.js
```diff
@@ -100,8 +100,12 @@ export function prefetchNativeEligibility({ props, serviceData } : PrefetchNativ
     });
 }
 
+function isIOS15() : boolean {
+    return (/ iPhone OS 15_0/).test(window.navigator.userAgent);
+}
+
 export function canUsePopupAppSwitch({ fundingSource } : {| fundingSource : ?$Values<typeof FUNDING> |}) : boolean {
-    if (!isIOSSafari() && !isAndroidChrome()) {
+    if (!isIOSSafari() && !isAndroidChrome() && !isIOS15()) {
```
bluepnume, line 108: This condition won't pass since `isIOSSafari` is true. Can we make this a different conditional block?

bluepnume (approved):

src/payment-flows/native/eligibility.js
```diff
@@ -100,7 +100,15 @@ export function prefetchNativeEligibility({ props, serviceData } : PrefetchNativ
     });
 }
 
+function isIOS15() : boolean {
+    return (/ iPhone OS 15_/).test(window.navigator.userAgent);
+}
+
 export function canUsePopupAppSwitch({ fundingSource } : {| fundingSource : ?$Values<typeof FUNDING> |}) : boolean {
+    if (isIOS15()) {
```
bluepnume, line 108: BTW we should still allow when `enableNativeCheckout` flag is passed so we can keep testing

TNuzzi (approved):

PR #261, opened 2021-09-27
set error message based on issue code

sridharrajs:
### Description

Smart payment buttons throws a generic `Validate payment failed with status: ${ status }` error when `validate-payment-method` call  fails as per [this](https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/payment-flows/vault-capture.js#L100). Instead we should throw the error message with`issue` code, if available.

Merchants needs to know the exact `issue` code to display the relevant actionable message to the user. 

### Screenshots

**Existing**
Screenshot 2021-09-27 at 1 01 40 PM

**Proposed**
Screenshot 2021-09-27 at 12 59 44 PM

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -346,10 +346,10 @@ export type ValidatePaymentMethodResponse = {|
     links? : $ReadOnlyArray<{|
         rel : string
     |}>,
-    details? : $ReadOnlyArray<{
+    details? : $ReadOnlyArray<{|
         issue? : string,
         description? : string
-    }>
+    |}>
 |};
 
 type PaymentSource = {|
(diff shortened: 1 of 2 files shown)


bluepnume:
One thing I'm not clear on -- are we expecting merchants to parse the `err.message` string here to decide what to display to the user?

bluepnume:
Yeah this doesn't seem ideal -- the merchant should not have to parse a string to determine how to handle an error. Can we pass back an `err.code` value instead with an error code?

sridharrajs:
I've updated the check to use the `issue` code, if available. This is how it will look in the UI. @bluepnume 

Screenshot 2021-12-16 at 1 27 26 PM

bluepnume (approved):

PR #262, opened 2021-10-02
fix: add additional logic to onclose in checkout flow for non-instant

jscheinhorn:
### Description
Add `!approved` check to onClose logic in checkout flow in order to prevent merchant's onApprove callback from being fired twice for non-instant payment methods when the buyer proceeds past the PDF page and goes through Latinumnodeweb post-processing.

onApprove callback in merchant integration is being called twice when buyer goes to postProcessing flow using non-instant APM (oxxo).
Fires once when Latinumnodeweb calls xprops.onAuthorize.
Fires again when paypal-smart-payment-buttons calls onClose and checks for `doApproveOnClose` global set to true (non-instant apms).

### Reproduction Steps
1. Select non-instant payment method using unbranded (drop-in) UI
2. Click payment button (Latinum pre-processing runs)
3. Select PENDING status on simulator page and click submit
4. Allow Latinumnodeweb post-processing to run (pop-up will close)
5. onApprove callback fires twice

### Screenshots
image

This PR relates to Latinumnodeweb pre- and post-processing flows for non-instant payments

❤️  Thank you!


bluepnume:
Lgtm! Can we get a test to cover the broken case? Thanks!

PR #263, opened 2021-10-04
Fix iOS 15

mnicpt:
### Description
In some cases in iOS 15, the user is returned to the wrong tab.  This PR attempts to fix that by controlling navigation.


src/native/popup/popup.js
```diff
@@ -245,11 +245,15 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
         }
         case HASH.ON_APPROVE: {
             const { payerID, paymentID, billingToken } = parseQuery(queryString);
-            sendToParent(MESSAGE.ON_APPROVE, { payerID, paymentID, billingToken }).finally(closeWindow);
+            sendToParent(MESSAGE.ON_APPROVE, { payerID, paymentID, billingToken })
+                .then(window.history.back)
```
bluepnume, line 249: By the time we get to this .then the popup will have been closed by the parent, right?

bluepnume: Wait nevermind I needed to keep reading the PR. :)

PR #272, opened 2021-10-14
Divide logo animation POC

erick2014:
### Description
This pr shows is how will look like moving the animation one from paypa-checkout-components repo to this repository and accomplish the same result.


mnicpt (changes requested):
I'm not so sure moving this logic here is the best choice as SPB's role with buttons is to get it ready for payment and not to component construction and styling.

PR #274, opened 2021-10-18
Fix Fallback for iOS

mnicpt:
## Description
Fallback path doesn't work in iOS for opt-out scenario.

### DIFF
--- a/src/payment-flows/native/eligibility.js
+++ b/src/payment-flows/native/eligibility.js
@@ -198,9 +198,7 @@ export function isNativePaymentEligible({ props, payment } : IsPaymentEligibleOp
 export type NativeFallbackOptions = {|
     type? : string,
     skip_native_duration? : number,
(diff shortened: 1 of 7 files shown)


src/payment-flows/native/eligibility.js
```diff
@@ -185,21 +185,11 @@ export function isNativePaymentEligible({ props, payment } : IsPaymentEligibleOp
...
-        if (
-            ineligibleReasons &&
-            !eligible
-        ) {
-            return false;
-        }
+        return nativeEligibilityResults && nativeEligibilityResults[fundingSource]
+            ? nativeEligibilityResults[fundingSource].eligibility
+            : false;
     }
```
vladfilyakov, line 193: I guess this logic to ignore 'isUserAgentEligible' and 'isBrowserMobileAndroid' in ineligibility reasons moved somewhere else?

PR #275, opened 2021-10-18
Single Card Fields auto focus fix for Safari & Firefox

crookedneighbor:
### Description

Worked with @leogedler on this.

Basically, the issue is that on Safari and Firefox, when you tab from outside of the iframe to inside the iframe, it first focusses on the body or html element in the iframe, and not the input, requiring an additional tab to focus the input. This interrupts the expected user experience.

It looks to me like we don't have test files for `card-utils.js` or the components, so I'm not sure how exactly we're supposed to be testing this behavior using the conventions here.

If we did have test mechanisms for both, I'd simply test the actual logic in the card-utils test file, and then test that the function was called with the correct ref for the components.

Looking for guidance about how to provide coverage for these areas.

As part of the audit to have parity with Braintree's Hosted Fields, we identified this need.

We will also need to make these changes to the multi card fields components, but they require additional work from @leogedler to get it working.

Jira: DTNOR-236

In Safari or Firefox, attempt to tab into the iframe's inputs.


westeezy (approved):

crookedneighbor:
One last note, I do think we should have component level tests, but since that will require some choices for _how_ we implement that, I'm fine waiting to do those in a separate dedicated PR.

PR #277, opened 2021-10-19
refactor: simplify Props definition

crookedneighbor:
### Description

Noticed that there is a lot of overlap between the definition for Props and the definition for XProps and figured we could probably reduce some of the redundancy by having a BaseProps type that is not exported.

When making a change, I found I had to add the same prop to multiple places. For Props and XProps, the definitions were so similar, that it seemed good to inherit them.

### Additional Questions

I noticed that `enableNativeCheckout` has `boolean | void` in `XProps` and just `boolean` in `Props`. Should it be the same, or is there a special reason for it to be different?

`branded?` is optional in `XProps` and `branded` in `Props`. Should it be different?

`createSubscription` is marked as the type `XCreateSubscription`, when it should probably be `CreateSubscription` (which itself just equals `XCreateSubscription` in the create-subscription type definition file) https://github.com/paypal/paypal-smart-payment-buttons/blob/0638dfa16c39489c0ab398be00585776d465cae9/src/props/createSubscription.js#L38


crookedneighbor:
Worked on this with @westeezy

crookedneighbor:
I'm just going to close this for now then. If any flow experts want to work on this to match the specification laid out by @bluepnume, go for it.

PR #284, opened 2021-10-22
Animation experiment setup

ssono:
### Description

* Adding layout and buttonSize to the call to personalization
* Needed to determine which experiments can be run on a flow

❤️  Thank you!


server/service/personalization.jsx
```diff
@@ -149,14 +163,38 @@ export async function resolvePersonalization(req : ExpressRequest, gqlBatch : Gr
...
+        period,
+        taglineEnabled,
+        renderedButtons,
+        layout,
+        buttonSize
+    };
+
+    // Fix enum checking errors for strings on graphql by only sending truthy variables
+    for (const key of Object.keys(personalizationVariables)) {
+        if (personalizationVariables[key] === '') {
```
mnicpt, line 188: @ssono Why not just `if (!personalizationVariables[key]) {` to have a more broader truthy check?

mnicpt: Is `taglineEnabled` boolean or string?

mnicpt (approved):
Left a comment but other than that looks good!

server/service/personalization.jsx
ssono: It can be either @mnicpt . It should only check strings otherwise false values will be sent as undefined which can mess stuff up further down the line

gregjopa (approved):
Looks good!

PR #289, opened 2021-11-01
chore: use indexOf over includes for IE11 support

crookedneighbor:
Still some question about whether or not we need to support IE11 for this project, but if we do, here's a PR that swaps out `includes` for the more compatible `indexOf`.


westeezy (approved):
we really need to correct the transpiling to take care of this, but in terms of the fix lgtm. Thanks for updating this @crookedneighbor !

nbierdeman (approved):
🏆

PR #291, opened 2021-11-03
Export function to check guestEnabled

amyegan:
### Description
Adding `isGuestEnabled` function to `exports` so the status can be checked programatically

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1036

We previously had no way to check from the client if guest was enabled. Now we can:
1. Open a sample integration page in browser
2. In browser console, run `window.exports.isGuestEnabled()` to see whether the test merchant has guest enabled or blocked

n/a

All other changes related to the `guestEnabled` flag are already in production


TNuzzi:
@amyegan Curious not sure how honey works but how will you get the merchantId?  We are seeing a lot of of V5 merchant integration with just the clientId.  I implemented an update to GetFundingEligibility to lookup the merchant id so we could filter out vs MCC code (for venmo).

amyegan:
@TNuzzi Good question. Honey should be passing us the merchantId when they load the SDK. I'll take a look your update to GetFundingEligibility and try to make this change a bit smaller. Thanks for letting me know!

TNuzzi:
@amyegan The honey integration is still new to me but could honey use the fundingEligibility results in JS SDK?  Since those are cached it would be immediately available when the SDK is loaded.  I think we would update the query in clientsdknodeweb to include your flag in the payload.

amyegan:
@TNuzzi That's how we originally set it up. [CSNW already has the flag](https://github.paypal.com/Checkout-R/clientsdknodeweb/blob/211a771e6c83abf994614ee70317e198c94ef031/server/lib/eligibility.js#L149) and I am seeing it come through on xprops in sandbox, but I'm only getting back basic funding eligibility in stage so we could never see the status to test or confirm it was really working. I must be missing something about how fundingEligibility works in stage

src/button/exports.js
```diff
@@ -30,6 +33,9 @@ export function setupExports({ props, isEnabled, facilitatorAccessToken } : Expo
             currency,
             intent
         },
+        currency,
+        intent,
+        isGuestEnabled: () => fundingEligibility?.card || getGuestEnabledStatus(merchantID),
```
wsbrunson, line 38: should this be `fundingEligibility?.card.hasOwnProperty('guestEnabled') ? fundingEligibility.card.guestEnabled : getGuestEnabledStatus(merchantID)`? 

I don't think we have a test for when `setupButton` actually receives a funding eligibility object, it looks like its using the default one:
```
export const DEFAULT_FUNDING_ELIGIBILITY : FundingEligibilityType = {
    [ FUNDING.PAYPAL ]: {
        eligible: true
    }
};
```

wsbrunson (approved):

amyegan:
Thanks @wsbrunson for the thoughtful review and for catching my mistakes!

PR #292, opened 2021-11-08
Auto Retry for Native Flow

mnicpt:
### Description
In some cases, when the user has the app, PayPal or Venmo, installed; the user doesn't app switch.  We have found that by trying again, the app switch goes through.  For cases where the app has been detected on the Android device, and is native eligible after getting order id, we will attempt to retry the app switch once.

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1028


wsbrunson (approved):
Do we have any way to write tests for this?

PR #293, opened 2021-11-09
fix(native payment flow): not eligible if desktop and non-venmo

TNuzzi:
### Description

Found an issue where native payment flow was being allowed for non-venmo funding sources on desktop.

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1034

❤️  Thank you!

### DIFF
--- a/src/payment-flows/native/eligibility.js
+++ b/src/payment-flows/native/eligibility.js
@@ -135,11 +135,17 @@ export function canUseNativeQRCode({ fundingSource, win } : {| fundingSource : ?
 }
 
 export function isNativeEligible({ props, config, serviceData } : IsEligibleOptions) : boolean {
-    const { clientID, fundingSource, onShippingChange, createBillingAgreement, createSubscription, env } = props;
+    const { clientID, fundingSource, onShippingChange, createBillingAgreement, createSubscription, env, platform } = props;
     const { firebase: firebaseConfig } = config;
-    const { merchantID } = serviceData;
+    const { merchantID, fundingEligibility } = serviceData;
+    const isVenmoEligible = fundingEligibility?.venmo?.eligible;
 
     if (!firebaseConfig) {
+        return false;
+    }
+
+    // If Desktop and venmo not eligible, native payment flow is not eligible
+    if (platform && platform === PLATFORM.DESKTOP && !isVenmoEligible) {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


mnicpt (approved):
Awesome, @TNuzzi !!!

test/client/native-qrcode.js
```diff
@@ -503,5 +479,56 @@ describe('native qrcode cases', () => {
...
+                        throw new Error(`Expected payerID to be ${ payerID }, got ${ data.payerID }`);
+                    }
+                }));
+        
+                createButtonHTML({ fundingEligibility: DEFAULT_FUNDING_ELIGIBILITY });
+    
+                // Use default fundingEligibility which is PayPal only
+                await mockSetupButton({
+                    fundingEligibility: DEFAULT_FUNDING_ELIGIBILITY,
+                    // eslint-disable-next-line key-spacing
```
TNuzzi, line 520: I have not clue why linting did not like this.  I had to add this stupid rule to get it working.

gregjopa: I think it wants the values of `fundingEligibility` and `eligibility` to line up which is a bit awkward when the value is an object. How about this?

```js
const eligibility: {
    cardFields: false,
    native:     true
};

await mockSetupButton({
    fundingEligibility: DEFAULT_FUNDING_ELIGIBILITY,
    eligibility
});
```

gregjopa (approved):

test/client/native-qrcode.js
TNuzzi: Ugh, I love and hate linting.  ... Just tell me what you want!!!!

src/payment-flows/native/eligibility.js
```diff
@@ -135,14 +135,20 @@ export function canUseNativeQRCode({ fundingSource, win } : {| fundingSource : ?
...
-    const { merchantID } = serviceData;
+    const { merchantID, fundingEligibility } = serviceData;
+    const isVenmoEligible = fundingEligibility?.venmo?.eligible;
 
     if (!firebaseConfig) {
         return false;
     }
 
+    // If Desktop and venmo not eligible, native payment flow is not eligible
+    if (platform && platform === PLATFORM.DESKTOP && !isVenmoEligible) {
```
westeezy, line 148: flow types do have platform as required so might be safe to just have the equality. no deal breaker on merge just figured id mention it

westeezy (approved):
lgtm

test/client/native-qrcode.js
westeezy: you can do `npm run lint -- --fix` in the future and the autofixer will catch this one. i also really dislike this rule so i code with it off then fix on precommits

TNuzzi:
@westeezy That is a good callout on `platform`.  This is used in a couple of places so would be good as a tech debt task.

PR #296, opened 2021-11-09
Fix native app detection

mnicpt:
### Description
Before we returned null if we didn't get anything back for `getInstalledRelatedApps`.  Decision from Android team is to return false to ensure version below are not included.

### DIFF
--- a/src/native/popup/popup.js
+++ b/src/native/popup/popup.js
@@ -265,6 +265,10 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
             if (appSwitch) {
                 const timer = setTimeout(() => {
                     if (!didRedirect) {
+                        logger.info('Redirecting to native...');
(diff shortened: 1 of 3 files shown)


src/native/popup/popup.js
```diff
@@ -265,6 +265,10 @@ export function setupNativePopup({ parentDomain, env, sessionID, buttonSessionID
             if (appSwitch) {
                 const timer = setTimeout(() => {
                     if (!didRedirect) {
+                        logger.info('Redirecting to native...');
+                        redirectUrl = redirectUrl.replace('https://localhost.paypal.com:8443', 'https://www.paypal.com');
```
TNuzzi, line 269: @mnicpt Is this suppose to be here?

mnicpt: Whoa!  Must have been uncommitted from other task.  Thanks.

TNuzzi (approved):

TNuzzi:
This actually makes more sense when we do incognito.

PR #297, opened 2021-11-11
Create LSAT on button click

mnicpt:
### Description
Fixes cached lsat for customers receiving 403 errors on capture.

### DIFF
--- a/src/api/auth.js
+++ b/src/api/auth.js
@@ -12,10 +12,11 @@ import { HEADERS } from '../constants';
 import { callGraphQL } from './api';
 
 type GenerateAccessTokenOptions = {|
+    cache? : boolean,
     targetSubject? : string
 |};
 
-export function createAccessToken(clientID : string, { targetSubject } : GenerateAccessTokenOptions = {}) : ZalgoPromise<string> {
+export function createAccessToken(clientID : string, { cache = true, targetSubject } : GenerateAccessTokenOptions = {}) : ZalgoPromise<string> {
     return inlineMemoize(createAccessToken, () => {
 
         getLogger().info(`rest_api_create_access_token`);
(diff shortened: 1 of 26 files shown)


src/api/auth.js
```diff
@@ -24,6 +25,9 @@ export function createAccessToken(clientID : string, { targetSubject } : Generat
         const data : Object = {
             grant_type: `client_credentials`
         };
+        if (!cache) {
```
TNuzzi, line 28: Not a functional comment just laughed at, if no then yes.

src/button/pay.js
```diff
@@ -107,8 +107,17 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
...
+        
+        const accessTokenPromise = ZalgoPromise.try(() => {
+            return createAccessToken(clientID, { cache: false })
+                .then(facilitatorAccessToken => facilitatorAccessToken);
+        });
 
-        const clickPromise = click ? ZalgoPromise.try(click) : ZalgoPromise.resolve();
+        const clickPromise = click
+            ? ZalgoPromise.try(() => {
+                return accessTokenPromise.then(click);
```
TNuzzi, line 118: I don't think it is a big deal or even an issue but `click` could return `false` and in that case we would not need to create LSAT token.  I am not sure how much we want to optimize it.

TNuzzi:
Other than my comment this looks good.

TNuzzi (approved):

src/button/pay.js
```diff
@@ -107,8 +107,17 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 [FPTI_KEY.IS_VAULT]:        instrumentType ? '1' : '0',
                 [FPTI_CUSTOM_KEY.INFO_MSG]: enableNativeCheckout ? 'tester' : ''
             }).flush();
+        
+        const accessTokenPromise = ZalgoPromise.try(() => {
```
bluepnume, line 111: Little confused, we're creating this token but then not passing it anywhere. So this is a noop right now right?

mnicpt: @bluepnume It's in the `clickPromise`.

mnicpt: I refactored this and saw that I didn't need it at all.

src/button/pay.js
mnicpt: No good point!

PR #299, opened 2021-11-15
[Feature] - Venmo App label

fedperezp:
### Description

Adding the App label into the venmo button

https://venmoinc.atlassian.net/browse/VMORECOMBO-345

| | |
| ---- | ---- |
image Pay With `horizontal` | image Pay With `vertical` |
| image Donate `Horizontal` | image Donate `Vertical` |
| image Checkout `Horizontal` | image Checkout `Vertical` |
| image Buy Now `horizontal` | image Buy Now `vertical` |

https://github.com/paypal/paypal-checkout-components/pull/1792
https://github.paypal.com/Checkout-R/smartcomponentnodeweb/pull/205

❤️  Thank you!


server/components/buttons/middleware.js
```diff
@@ -147,6 +152,7 @@ export function getButtonMiddleware({
             const wallet = await walletPromise;
             const personalization = await personalizationPromise;
             const brandedDefault = await isFundingSourceBranded(req, { clientID, fundingSource, wallet });
+            const venmoAppLabelExperiment = await getVenmoAppLabelExperiment(req, { buttonSessionID });
```
mnicpt, line 155: Would be nice to make this generic and get all experiments and pass them along rather than one at a time.  Won't scale well.

mnicpt (changes requested):

server/lib/get-experiments.js
```diff
@@ -0,0 +1,15 @@
+/* @flow */
```
mnicpt, line 1: Why do we need this if we are fetching experiments from SCNW?

fedperezp: this is to generate the default version of the experiment so we don't have issues with deploys, also so we have typings of the experiments on this file. let me know if you think this is unnecessary

mnicpt (approved):

server/lib/get-experiments.js
```diff
@@ -0,0 +1,32 @@
...
+export type GetExperimentsType = {|
+    enableVenmoAppLabel : boolean,
+    isFundingSourceBranded : boolean,
+    isCardFieldsExperimentEnabled : boolean
+|};
+
+export function getDefaultExperiments() : Promise<GetExperimentsType> {
+    return Promise.resolve({
+        enableVenmoAppLabel:           false,
+        isFundingSourceBranded:        false,
```
mnicpt, line 29: @fedperezp Is this an experiment?

fedperezp: yes, all the defaults added in here are elmo experiments

TNuzzi (approved):

PR #306, opened 2021-12-07
[EC-1478] Pass all merchantData through 

amyegan:
### Description

Pass all available `onApprove` data from Honey Checkout through to merchant via `exports`

Merchant site is expecting orderID to be be included in `onApprove` data, but it was not being passed through `exports`. This causes problems with the merchant's checkout steps.

https://joinhoney.atlassian.net/browse/EC-1478

Before
image

After
image


agnjunio (commented):
LGTM. But I wonder if there are any sensible data there that shouldn't be exported.

amyegan:
Further exploration revealed this change is not needed

PR #307, opened 2021-12-08
Card holder name component

leogedler:
### Description

Created new preact component and logic for the **optional** Card Holder's Name input for the card fields.

- https://engineering.paypalcorp.com/jira/browse/DTNOR-288

For the card fields we need an optional hosted field for the card holder's name.

The name is not PCI compliance but for the auto-fill browser autofill feature all the card fields related inputs need to be on an iframe for it to work as expected.

This new input can be defined the same way as the `number`, `expiry` and `cvv` the difference is that this input is optional and if is not defined the card fields will be still submit-able.

This new component will only be applicable for the multi card fields meaning that this field won't be able to use in the current implementation for the single card fields. 

image

This change depends on a change in `paypal-checkout-components`.

- https://github.com/paypal/paypal-checkout-components/pull/1809

❤️  Thank you!

### DIFF
--- a/src/card/components/page.jsx
+++ b/src/card/components/page.jsx
@@ -136,64 +136,64 @@ function Page({ cspNonce, props } : PageProps) : mixed {
             {
                 (type === CARD_FIELD_TYPE.SINGLE)
                     ? <CardField
-                        gqlErrorsObject={ fieldGQLErrors.singleField }
-                        cspNonce={ cspNonce }
-                        onChange={ onFieldChange }
-                        styleObject={ style }
-                        placeholder={ placeholder }
-                        autoFocusRef={ (ref) => setRef(ref.current.base) }
+                            gqlErrorsObject={ fieldGQLErrors.singleField }
+                            cspNonce={ cspNonce }
+                            onChange={ onFieldChange }
+                            styleObject={ style }
+                            placeholder={ placeholder }
+                            autoFocusRef={ (ref) => setRef(ref.current.base) }
                     /> : null
             }
 
             {
                 (type === CARD_FIELD_TYPE.NUMBER)
                     ? <CardNumberField
-                        ref={ mainRef }
-                        gqlErrors={ fieldGQLErrors.numberField }
-                        cspNonce={ cspNonce }
-                        onChange={ onFieldChange }
-                        styleObject={ style }
-                        placeholder={ placeholder }
-                        autoFocusRef={ (ref) => setRef(ref.current.base) }
+                            ref={ mainRef }
+                            gqlErrors={ fieldGQLErrors.numberField }
+                            cspNonce={ cspNonce }
+                            onChange={ onFieldChange }
+                            styleObject={ style }
+                            placeholder={ placeholder }
+                            autoFocusRef={ (ref) => setRef(ref.current.base) }
                     /> : null
             }
 
             {
                 (type === CARD_FIELD_TYPE.CVV)
                     ? <CardCVVField
-                        ref={ mainRef }
-                        gqlErrors={ fieldGQLErrors.cvvField }
-                        cspNonce={ cspNonce }
-                        onChange={ onFieldChange }
-                        styleObject={ style }
-                        placeholder={ placeholder }
-                        autoFocusRef={ (ref) => setRef(ref.current.base) }
+                            ref={ mainRef }
+                            gqlErrors={ fieldGQLErrors.cvvField }
+                            cspNonce={ cspNonce }
+                            onChange={ onFieldChange }
+                            styleObject={ style }
+                            placeholder={ placeholder }
+                            autoFocusRef={ (ref) => setRef(ref.current.base) }
                     /> : null
             }
 
             {
                 (type === CARD_FIELD_TYPE.EXPIRY)
                     ? <CardExpiryField
-                        ref={ mainRef }
-                        gqlErrors={ fieldGQLErrors.expiryField }
-                        cspNonce={ cspNonce }
-                        onChange={ onFieldChange }
-                        styleObject={ style }
-                        placeholder={ placeholder }
-                        autoFocusRef={ (ref) => setRef(ref.current.base) }
+                            ref={ mainRef }
+                            gqlErrors={ fieldGQLErrors.expiryField }
+                            cspNonce={ cspNonce }
+                            onChange={ onFieldChange }
+                            styleObject={ style }
+                            placeholder={ placeholder }
+                            autoFocusRef={ (ref) => setRef(ref.current.base) }
                     /> : null
             }
 
             {
                 (type === CARD_FIELD_TYPE.NAME)
                     ? <CardNameField
-                        ref={ mainRef }
-                        gqlErrors={ fieldGQLErrors.nameField }
-                        cspNonce={ cspNonce }
-                        onChange={ onFieldChange }
-                        styleObject={ style }
-                        placeholder={ placeholder }
-                        autoFocusRef={ (ref) => setRef(ref.current.base) }
+                            ref={ mainRef }
+                            gqlErrors={ fieldGQLErrors.nameField }
+                            cspNonce={ cspNonce }
+                            onChange={ onFieldChange }
+                            styleObject={ style }
+                            placeholder={ placeholder }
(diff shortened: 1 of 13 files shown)


src/card/lib/card-utils.js
```diff
@@ -314,6 +325,15 @@ export function setErrors({ isNumberValid, isCvvValid, isExpiryValid, gqlErrorsO
         }
     }
 
+    if (typeof isNameValid === 'boolean' &&  !isNameValid) {
+
+        if (field === CARD_FIELD_TYPE.NAME  && gqlErrors.length) {
```
nbierdeman, line 330:
```suggestion
        if (field === CARD_FIELD_TYPE.NAME && gqlErrors.length) {
```

src/card/lib/card-utils.js
```diff
@@ -314,6 +325,15 @@ export function setErrors({ isNumberValid, isCvvValid, isExpiryValid, gqlErrorsO
         }
     }
 
+    if (typeof isNameValid === 'boolean' &&  !isNameValid) {
```
nbierdeman, line 328:
```suggestion
    if (typeof isNameValid === 'boolean' && !isNameValid) {
```

test/client/card-fields.js
```diff
@@ -968,6 +964,158 @@ describe('card fields cases', () => {
 
         });
 
+        it('should render multi card fields with createOrder and optinal field card holder\'s name', async () => {
```
nbierdeman, line 967:
```suggestion
        it('should render multi card fields with createOrder and optional cardholder name field', async () => {
```

nbierdeman (changes requested):
It's looking good so far, @leogedler! :godmode: 💻

src/card/interface.js
```diff
@@ -65,17 +67,19 @@ export function getCardFields() : ?Card {
...
-        cardExpiryFrame && cardExpiryFrame.isFieldValid()
+        cardExpiryFrame && cardExpiryFrame.isFieldValid() &&
+        (cardNameFrame ? cardExpiryFrame.isFieldValid() : true)
     ) {
         return {
             number: cardNumberFrame.getFieldValue(),
             cvv:    cardCVVFrame.getFieldValue(),
-            expiry: cardExpiryFrame.getFieldValue()
+            expiry: cardExpiryFrame.getFieldValue(),
+            name:   cardNameFrame && cardExpiryFrame.isFieldValid() ? cardNameFrame.getFieldValue() : ''
```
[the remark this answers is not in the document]
leogedler: @nbierdeman thanks for noticing this!, it was an error we shouldn't mix thoses.

I have pushed a fix for it.

src/card/lib/card-utils.js
```diff
@@ -271,6 +271,17 @@ export function checkCVV(value : string, cardType : CardType) : {| isValid : boo
     };
 }
 
+export function checkName(value : string) : {| isValid : boolean, isPotentiallyValid : boolean |} {
+    let isValid = false;
+    if (value.length >= 1 && value.length <= 255) {
```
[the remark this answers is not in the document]
leogedler: @nbierdeman for now I think this is it.

@crookedneighbor told me that we shouldn't overcomplicate this and only check for at lest 1 character and just set a max length.

src/card/constants.js
```diff
@@ -148,5 +151,6 @@ export const DEFAULT_STYLE = {
 export const DEFAULT_PLACEHOLDERS : CardPlaceholder = {
     number: 'Card number',
     expiry: 'MM/YY',
-    cvv:    'CVV'
+    cvv:    'CVV',
+    name:   'Card holder\'s name'
```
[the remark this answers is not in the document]
leogedler: Changed

src/card/interface.js
```diff
@@ -65,17 +67,19 @@ export function getCardFields() : ?Card {
...
 
-    const { cardNumberFrame, cardCVVFrame, cardExpiryFrame } = getCardFrames();
+    const { cardNumberFrame, cardCVVFrame, cardExpiryFrame, cardNameFrame } = getCardFrames();
 
     if (
         cardNumberFrame && cardNumberFrame.isFieldValid() &&
         cardCVVFrame && cardCVVFrame.isFieldValid() &&
-        cardExpiryFrame && cardExpiryFrame.isFieldValid()
+        cardExpiryFrame && cardExpiryFrame.isFieldValid() &&
+        (cardNameFrame ? cardExpiryFrame.isFieldValid() : true)
```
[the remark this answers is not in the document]
leogedler: Fixed

src/card/lib/card-utils.js
leogedler: Changed

test/client/card-fields.js
leogedler: Changed

src/card/components/page.jsx
```diff
@@ -49,13 +49,15 @@ function Page({ cspNonce, props } : PageProps) : mixed {
             errorObject.expiryField = [ ...errors ];
         } else if (type === CARD_FIELD_TYPE.CVV && errors && errors.length) {
             errorObject.cvvField = [ ...errors ];
+        } else if (type === CARD_FIELD_TYPE.NAME && errors && errors.length) {
```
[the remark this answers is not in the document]
leogedler: @nbierdeman using a switch statement we cannot check for `errors && errors.length` and we need to check if that part is truty.

We could use a mix between a switch and some inner if/else but that will make it more complex so I don't think it worth it.

If you have another idea let me know.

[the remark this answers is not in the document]
leogedler: @nbierdeman I have changed this to a combination between an if/else and a switch.

src/card/lib/card-utils.js
```diff
@@ -314,6 +325,15 @@ export function setErrors({ isNumberValid, isCvvValid, isExpiryValid, gqlErrorsO
         }
     }
 
+    if (typeof isNameValid === 'boolean' && !isNameValid) {
```
crookedneighbor, line 328: Can this be simplified to:

```suggestion
    if (isNameValid === false) {
```

src/card/lib/card-utils.js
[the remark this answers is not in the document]
crookedneighbor: Yeah, the requirements for cardholder names are so loose, that we're bound to run into edge cases if we're too strict on it.

src/card/interface.js
```diff
@@ -65,17 +67,19 @@ export function getCardFields() : ?Card {
...
-        cardExpiryFrame && cardExpiryFrame.isFieldValid()
+        cardExpiryFrame && cardExpiryFrame.isFieldValid() &&
+        (cardNameFrame ? cardNameFrame.isFieldValid() : true)
     ) {
         return {
             number: cardNumberFrame.getFieldValue(),
             cvv:    cardCVVFrame.getFieldValue(),
-            expiry: cardExpiryFrame.getFieldValue()
+            expiry: cardExpiryFrame.getFieldValue(),
+            name:   cardNameFrame && cardNameFrame.isFieldValid() ? cardNameFrame.getFieldValue() : ''
```
crookedneighbor, line 82: can you explain this extra logic here?

leogedler: @crookedneighbor they way SPB gets the value for each field from the iFrames is by getting the frames which has access to the `win.exports` (zoid). 

Here is the function use for this purpose ---> https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/card/interface.js#L16

Having access to the frame then we have access for the methods defined on [Page](https://github.com/paypal/paypal-smart-payment-buttons/blob/master/src/card/components/page.jsx) for each field.

For the particular case of the `name` field which is optional I'm adding a logic to first check if the frame was defined (meaning that the field was set by the user), checking validity and if true returning the field value.

crookedneighbor: Would this syntax work in flow? If the object exists, call `getFieldValue()` and default any falsy values to `""`.

```suggestion
            name:   cardNameFrame?.getFieldValue() || ''
```

leogedler: @crookedneighbor It worked, I'm pushing a change.

nbierdeman (approved):
Great work, @leogedler! This looks ready to merge (other than if you do end up needing to make changes to accommodate GQL). 🥳

PR #316, opened 2022-01-10
chore: fix lint failures from grumbler version updates

westeezy:
### Description

We updated grumbler-scripts earlier for a color fix. For some reason it looks like some new lint errors are triggering which are valid. Not sure why they weren't caught before but easy to fix now.


wsbrunson (approved):

westeezy:
looks like its got some flake from the belter ios checks. @gregjopa & @wsbrunson  you mind if i merge through those since lint is passing?

wsbrunson:
@westeezy 👍

PR #318, opened 2022-01-11
Support iframe only render when data-popups-disabled attribute is used to initialize JS SDK

ravishekhar:
### Description

* Render Checkout UI in iframe instead of popup is data-popups-disabled attributed is passed to initialize the SDK. 
* This new "data-popups-disabled" data attribute will be used by merchants so they can use our components in non-standard browsers. For example, merchant can use SPB inside a C++ desktop application that embeds Chrome which does not support popups.

https://github.com/paypal/paypal-sdk-client/pull/99 - already merged and deployed
https://github.com/paypal/paypal-checkout-components/pull/1838 - already merged and deployed

❤️  Thank you!


gregjopa (approved):
Awesome work @ravishekhar! 💯

westeezy (commented):
lgtm

PR #319, opened 2022-01-12
Ensure dom tests have elements in body

westeezy:
ref: https://github.com/jsdom/jsdom/issues/2924#issuecomment-841831308

thank to @wsbrunson for finding the fix and @elizabethmv for noticing the failures!

We updated to jest 27, via grumbler-scripts v5, which included a new version of JSDOM where this behavior started to fail


elizabethmv (approved):

gregjopa:
Great work fixing this test to work with the latest jest @westeezy @wsbrunson @elizabethmv ! 👏

PR #332, opened 2022-01-31
key time added to fpti logs

gabrielo91:
### Description

Some events logged in Amplitude are not appearing in the right order, this is because we are not sending the key `time` in the properties of each log. We are sending the key `t` however it is got as an extra event property, not as the timestamp of the event. 

Here there is a small experiment made locally using the library [beaver-logger](https://github.com/krakenjs/beaver-logger)

[In the first case](https://analytics.amplitude.com/paypal/project/295320/search/amplitude_id%3D356432994144), logs are sent just like are sent here, without the time key, and they appear disordered... 

[Here is the result](https://analytics.amplitude.com/paypal/project/295320/search/amplitude_id%3D356438516641) when we add the key time

[Jira Ticket](https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1067)

### Screenshots (Before)

image

### Screenshots (After)

image

### Dependent Changes

Depends on https://github.com/paypal/paypal-sdk-constants/pull/80

### DIFF
--- a/src/lib/logger.js
+++ b/src/lib/logger.js
@@ -76,7 +76,6 @@ export function setupLogger({ env, sessionID, clientID, sdkCorrelationID, buyerC
             [FPTI_KEY.USER_AGENT]:             window.navigator && window.navigator.userAgent,
             [FPTI_KEY.CONTEXT_CORRID]:         sdkCorrelationID,
             [FPTI_KEY.TIMESTAMP]:              Date.now().toString(),
-            [FPTI_KEY.TIME]:                   Date.now().toString(),
             [FPTI_KEY.CHOSEN_FUNDING]:         fundingSource
         };
     });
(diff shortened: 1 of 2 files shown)


mnicpt:
Let's chat tomorrow on this.

mnicpt:
@gabrielo91 We send the TIME key here: https://github.com/paypal/paypal-smart-payment-buttons/blob/152102a7ba3699ad7a8a7fe9d6710611c306faad/src/button/logger.js#L57 We should see where and why this isn't passed along.  I've seen it in the logs.

src/lib/logger.js
```diff
@@ -76,6 +76,7 @@ export function setupLogger({ env, sessionID, clientID, sdkCorrelationID, buyerC
             [FPTI_KEY.USER_AGENT]:             window.navigator && window.navigator.userAgent,
             [FPTI_KEY.CONTEXT_CORRID]:         sdkCorrelationID,
             [FPTI_KEY.TIMESTAMP]:              Date.now().toString(),
+            [FPTI_KEY.TIME]:                   Date.now().toString(),
```
mnicpt, line 79: I'm not seeing this defined.

mnicpt (approved):
Thanks @gabrielo91 for taking this on!

PR #338, opened 2022-02-14
Removing branded vault card flow - Honey deprecated use of this flow

sioked:
### Description

Removing the branded vault card flow which is currently not in use any more.

Honey was the only user of this specific branded vault card flow. They recently deprecated this functionality in favor of a standard guest flow - apparently they weren't getting much use out of it and it ended up being a bigger challenge to maintain and led to some confusing issues. We also recently released updates to Honey XO to use a vaulted billing agreement for one-click buttons on SPB, so similar functionality for PayPal members that are logged in.

On our side, this has led to some support challenges as there are some upcoming changes to vaulting / identity that would require updates to this functionality.

See [DTCHKINT-1025](https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1025)

Using a test site with vaulting set up, initialize the payment buttons using a user-id-token and the following configuration:
```
    window.paypal
      .Buttons({
        paymentMethodNonce: pmt,
        branded: true,
```

Notice in the first screenshot below- when pointing to the staging environment there is a mutation to `approvePaymentMethodWithNonce`. This is only enabled with the specific branded guest flow that has been deleted.
Screen Shot 2022-02-15 at 3 04 35 PM

Note that when running against the local environment, instead of calling the `approvePaymentMethodWithNonce` flow, we are calling `validate-payment-method` which is the standard one-click payment flow. Note that it failed because this merchant is not set up to accept credit cards at this moment.
Screen Shot 2022-02-15 at 3 05 23 PM


amyegan (approved):
lgtm. Are you removing the `approvePaymentWithNonce` gql mutation too?

remotevision (approved):
👍

sioked:
@amyegan Yep, I'll have a PR out eventually to remove that mutation as well. Currently can see that there's 0 traffic on it but want to get rid of the SDK flow first.

PR #344, opened 2022-02-24
add unprocessable entity error handler

gabrielo91:
### Description

User is receiving HTTP 500 "Unhandled api error" after HTTP 422 "DUPLICATE_INVOICE_ID" on PayPal checkout.

image

image

The client needs to handle the error as follows 

```javascript
onError: function(err) {
  err.data.details.forEach(function(detail) {
    if (detail.issue == "DUPLICATE_INVOICE_ID") {
      // Catch duplicate transaction
      console.log('Duplicate Invoice ID')
    }
  })
}
```

However, due to the current behavior, when the call to order API fails, the fallback uses `callSmartAPI` [on the catch expresion](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/api/order.js#L152) no matter what the error calling when calling `ORDERS_API_URL` was, so the user is getting always a 500 error without the information he needs.

This PR fixes this issue by adding an extra validation `isUnprocessableEntity`, which, if true, throws the error instead of using the fallback calling to smart API orders endpoint.

Additionally, we passed the `onError` callback to the `executePayment` in order to allow it to launch the function with the response error.

[Jira Ticket](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-437)

Reproduction steps are available in [Jira ticket](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-437)

### Evidence:

The change works properly, the error is captured by onError callback as expected:

image

### Collaborators:

@garbray


src/api/order.js
```diff
@@ -126,6 +126,15 @@ export function isProcessorDeclineError(err : mixed) : boolean {
     }));
 }
 
+
+export function isUnprocessableEntity(err : mixed) : boolean {
```
gregjopa, line 130:
```suggestion
export function isUnprocessableEntityError(err : mixed) : boolean {
```

src/props/onApprove.js
```diff
@@ -98,16 +98,24 @@ const redirect = (url) => {
...
     return redir(url, window.top);
+
+
 };
 
-const handleProcessorError = <T>(err : mixed, restart : () => ZalgoPromise<void>) : ZalgoPromise<T> => {
+const handleProcessorError = <T>(err : mixed, restart : () => ZalgoPromise<void>, onError : OnError) : ZalgoPromise<T> => {
+
+    if (isUnprocessableEntity(err)) {
+        return onError(err).then(unresolvedPromise);
```
gregjopa, line 108:
```suggestion
        return onError(err);
```

src/props/onApprove.js
```diff
@@ -166,9 +177,10 @@ type PaymentActionOptions = {|
     buyerAccessToken : ?string,
     partnerAttributionID : ?string,
     forceRestAPI : boolean
+,    onError : OnError
```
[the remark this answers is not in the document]
gabrielo91: done, thanks!

src/props/onApprove.js
gabrielo91: actually, we can't , since the function return type is `ZalgoPromise<T>` , as you can see here https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/onApprove.js#L105 the other error is handled in the exact same way

src/api/order.js
gabrielo91: changed, thanks!

wsbrunson (approved):

src/props/onApprove.js
gregjopa: I see. Good point. Thanks for explaining.

src/props/onApprove.js
```diff
@@ -98,16 +98,24 @@ const redirect = (url) => {
     }
 
     return redir(url, window.top);
+
+
```
gregjopa, line 102:
```suggestion
```

src/api/order.js
```diff
@@ -126,6 +126,15 @@ export function isProcessorDeclineError(err : mixed) : boolean {
...
 }
 
+
+export function isUnprocessableEntityError(err : mixed) : boolean {
+    // $FlowFixMe
+    return Boolean(err?.response?.body?.data?.details?.some(detail => {
+        return  detail.issue === ORDER_API_ERROR.DUPLICATE_INVOICE_ID;
+    }));
+}
+
```
gregjopa, line 136:
```suggestion
```

PR #353, opened 2022-03-17
Remove `npm run release:pre`

nbierdeman:
### Purpose

We no longer need the `npm run release:pre` command, since a new version of **grabthar-release** was cut. Dist-tag and version auto-selection got merged, and `alpha` releases will now be handled via the `npm run release` command: https://github.com/krakenjs/grabthar-release/pull/26

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "release": "npm run build-commit && grabthar-release",
     "release:cli": "rm -rf node_modules && npm i && npm run build-commit && grabthar-release",
     "postrelease": "npm run cdnify -- --commitonly",
-    "release:pre": "npm run build-commit && grabthar-release --DIST_TAG=alpha --BUMP=prerelease",
     "deploy": "npm run cdnify",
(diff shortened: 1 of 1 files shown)


TNuzzi:
@nbierdeman Can you explain this a little bit more?  I am assuming from your comment that `grabthar` will use the branch to determine if it is a release or alpha correct?  My only hesitation is the hope nothing breaks in `grabthar` and a test branch does not get published to `latest`.

nbierdeman:
@TNuzzi, yeah, grabthar-release will use the current branch name to compare against the repo's default branch name to determine if it's an `alpha` or `latest` release. It also uses the GitHub Repos API to find out if the repo has been forked, and blocks it if it has (this got me yesterday 😄 ). These features are already implemented in grabthar-release, so this PR is just to remove an unnecessary/dead command.

TNuzzi:
@nbierdeman Cool you answered my follow-up question about forks.

PR #371, opened 2022-04-05
onComplete Callback

mnicpt:
### Description
Adds an `onComplete` callback for flow and APM's where `commit=true` and PayPal captures payment.


test/client/inlinexo.js
```diff
@@ -0,0 +1,242 @@
...
+                    eligible: true
+                },
+                [ CARD.MASTERCARD ]: {
+                    eligible: true
+                }
+            }
+        }
+    };
+
+    it('should call onComplete if experience is inline', async () => {
```
gregjopa, line 32: Could we add another test for the opposite case? `should not call onComplete when the experience is not inline`

gregjopa (approved):
PR looks good to me! I left one comment about testing that the `onComplete()` callback is not called for non-inline experiences.

mnicpt:
@gregjopa Will add test.

test/client/inlinexo.js
mnicpt: @gregjopa `onComplete` can also be used for any APM like Apple Pay or Google Pay where we capture or authorize for the merchant as well. It can also be use for merchants with commit=true if they wanted to.

PR #381, opened 2022-04-15
add shipping flags to get checkout details

andyfen:


wsbrunson (approved):

andyfen:
@wsbrunson  hold up on this please, just working on a bit more to come.

src/api/order.js
```diff
@@ -682,6 +682,11 @@ export const getSupplementalOrderInfo : GetSupplementalOrderInfo = memoize(order
 
 export type DetailedOrderInfo = {|
     checkoutSession : {|
+        flags : {|
+            isShippingAddressRequired : boolean,
+            isDigitalGoodsIntegration : boolean,
```
JianyiGao, line 687: why do we need isDigitalGoodsIntegration flag?

andyfen: it may be needed later if we extend the functionality to cover that, i just focused on NO_SHIPPING for this for now, i think it's worth keeping it in the payload so it's available / known that it's there even if its currently not used.

JianyiGao (approved):

PR #382, opened 2022-04-20
Registering a service worker in the Paypal button

JoseFdri:
### Description
This PR is for registering a service worker during the right away after the button renders, if `eligibility.isServiceWorkerEligible` is true then the service worker will be registered otherwise it will be unregistered.

Ref: https://github.com/paypal/paypal-smart-payment-buttons/pull/382/files#diff-05448c44a11aec8c317ff422def5b0f7b332d62b0234a10ea45f0021e2c28b14R277

The service worker will cache all the assets from the Paypal checkout Review page in order to have almost zero latency when it comes to load those assets.

When the user clicks on the button, an event is emitted to request the service worker logs and then publish them to the logs platform.

Flow:

1. The service worker will receive a release hash via query string parameter.
2. This release hash parameter is used to get the current assets manifest url from the release meta json.
3. The SW deletes old caches names prior the precaching process.
4. The SW is using BroadcastChannel to communicate with the main thread.
5. Once all the precache process is done the SW flush the logs via BroadcastChannel.
6. Main thread (Client) is listening to GET_SW_LOGS_RESPONSE events to post the logs to the server.

https://engineering.paypalcorp.com/confluence/display/BEAM/App+Shell

https://engineering.paypalcorp.com/confluence/display/BEAM/Set+up+the+service+worker

### BEFORE

image

### AFTER

1. Assets loading from the service worker.
CleanShot 2022-09-07 at 18 02 48

2. Service worker successfully installed in the browser.
CleanShot 2022-09-07 at 18 03 47

3. Assets cached successfully.
CleanShot 2022-09-07 at 18 04 43

https://github.paypal.com/Checkout-R/smartcomponentnodeweb/pull/208

❤️  Thank you!


src/button/button.js
```diff
@@ -232,7 +234,15 @@ export function setupButton(opts : SetupButtonOptions) : ZalgoPromise<void> {
     const setupExportsTask = setupExports({ props, isEnabled, facilitatorAccessToken, fundingEligibility, merchantID });
 
     const validatePropsTask = setupButtonLogsTask.then(() => validateProps({ env, clientID, intent, createBillingAgreement, createSubscription }));
-
+    
+    if (eligibility.isServiceWorkerEligible) {
+        getLogger().info(`SERVICE_WORKER_ELIGIBLE`).flush();
```
wsbrunson, line 239: can we remove the `.flush()`? this will start a new network request instead of bundling up with our other log calls

src/button/button.js
```diff
@@ -232,7 +234,15 @@ export function setupButton(opts : SetupButtonOptions) : ZalgoPromise<void> {
...
+        registerServiceWorker(opts.currentReleaseHash);
+    } else {
+        getLogger().info(`SERVICE_WORKER_NOT_ELIGIBLE`).flush();
```
wsbrunson, line 242: same here about flush

src/constants.js
```diff
@@ -282,3 +282,16 @@ export const BUTTON_LABEL = {
 export const STATUS_CODES = {
     TOO_MANY_REQUESTS: 429
 };
+
+export const SERVICE_WORKER = {
+    PROD_CHECKOUTWEB_CDN_DUMBLEDORE: 'https://www.paypalobjects.com/checkoutweb',
+    DEFAULT_DIR:                     'release',
+    DUMBLEDORE_APP:                  'dumbledore',
+    RELEASE_CONFIG_FILE_NAME:        'release-config.json',
+    // SW_URL:                          'https://localhost.paypal.com:8443/checkoutweb/public/dumbledore-service-worker.js?',
```
wsbrunson, line 291:
```suggestion
```

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,125 @@
...
+
+const listenBroadCastChannelEvents = () => {
+    // Listen SW flush logs event
+    broadcastChannel.addEventListener('message', (event : Event) => {
+        // $FlowFixMe
+        const { payload = [], eventName } = event.data;
+        if (payload && eventName === GET_SW_LOGS_RESPONSE_EVENT_NAME) {
+            getLogger().info(`${ LOG_PREFIX }LOGS`, {
+                logs: JSON.stringify(payload)
+            }).flush();
```
wsbrunson, line 48: can we remove all the `.flush()` in this file?

src/lib/register-service-worker.spec.js
```diff
@@ -0,0 +1,74 @@
+// eslint-disable-next-line eslint-comments/disable-enable-pair
```
wsbrunson, line 1: we name our test files `.test.js`

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,146 @@
...
+    });
+}
+
+export function registerServiceWorker(currentReleaseHash?: string) {
+    if ('serviceWorker' in navigator === false) {
+        getLogger().info(`${ LOG_PREFIX }NOT_SUPPORTED`);
+        return;
+    }
+    if (!currentReleaseHash) {
+        getLogger().error(`${ LOG_PREFIX }RELEASE_HASH_NO_PROVIDED`, { releaseHash: currentReleaseHash });
```
wsbrunson, line 126:
```suggestion
        getLogger().error(`${ LOG_PREFIX }RELEASE_HASH_NOT_PROVIDED`, { releaseHash: currentReleaseHash });
```

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,146 @@
...
+        const { payload = [], eventName } = event.data;
+        if (payload && eventName === GET_SW_LOGS_RESPONSE_EVENT_NAME) {
+            getLogger().info(`${ LOG_PREFIX }LOGS`, {
+                logs: JSON.stringify(payload)
+            });
+        }
+    });
+};
+
+const handleButtonClick = () => {
```
wsbrunson, line 62: I would consider calling this `registerButtonHandlers` or `attachServiceWorkerHandlersToPayPalButtons`

I would also suggest making an `unregister` or `detach` version in case there is an issue during setup and we've registered the event handlers but then failed to instantiate the service worker

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,146 @@
...
+        startRegistration(swUrl);
+    })
+    .catch((err) => {
+        getLogger().info(`${ LOG_PREFIX }FETCH_RELEASE_META_ERROR`, {
+            error: err.message
+        });
+    });
+}
+
+export function registerServiceWorker(currentReleaseHash?: string) {
```
wsbrunson, line 120: we have `register` and `registerServiceWorker`. can we call one `registerServiceWorker` and one `executeSerivceWorker` or something like that? Making it clear that its not just registering the service worker but adding event handlers and setting up broadcast channels?

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,146 @@
...
+        handleButtonClick();
+        startBroadCastChannel();
+        listenBroadCastChannelEvents();
+        register(currentReleaseHash);
+    } catch (err) {
+        getLogger().error(`${ LOG_PREFIX }ERROR_DURING_INITIALIZATION`, { err: stringifyError(err) });
+    }
+}
+
+export function unregisterServiceWorker() {
```
wsbrunson, line 139: down here we could `detach` the button listeners

wsbrunson (approved):

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,155 @@
...
+    // eslint-disable-next-line compat/compat
+    return fetch(`${RELEASE_META_PATH}.${releaseHash}.json`)
+        .then((resp) => {
+            return resp.json();
+        });
+}
+
+const startBroadCastChannel = () => {
+    // eslint-disable-next-line compat/compat
+    broadcastChannel = new BroadcastChannel(LOGS_CHANNEL_NAME);
```
gregjopa, line 46: @JoseFdri Can we add some feature detection to make sure this code does not blow up in older browsers that do not support `fetch` or `BroadcastChannel`?

JoseFdri: Hi @gregjopa , I added a try catch to handle errors like this.
Ref: https://github.com/paypal/paypal-smart-payment-buttons/pull/382/files#diff-f6fa240eb8c58c1af2dc4729631bd5b7659393ca1c9ea00fc05175b0c38ad51eR137

gregjopa: @JoseFdri great to see.

src/lib/register-service-worker.js
```diff
@@ -0,0 +1,155 @@
...
+                logs: JSON.stringify(payload)
+            });
+        }
+    });
+};
+
+const registerButtonHandlers = () => {
+    const paypalButtons = document.getElementsByClassName('paypal-button');
+    for (let i = 0; i < paypalButtons.length; i++) {
+        paypalButtons[i].addEventListener('click', requestSwLogs);
```
gregjopa, line 65: Two questions for you about this code path:
1. Can we avoid hardcoding the `paypal-button` class name here? I believe its defined in this constants file: https://github.com/paypal/paypal-checkout-components/blob/53ddb88f56f7878922eee2f8f866a6761218d100/src/constants/class.js#L6. Not sure if we can leverage that or if we should leverage the constants file in this existing repo (https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/constants.js#L45).
2. Does this code fetch the service-worker and asset files on button click? If so, can you update your PR description to mention that? I'd like to make sure the PR description accurately describes what happens on button render vs button click.

JoseFdri: Hi @gregjopa, adding a new constant in the existing looks good to me, I added more information for what happens on the button click.
```
When the user clicks on the button, an event is emitted to request the service worker logs and then publish them to the logs platform.
```
Fetching the service worker and asset files only happens when the service worker is being installed, that installation will happen during button setup
When the user clicks on the button, an event is emitted to request the service worker logs and then publish them to the logs platform.

gregjopa: >Fetching the service worker and asset files only happens when the service worker is being installed, that installation will happen during [button setup](https://github.com/paypal/paypal-smart-payment-buttons/pull/382/files#diff-05448c44a11aec8c317ff422def5b0f7b332d62b0234a10ea45f0021e2c28b14R61)

Thanks for the info. I'm wondering how many additional network requests we are adding with this new behavior on render. Is it just one network request to load the service worker on button render? Or are there more?

JoseFdri: Network requests:

1. Fetch the release meta json in order to get the service worker url
2. Fetch the service worker file.
3. I'm not manually flushing the logs, so it may happen automatically.

gregjopa: > Fetch the release meta json in order to get the service worker url

@JoseFdri can we move this one to the server-side and embed the meta json contents into the html returned by /smart/buttons? The we'd be able to cut this down to only one additional request on the client-side for button render.

JoseFdri: @gregjopa  I opened a [PR](https://github.paypal.com/Checkout-R/smartcomponentnodeweb/pull/362) for calling the release meta on `smartcomponentnodeweb` and passing it through the button params.

PR #384, opened 2022-04-20
chore: ensure beaver logger uses krakenjs scope

westeezy:
After this PR we can see the eradication of the non namespaced zalgo-promise with the following:

```
❯ npm ls zalgo-promise
@paypal/smart-payment-buttons@5.0.95
└── (empty)
```


gregjopa (approved):
💯

PR #387, opened 2022-04-21
update applepay utils

andyfen:
update applepay utils file + tests

- add default params to handle undefined on mapping functions
- use optional chaining
- update to es6 methods


src/payment-flows/applepay/utils.js
```diff
@@ -15,29 +19,23 @@ type ValidNetworks = {|
...
+        jcb:            'jcb'
+    };
 
     const validIssuers = [];
+
     function validateIssuers(issuer : string) : ?ApplePaySupportedNetworks {
         const network = issuer.toLowerCase().replace(/_/g, '');
-        if (Object.keys(validNetworks).indexOf(network) !== -1) {
+
+        if (Object.keys(validNetworks).includes(network)) {
```
mnicpt, line 38: @andyfen Make sure we are supporting `includes` in this repo.  I remember using before and recall it wasn't.

src/payment-flows/applepay/utils.js
```diff
@@ -77,32 +75,24 @@ function getShippingContactFromAddress(shippingAddress : ?ShippingAddress) : App
...
-    }
-
-    const result = [ ...shippingMethods ].sort(method => {
+export function getApplePayShippingMethods(shippingMethods : $ReadOnlyArray<ShippingMethod> = []) : $ReadOnlyArray<ApplePayShippingMethod> {
+    return [ ...shippingMethods ].sort(method => {
         return method.selected ? -1 : 0;
     }).map(method => {
         return {
-            amount:     method.amount && method.amount.currencyValue ? method.amount.currencyValue : '0.00',
+            amount:     method?.amount?.currencyValue || '0.00',
```
mnicpt, line 83: @andyfen Also just double-check on optional chaining as well.

src/payment-flows/applepay/utils.js
andyfen: this code is only ever ran on safari browsers supporting applepay version 4, so we are all good for the es6 stuff.

src/payment-flows/applepay/utils.js
andyfen: based on the use case safari only we are good with support for this

andyfen: Screen Shot 2022-04-22 at 12 04 31 PM
safari relative usage is 99.77%

PR #390, opened 2022-04-22
Proposal to notify parent.onError

borodovisin:
### Description
The PR contains the changes to notify the `window.xprops.onError` when the server response fails in order to notify the JS SDK on the source client site.

The idea is to catch any error on the server and return an HTML page containing the JS SDK to enable communication between the requested resources and the parent window through the `xprops.onError` mechanism from Zoid.

Please refer to this [ticket](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-725)

❤️  Thank you!


server/lib/util.js
```diff
@@ -264,3 +275,69 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+}
+
+/**
+ * Handler for server error. If the `meta.getSDKLoader` function is not available
+ * the response will be a text and not an HTML page
+ *
+ * @param {ErrorArgument} param0 the needed information to track and handle the errors
+ * @returns the response error
+ */
+export function serverErrorHandler ({ req, res, err, logger, meta } : ErrorArgument) : void {
```
gregjopa, line 321: I'd like to suggest renaming this function to `errorCodeHandler()`, assuming that's a pattern we want to extend. We do have `req.query.sdkMeta` available to us from the `req` param so I'd like to avoid having to pass `meta` though. Instead, the error handler could try unpacking and validating it and then calling onError() when it's valid.

```suggestion
export function errorCodeHandler ({ req, res, err, logger } : ErrorArgument) : void {
```

server/lib/util.js
```diff
@@ -264,3 +275,69 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+    if (typeof meta?.getSDKLoader !== 'function') {
+        throw new TypeError('Missing "getSDKLoader" function in meta');
+    }
+    const cspNonce = getCSPNonce(res);
+    const pageHTML = `<!DOCTYPE html>
+        <head></head>
+        <body data-nonce="${ cspNonce }">
+            ${ meta.getSDKLoader({ nonce: cspNonce }) }
+            <script nonce="${ cspNonce }">
+                window?.xprops?.onError("${ err.message }");
```
gregjopa, line 307:
```suggestion
                if (window.xprops && xprops.onError) {
                    window.xprops.onError(err.message);
                }
                
```

server/lib/util.js
```diff
@@ -264,3 +275,69 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+            </script>
+        </body>`;
```
gregjopa, line 309:
```suggestion
        </body></html>`;
```

PR #393, opened 2022-04-25
[no merge] remove applepay errorhandling for shipping fields

andyfen:
out the box 
```
requiredShippingContactFields: [
            'postalAddress',
            'name',
            'phone',
            'email'
        ],
```
gives required field validation, so there is no need to have additional checks individually for these fields.


andyfen:
closing as it looks like we do actually need individual checks when loaded from checkout session

PR #395, opened 2022-04-26
applepay fix flow issues

andyfen:
fix some flow issues casting undeclared types to any.


src/payment-flows/applepay/applepay.js
```diff
@@ -98,22 +98,22 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
         if ((shipping && !isZeroAmount(shipping)) || isPickup) {
             newLineItems.push({
                 label:  shippingLabel || 'Shipping',
-                amount: shipping
+                amount: shipping || '0.00'
```
mnicpt, line 101: @andyfen Seems like this would read better like this: `amount: isPickup ? '0.00' : shipping`

src/payment-flows/applepay/applepay.js
```diff
@@ -138,7 +138,7 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
                 return ZalgoPromise.resolve(update);
             }
 
-            const { errors, shipping_address } = validateShippingContact(shippingContact);
+            const { errors, shipping_address } = validateShippingContact(shippingContact || {});
```
mnicpt, line 141:
```suggestion
            const { errors, shipping_address } = validateShippingContact(shippingContact) || {};
```

src/payment-flows/applepay/applepay.js
andyfen: shipping would still be optional 
Screen Shot 2022-04-27 at 11 35 01 AM
and need a fallback still to work
Screen Shot 2022-04-27 at 11 35 31 AM
problem i see also is that would make the assumption pickup is 0.00, it might possibly not be that

andyfen: shipping would still be optional 
Screen Shot 2022-04-27 at 11 35 01 AM
and need a fallback still to work
Screen Shot 2022-04-27 at 11 35 31 AM
problem i see also is that would make the assumption pickup is 0.00, it might possibly not be that

src/payment-flows/applepay/applepay.js
andyfen: because `shippingContact` could be undefined it needs this or its checking for object values on undefined  https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/payment-flows/applepay/utils.js#L202-L236

src/payment-flows/applepay/applepay.js
mnicpt: But the condition checks for not a zero amount so shouldn't happen: `!isZeroAmount(shipping)`

src/payment-flows/applepay/applepay.js
mnicpt: Then `validateShippingContact` should handle it and return undefined and then it will go to `{}`. If not that path we should be setting default values in the method params and not passed into the method.

src/payment-flows/applepay/applepay.js
andyfen: and the value would be overwritten to '0.00' even if it wasn't that

src/payment-flows/applepay/applepay.js
andyfen: i've made `shippingContact`optional and unwrapped the values in `validateShippingContact` fn

src/payment-flows/applepay/applepay.js
mnicpt: Condition 1: If shipping is truth and it is NOT a zero amount, then the amount is the shipping amount.
Condition 2: If NOT shipping OR if shipping AND IS a zero amount, then the amount is 0.00.
Condition 3:  If shipping is truth and it is NOT a zero amount AND isPickup is true, then show shipping amount. **(this is incorrect)**
Condition 4: If NOT shipping OR if shipping AND IS a zero amount AND isPickup is true, then show 0.00.

Because Condition 3 could happen, and the result would be to show the shipping, I think we need to separate this out into two conditions or rewrite the condition like this and have the `isZeroAmount` return false if shipping is null or undefined.

```
if (isZeroAmount(shipping) || isPickup) {
  ...
  amount: '0.00'
} else {
  amount: shipping
}
```

andyfen: by that if it is a zero amount it will be displayed, we don't want to display zero amounts unless it is PICKUP

andyfen: by that if it is a zero amount it will be displayed, we don't want to display zero amounts unless it is PICKUP

andyfen:
```
amount: isPickup || isZeroAmount(shipping) ? '0.00' : shipping;
```
this is my point, pickup amount may not be '0.00' we should take the value given and use shipping rather then hardcode it to '0.00'

andyfen: this has diverged away from the intended changes which was to fix the types and make the non optionals optional which required adding a fallback value.

src/payment-flows/applepay/applepay.js
```diff
@@ -98,22 +98,22 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
...
-        let currentShippingMethod;
+        let currentTotalAmount : ?string;
+        let currentSubtotalAmount : ?string;
+        let currentTaxAmount : ?string;
+        let currentShippingAmount : ?string;
+        let currentShippingContact : ?ApplePayPaymentContact;
+        let currentShippingMethod : ?ApplePayShippingMethod;
 
-        const onShippingChangeCallback = <T>({ orderID, shippingContact, shippingMethod = null } : {| orderID : string, shippingContact : ApplePayPaymentContact, shippingMethod? : ?ApplePayShippingMethod |}) : ZalgoPromise<T> => {
+        const onShippingChangeCallback = <T>({ orderID, shippingContact, shippingMethod = null } : {| orderID : string, shippingContact : ?ApplePayPaymentContact, shippingMethod? : ?ApplePayShippingMethod |}) : ZalgoPromise<T> => {
```
mnicpt, line 116: @andyfen Does Apple ever not send these? `shippingContact or shippingMethod`

src/payment-flows/applepay/applepay.js
mnicpt: @andyfen Want to make sure I understand what the requirements are behind this so I can review correctly.  Might be good to include in the future so I'm not trying to figure them out.  The requirements have changed quite a bit since I worked on this so will need help staying up-to-date.

mnicpt (approved):

PR #404, opened 2022-05-13
update flow types applepay

andyfen:
- remove `$FlowFixMe`'s

- Consolidate `ApplePayShippingMethodUpdate` / `ApplePayShippingContactUpdate` into a single `ApplePayShippingUpdate` type
removes need for generics in `onShippingChangeCallback`

Info:

`ApplePayShippingMethodUpdate`
https://developer.apple.com/documentation/apple_pay_on_the_web/applepayshippingmethodupdate

```
dictionary ApplePayShippingMethodUpdate {
    required [ApplePayLineItem] newTotal;
    sequence <ApplePayLineItem> newLineItems;
    sequence <ApplePayShippingMethod> newShippingMethods;
};
```

`ApplePayShippingContactUpdate`
https://developer.apple.com/documentation/apple_pay_on_the_web/applepayshippingcontactupdate

```
dictionary ApplePayShippingContactUpdate {
    required [ApplePayLineItem] newTotal;
    sequence <ApplePayLineItem> newLineItems;
    sequence <ApplePayError> errors;                       // Not required on ApplePayShippingMethodUpdate
    sequence <ApplePayShippingMethod> newShippingMethods;
};
```

New common `ApplePayShippingUpdate` type
```
export type ApplePayShippingUpdate = {|
    newShippingMethods? : $ReadOnlyArray<ApplePayShippingMethod>,
    newTotal : ApplePayLineItem,
    newLineItems? : $ReadOnlyArray<ApplePayLineItem>
    errors? : $ReadOnlyArray<ApplePayError>,
|};
```

- updates the onshippingchange data


src/payment-flows/applepay/applepay.js
```diff
@@ -117,13 +118,13 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
         let currentShippingContact : ?ApplePayPaymentContact;
         let currentShippingMethod : ?ApplePayShippingMethod;
 
-        const onShippingChangeCallback = <T>({ orderID, shippingContact, shippingMethod = null, callbackTrigger } : {| orderID : string, shippingContact : ?ApplePayPaymentContact, shippingMethod? : ?ApplePayShippingMethod, callbackTrigger : string |}) : ZalgoPromise<T> => {
+        const onShippingChangeCallback = ({ orderID, shippingContact, shippingMethod = null, callbackTrigger } : {| orderID : string, shippingContact : ?ApplePayPaymentContact, shippingMethod? : ?ApplePayShippingMethod, callbackTrigger : string |}) : ZalgoPromise<ApplePayShippingUpdate> => {
```
mnicpt, line 121: @andyfen Just FYI, this `callbackTrigger` will not be in the new spec.  For now this will work as it will be a pass-through so ok with it.  We will be moving to 2 separate callbacks with this current onShippingChange being backward compatibility option.

src/payment-flows/types.js
```diff
@@ -135,7 +135,7 @@ export type ShippingMethod = {|
     |},
     label : string,
     selected : boolean,
-    type : 'SHIPPING' | 'PICKUP'
+    type : SHIPPING_OPTION_TYPE
```
mnicpt, line 138: 👍

src/payment-flows/applepay/applepay.js
```diff
@@ -138,18 +139,17 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
...
 
-                // $FlowFixMe
                 return ZalgoPromise.resolve(update);
             }
 
             const { errors, shipping_address } = validateShippingContact(shippingContact);
             
             if (errors && errors.length) {
-                const update = {
+                const update : ApplePayShippingUpdate = {
```
mnicpt, line 148: @andyfen `ApplePayShippingMethodUpdate` and `ApplePayShippingContactUpdate` were to define Apple's type.  Why are we combining to one?  We adapt Apple's interface to PayPal's requirements. Not opposed, just wondering the thought process.

src/payment-flows/applepay/applepay.js
```diff
@@ -117,13 +118,13 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
         let currentShippingContact : ?ApplePayPaymentContact;
         let currentShippingMethod : ?ApplePayShippingMethod;
 
-        const onShippingChangeCallback = <T>({ orderID, shippingContact, shippingMethod = null, callbackTrigger } : {| orderID : string, shippingContact : ?ApplePayPaymentContact, shippingMethod? : ?ApplePayShippingMethod, callbackTrigger : string |}) : ZalgoPromise<T> => {
```
mnicpt, line 120: @andyfen What is the reason to remove the generic?

andyfen: simpler is better it was not needed.

src/payment-flows/applepay/applepay.js
andyfen: because they share they share the same interface with the exception of errors which is optional on one and not on the other a single type can be used to handle it.

weihou (approved):

src/payment-flows/applepay/applepay.js
mnicpt: @andyfen I'm ok with this for now.  The reason these types were created, although minimal differences, was to ensure accurate shape def since we were adapting Apple's interface to ours.  We just want to keep an eye on this when we update versions of Apple Pay to make sure they don't become too different.  My guess is they won't, but still to keep an eye on it.  Also, I believe these are typed in checkout components as well and may want to keep them in sync.

src/payment-flows/applepay/applepay.js
mnicpt: @andyfen As related to above comment, I wouldn't do things always for simpler when it comes to contracts.  In this case it is ok because the contract differs minimally.

src/props/onShippingChange.js
```diff
@@ -9,18 +9,38 @@ import { getLogger } from '../lib';
...
 import type { CreateOrder } from './createOrder';
 
-type SHIPPING_OPTION_TYPE = 'SHIPPING' | 'PICKUP';
+export type SHIPPING_OPTION_TYPE = 'SHIPPING' | 'PICKUP';
 export type XOnShippingChangeDataType = {|
     orderID? : string,
+    callbackTrigger? : string,
     paymentID? : string,
     paymentToken? : string,
+    amount? : {|
```
mnicpt, line 18: @andyfen This is not required to pass the `amount`.  Anything passed from the app flows through to the merchant.  Dumbledore currently does this.  I'm not opposed, it's just not needed.  Same with `callbackTrigger`.  This also won't enforce anything with the application unless the application has typings in place. Again, non-blocking comment.

PR #405, opened 2022-05-19
Inline payment fields

AnshuChaudhari:
### Description
Branded to Unbranded experience migration via SPB vertical stack.

APMs have branded (PayPal logos / styles / guest user creation) flow via smart payment buttons.
We also built [unbranded](https://developer.paypal.com/docs/checkout/apm/) (no PayPal logos / configurable style / inline payment fields)  flow through [standalone buttons](https://developer.paypal.com/docs/multiparty/checkout/standard/customize/standalone-buttons/).

Our product requirement is to move away from branded flows and bring in an inline checkout experience for APMs that we built for unbranded(existing) to vertical stack as well.

- Go to `[go/smartbuttons](http://10.176.21.113:8000/)` add the custom stage - `te-apm-unbranded-vertical`
- Select country as Austria and integration tye as OrdervsV2
- Hit Refresh Button
- Click on eps funding source once vertical stack buttons are rendered
- Inline-payment-fields should render
- Fill out the inline form
- Hit Continue, it will go to the simulator

**Before (Branded/Guest flow):**

Branded-eps

### ----------------------------------------------------------------------------------------------
**After (Branded flow migrated to unbranded flow):**

Branded-to-Unbranded-migration

**Animations: Cross button**

Cross-button

**NOTE:** **_Following PRs should go in before current one._**

`paypal-checkout-components` - https://github.com/paypal/paypal-checkout-components/pull/1944
`smartcomponentnodeweb` - https://github.paypal.com/Checkout-R/smartcomponentnodeweb/pull/259
`altpayfieldsnodeweb` - https://github.paypal.com/DigitalGoods-Payments-R/altpayfieldsnodeweb/pull/245

❤️  Thank you!


src/payment-flows/payment-fields.js
```diff
@@ -0,0 +1,262 @@
...
+}
+
+function isPaymentFieldsPaymentEligible({ payment } : IsPaymentEligibleOptions) : boolean {
+    const { win, fundingSource } = payment || {};
+
+    if (win) {
+        return false;
+    }
+
+    if (fundingSource && !INLINE_PAYMENT_FIELDS_APM_LIST.includes(fundingSource)) {
```
weihou, line 61: we need to replace this logic with the elmo experiment result.

AnshuChaudhari: @weihou 
This is[ Elmo experiment](https://internal.msmaster.qa.paypal.com/elmo/experiment/103531) I am using right now for testing. 
It has treatment 100% allocated and I tried adding a custom filter for `fundingsource` to see if it works in our case 

So how this Elmo works is,  the setup is done in `smartcomponentnodeweb` for e.g. [inlineguest exp](https://github.paypal.com/Checkout-R/smartcomponentnodeweb/blob/master/lib/getInlineGuestExperiment.js) and we access that in `spb` via server/components/buttons/middleware  [here](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/server/components/buttons/middleware.js#L151).  

The actual Elmo call happens at the time of 2nd render of vertical stack buttons, so we have the result of Elmo before we click any specific button from vertical stack. In this case we can use Elmo result by whitelisting the merchant.   

But when it comes to handling the enabling of APM method (say eps) as well via Elmo, we will need to pass the funding source specifically and as the Elmo call happens before we click on a specific button therefore the fundingsource remains `undefined` at the time of Elmo call and we get a `false` in the result as the filter added in Elmo expected `eps` value for fundingsource in elmo filter parameters.   

I don’t think we can do an Elmo api call after the user clicks on the button from vertical stack.  

So one option is we keep this `INLINE_PAYMENT_FIELDS_APM_LIST` check as is and use the Elmo exp result to enable for the whitelisted merchant. 

The other option is we somehow get a list of eligible funding sources for e.g’ [ ‘PayPal’, ‘eps’, ‘sofort’, ‘card’] in `smartcomponentnodeweb` and do multiple Elmo call and check against every funding source the merchant is eligible for. 

 If there are any other ways or suggestions you have please let me know. 
Also, any idea how remote config works with smart-payment-buttons repo?

src/types.js
```diff
@@ -40,40 +41,40 @@ export type ZoidComponent<P> = {|
...
-    onShippingAddressChange : ?(data : OnShippingAddressChangeData, {| resolve : () => ZalgoPromise<void>, reject : (string) => ZalgoPromise<void> |}) => ZalgoPromise<void> | void,
-    onShippingOptionsChange : ?(data : OnShippingOptionsChangeData, {| resolve : () => ZalgoPromise<void>, reject : (string) => ZalgoPromise<void> |}) => ZalgoPromise<void> | void,
+    onShippingChange? : ?(data : OnShippingChangeData, {| resolve : () => ZalgoPromise<void>, reject : (string) => ZalgoPromise<void> |}) => ZalgoPromise<void> | void,
+    onShippingAddressChange? : ?(data : OnShippingAddressChangeData, {| resolve : () => ZalgoPromise<void>, reject : (string) => ZalgoPromise<void> |}) => ZalgoPromise<void> | void,
+    onShippingOptionsChange? : ?(data : OnShippingOptionsChangeData, {| resolve : () => ZalgoPromise<void>, reject : (string) => ZalgoPromise<void> |}) => ZalgoPromise<void> | void,
     onError : (mixed) => ZalgoPromise<void> | void,
     onClose : () => ZalgoPromise<void> | void,
     fundingSource : FundingType,
-    card : ?$Values<typeof CARD>,
+    card? : ?$Values<typeof CARD>,
```
AnshuChaudhari, line 60: We added `?` and made a few keys optional for now here because we were getting flow errors like in the screenshot below, for all the keys which are changed. 
As in `payment-fields` file Checkout component is being used for the pop-up, the CheckoutProps are being used too.
Please let us know if you have any other suggestion for this to not changing the existing CheckoutProps and also fix the flow errors. 
image

server/lib/get-experiments.js
```diff
@@ -20,13 +20,18 @@ export type GetExperimentsParams = {|
...
+    isInlinePaymentFieldsEligibleExperiment: InlinePaymentFieldsEligibility
 |};
 
 export function getDefaultExperiments() : Promise<GetExperimentsType> {
     return Promise.resolve({
         enableVenmoAppLabel:           false,
         isFundingSourceBranded:        false,
-        isCardFieldsExperimentEnabled: false
+        isCardFieldsExperimentEnabled: false,
+        isInlinePaymentFieldsEligibleExperiment: {
```
mnicpt, line 32: @AnshuChaudhari I think we want to handle these eligibility requirements in checkout-components and just have boolean for the value in SPB.

src/button/button.js
```diff
@@ -31,7 +31,8 @@ type ButtonOpts = {|
     wallet : ?Wallet,
     buyerAccessToken : ?string,
     eligibility : {|
-        cardFields : boolean
+        cardFields : boolean,
+        inlinePaymentFields: InlinePaymentFieldsEligibility
```
mnicpt, line 35: @AnshuChaudhari See above.

src/types.js
```diff
@@ -40,40 +41,40 @@ export type ZoidComponent<P> = {|
 export type CheckoutProps = {|
     window? : ?(ProxyWindow | CrossDomainWindowType),
     sessionID : string,
-    buttonSessionID : string,
-    stickinessID : string,
+    buttonSessionID? : string,
```
mnicpt, line 44: Keep existing.

src/types.js
```diff
@@ -40,40 +41,40 @@ export type ZoidComponent<P> = {|
...
+    stickinessID? : string,
```
mnicpt, line 45: Keep existing

weihou (approved):

src/button/button.js
AnshuChaudhari: Later we agreed to go with the Elmo experiment check for inline eligibility itself.

server/lib/get-experiments.js
AnshuChaudhari: Later we agreed to go with the Elmo experiment check for inline eligibility itself.
@mnicpt Does this look good?

PR #406, opened 2022-05-19
Auth button

adolfo-flowers:
### Description

New implementation of the Log in with Paypal button. To have the Log in with Paypal button render server side. The log in with paypal button zoid main component is at https://github.com/paypal/paypal-identity-components the code here is what will be loaded inside the iframe.

https://engineering.paypalcorp.com/jira/browse/DTPIXEL-1290

https://engineering.paypalcorp.com/jira/browse/DTLOCALZN-14065
Some bug fixes on paypal-identity-components auth button

❤️  Thank you!


server/components/auth-button/htmlTemplate.js
```diff
@@ -0,0 +1,69 @@
+/* @flow */
+/** @jsx node */
+import { html } from 'jsx-pragmatic';
+import { AuthButton } from '@paypal/identity-components/dist/button';
```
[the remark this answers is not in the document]
adolfo-flowers: Hi @gregjopa thanks for pointing that out. I considered the watcher approach but I had a couple of doubts so I decided to do it through the dependency since this is just a very simple button component which won't see a lot of updates. Plus I think there might also exist the possibility of it falling out of sync the watcher way. 

If I understand correctly due to the caching done through memoization there are some scenarios where this might happen. How do we handle the case of high availability? For example with multiple servers running behind a load balancer then we could have a situation where different calls to the same server will respond with different versions of the same component in the scenario where for example a server gets respawned so it downloads the js bundle again while the others served the cached version. Also how do we handle breaking changes to the components api this way?

I think there is another way of always installing the latest version of the module on server start maybe incorporating a node upgrade command for PayPal packages at server boot which would have exactly the same effect with less complexity to the code base. That way we would be able to have the code more standarized while keeping basically the same behavior as the getWatcher approach. With the benefit of no code changes when switching between update strategies.

Although if it is strictly necessary then I would gladly add the getWatcher if its optional then I would prefer to leave it like that.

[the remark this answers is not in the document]
adolfo-flowers: @gregjopa I implemented the get watcher but I pass an empty sdkLocationInformation object since this function allways returns an invalid object https://github.com/paypal/paypal-smart-payment-buttons/blob/8a0b36bd87de658a0ce7f518ff98bde9f324f966/server/devServer.js#L240 also instead of ACTIVE_TAG I use latest since I don't know whats the deal with the environment variable. Thanks.

server/components/auth-button/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+                responseType,
+                clientID,
+                returnurl,
+                customLabel,
+                locale,
+                style = {}
+            } = params;
+
+            if (!clientID) {
+                logger.info(req, 'smart_buttons_render');
```
wsbrunson, line 32: we either want to
1. remove this line completely
2. move this line above the `if` statement and change the event name to `auth_buttons_render`

probably go with 1. unless your team needs this log for something

server/components/auth-button/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+                return clientErrorResponse(res, 'Please provide a clientID query parameter');
+            }
+            logger.info(req, `auth_button clientID: ${clientID}`);
```
wsbrunson, line 35: I don't think we should put `clientID` in the log name because the cardinality on that would be extremely high. I would suggest:

```suggestion
            logger.info(req, 'auth_button', {clientID});
```

another suggestion would be to add this to the rootTxn like we do for the payment buttons: https://github.com/paypal/paypal-smart-payment-buttons/blob/main/server/components/buttons/instrumentation.js#L37

server/components/auth-button/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+            const script = await getPayPalAuthButtonsRenderScript({
+                logBuffer,
+                cache,
+                locationInformation,
+                sdkLocationInformation: {
+                    sdkCDNRegistry: undefined,
+                    sdkActiveTag: undefined
+                },
+            });
+            logger.info(req, `auth_button script version ${script.version}`);
```
wsbrunson, line 45:
```suggestion
            logger.info(req, `auth_button_version_${script.version}`);
```

server/watchers.js
```diff
@@ -49,6 +46,51 @@ function logInfo(logBuffer : LoggerBufferType, name : string, moduleDetails : Mo
     logBuffer.info(`${ name }_version_${ version.replace(/[^0-9]+/g, '_') }`, {});
 }
 
+export function getPayPalIdentityComponentsWatcher({ logBuffer, cache, locationInformation, sdkLocationInformation = {} } : {| logBuffer : ?LoggerBufferType, cache : ?CacheType, locationInformation : InstanceLocationInformation, sdkLocationInformation? : SDKLocationInformation |}) : Watcher {
```
wsbrunson, line 49: ah this is tricky. we just changed the way this works to support a phased rollout of the js-sdk through remote config. Give me til the end of the day and I can update this PR to use the new stuff we added if thats cool?

wsbrunson (approved):
once we get this tested in a TestEnv we can merge and publish

PR #409, opened 2022-05-25
Inline XO: onShippingChange callbacks

mnicpt:
### Description
Adds two new shipping change callbacks, `onShippingAddressChange` and `onShippingOptionsChange`.

For backward compatibility, the applications will check to see if merchant has implemented the new callback.  If present it will be called, otherwise, the current `onShippingChange` is called.


src/payment-flows/checkout.js
```diff
@@ -267,6 +267,24 @@ function initCheckout({ props, components, serviceData, payment, config, restart
...
+                    }
+                    
+                    return onShippingAddressChange({ buyerAccessToken, ...data }, actions);
+                } : null,
+
+            onShippingOptionsChange: onShippingOptionsChange
+                ? (data, actions) => {
+                    if (!data.selected_shipping_option) {
+                        throw new Error('Must pass selected_shipping_option in data to handle changes in shipping options.');
+                    } else {
```
wsbrunson, line 283: super minor, should we drop the `else` to match what `onShippingAddressChange` looks like?

src/props/onShippingAddressChange.js
```diff
@@ -0,0 +1,172 @@
...
+}
+
+export function buildXOnShippingAddressChangeActions({ data, actions: passedActions, orderID, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI } : {| data : OnShippingAddressChangeData, actions : OnShippingAddressChangeActionsType, orderID : string, facilitatorAccessToken : string, buyerAccessToken : ?string, partnerAttributionID : ?string, forceRestAPI : boolean |}) : XOnShippingAddressChangeActionsType {
+    const patchQueries = {};
+
+    let newAmount;
+    let breakdown = data.amount?.breakdown || {};
+
+    if (Object.keys(breakdown).length === 0) {
+        throw new Error('Must pass amount with breakdown into data attribute for onShippingAddressChange callback.');
```
wsbrunson, line 79: Something I've always wondered: are these error messages for merchants/merchant developers? Or would this be for us?

src/props/utils.js
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    return newAmount.toFixed(2);
+};
+
+export const buildBreakdown = ({ breakdown = {}, updatedAmounts = {} } : {| breakdown : Breakdown, updatedAmounts : {| [string] : string |} |}) : Breakdown => {
+    const breakdownKeys = Object.keys(breakdown);
+    const updatedAmountKeys = Object.keys(updatedAmounts);
+
+    const currency_code = breakdown[breakdownKeys[0]]?.currency_code;
```
wsbrunson, line 30: would it be more readable if we did `Object.values(breakdown)[0]?.currency_code`? just because I don't see `breakdownKeys` used again

src/props/utils.js
```diff
@@ -0,0 +1,56 @@
...
+export const convertQueriesToArray = ({ queries } : {| queries : {| [$Values<typeof ON_SHIPPING_CHANGE_PATHS>] : Query |} |}) : $ReadOnlyArray<Query> => {
+    const convertedQueries = [];
+    const patchQueryKeys = Object.keys(queries);
+    if (patchQueryKeys) {
+        patchQueryKeys.forEach(key => {
+            convertedQueries.push(queries[key]);
+        });
+    }
+
+    return convertedQueries;
```
wsbrunson, line 55: could this be
```
return Object.values(queries) || [] 
```

wsbrunson (approved):
minor comments, nothing to block the merge of this PR

nbierdeman (approved):
Nice work, @mnicpt! 🤩

src/props/onShippingAddressChange.js
mnicpt: @wsbrunson This one is an error for apps using the callback; like Dumbledore.

src/props/utils.js
mnicpt: Had an issue with typings by changing it as Object.values are of mixed type.

mnicpt:
Thanks for the review @nbierdeman and @wsbrunson and the team for the live review!  Also, thanks for all of the feedback that went into this.

wsbrunson (approved):

upigilam:
>For backward compatibility, the applications will check to see if merchant has implemented the new callback. If present it will be called, otherwise, the current onShippingChange is called.

NOTE: This change breaks for the new merchants(assuming old callback `onShippingChange` wont be present) falling back to existing member and guest applications for any unhandled errors in the new inlinexo app until we actually implement new callbacks in fallback apps.

mnicpt:
retest this please

PR #410, opened 2022-06-02
feat(server): use xprops.onError callback for critical errors

borodovisin:
### Description
The PR contains the changes to notify the `window.xprops.onError` when the server response fails in order to notify the JS SDK on the source client site.

The idea is to catch any error on the server and return an HTML page containing the JS SDK to enable communication between the requested resources and the parent window through the `xprops.onError` mechanism from Zoid.

Please refer to this [ticket](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-725)

### Screenshots
Before we aren't able to notify the error to the parent window.
previous_behaviour

With this change we can see the alert generated from the `onError` callback in the parent window
new_behaviour

❤️  Thank you!


server/lib/util.js
```diff
@@ -264,3 +267,32 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+* Return a HTML including the JS SDK allowing the communication with the parent SDK.
+*
+* @param {ErrorArgumentNonMaybeValue} param0 the needed information to track and handle the errors
+* @returns an HTML page with access to the parent window.xprops
+*/
+export function htmlErrorHandler({ res, err, meta } : ErrorArgument) : void {
+    const cspNonce = getCSPNonce(res);
+    const pageHTML = `<!DOCTYPE html>
+        <html>
+        <head>
```
gregjopa, line 282: Do we need to include a `<title>`? I'd prefer to remove it and keep this payload as small as possible. Ex:

```html
<!DOCTYPE html>
<html>
    <body>
    <!-- Add logic here -->
    </body>
</html>

```

server/lib/util.js
```diff
@@ -264,3 +267,29 @@ export function getCSPNonce(res : ExpressResponse) : string {
 
     return nonce;
 }
+
+/**
+* Handler for server error and communication with the parent window.
+* Return a HTML including the JS SDK allowing the communication with the parent SDK.
+*
+* @param {ErrorArgumentNonMaybeValue} param0 the needed information to track and handle the errors
```
gregjopa, line 275: I would prefer to remove the `@param` and `@returns` part of the comment. They seem outdated and we have the flow types to self describe the inputs and outputs.

server/lib/util.js
```diff
@@ -264,3 +267,26 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+                ${ meta.getSDKLoader({ nonce: cspNonce }) }
+                <script nonce="${ cspNonce }">
+                    if (window.xprops && typeof window.xprops.onError === "function") {
+                        window.xprops.onError(${ errorMessage || 'Unexpected error' });
+                    }
+                </script>
+            </body>
+        </html>`;
+
+    return htmlResponse(res, pageHTML);
```
gregjopa, line 289: Are we able to send back a 500 error with htmlResponse()?

server/lib/util.js
```diff
@@ -264,3 +267,26 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+*/
+export function htmlErrorHandler({ res, errorMessage, meta } : ErrorArgument) : void {
+    const cspNonce = getCSPNonce(res);
+    const pageHTML = `<!DOCTYPE html>
+        <html>
+            <body data-nonce="${ cspNonce }">
+                ${ meta.getSDKLoader({ nonce: cspNonce }) }
+                <script nonce="${ cspNonce }">
+                    if (window.xprops && typeof window.xprops.onError === "function") {
+                        window.xprops.onError(${ errorMessage || 'Unexpected error' });
```
gregjopa, line 283:
```suggestion
                        window.xprops.onError("${ errorMessage }");
```

server/lib/util.js
```diff
@@ -264,3 +267,26 @@ export function getCSPNonce(res : ExpressResponse) : string {
 
     return nonce;
 }
+
+/**
+* Handler for server error and communication with the parent window.
+* Return a HTML including the JS SDK allowing the communication with the parent SDK.
+*/
+export function htmlErrorHandler({ res, errorMessage, meta } : ErrorArgument) : void {
```
gregjopa, line 275:
```suggestion
export function htmlErrorHandler({ res, errorMessage = 'Unexpected error', meta } : ErrorArgument) : void {
```

test/server/__snapshots__/render.button.test.js.snap
```diff
@@ -0,0 +1,72 @@
...
+                            loadV4();
+                            loadV5();
+                        } catch (err) {
+                            return;
+                        }
+                    })();
+                </script>
+                <script nonce=\\"\\">
+                    if (window.xprops && typeof window.xprops.onError === \\"function\\") {
+                        window.xprops.onError(Mocking a critical failure);
```
gregjopa, line 67: @borodovisin this error message needs to be wrapped in quotes

server/lib/util.js
```diff
@@ -264,3 +267,28 @@ export function getCSPNonce(res : ExpressResponse) : string {
...
+* Return a HTML including the JS SDK allowing the communication with the parent SDK.
+*/
+export function htmlErrorHandler({ res, meta, errorMessage } : ErrorArgument) {
+    const cspNonce = getCSPNonce(res);
+    const pageHtml = `<!DOCTYPE html>
+        <html>
+            <body data-nonce="${ cspNonce }">
+                ${ meta.getSDKLoader({ nonce: cspNonce }) }
+                <script nonce="${ cspNonce }">
+                    const error = new Error("${ errorMessage }");
```
gregjopa, line 282: Let's use `var` here to be safe. This code does not go through a transpiler so let's make sure it only uses ES3 features.
```suggestion
                    var error = new Error("${ errorMessage }");
```

gregjopa (approved):
Great improvement! 💯

westeezy (approved):

PR #415, opened 2022-06-08
feat: replace grabthar with button version manager

wsbrunson:
Bringing the phased rollout changes we did for the js sdk to spb.

Before we merge this, we'll need to:
- [x] create a remoteconfig for smartcomponentnodeweb 
- [x] update smartcomponentnodeweb to create and pass through the `spbVersionManager`


server/components/buttons/middleware.js
```diff
@@ -128,7 +123,7 @@ export function getButtonMiddleware({
             }
 
             const renderButton = await renderPromise;
-            const client = await clientPromise;
+            const clientScript = await clientPromise;
```
westeezy, line 126: nice rename here

PR #417, opened 2022-06-13
[no merge] fix applepay user gesture handler issue safari 15.5

andyfen:
as of safari 15.5 applepay is now failing with 
```
Must create a new ApplePaySession from a user gesture handler
```
can be seen [here](https://live-applepay-paypal-js-sdk.herokuapp.com/async-order-create.html)

Specified by apple:
_You attempt to create the ApplePaySession outside of a user gesture handler. The exception error "Must create a new ApplePaySession from a user gesture handler" appears._

https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api/creating_an_apple_pay_session

existing flow of events 
```
click ->
    createOrder
        -> orderId
         getCheckoutsession
                -> data
                 new ApplepaySession(data) 
```
this is now occurring because `createOrder` and `getCheckoutSession` is between the `click` and `ApplepaySession` events

`ApplepaySession` has to be directly called by `click` and this need to propagate the [isTrusted](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) event property. 

a change in safari has occurred and this is not propagating to the ApplepaySession as it now sees `createOrder` as the initializer which is not a user event.

given checkout_session data is required to populate the ApplepaySession we need to initially create and order on render to obtain the required info. 

this restructures the flow of events to 
```
createOrder
    -> orderId
    getCheckoutsession
        -> data

        click
            new ApplepaySession(data)
```


mnicpt (changes requested):
@andyfen This issue would popup typically testing because of latency.  Not because of events occurring after click.  Given that, it could be that Apple has lessened that amount of time.  If they have this will also impact popups for Member flows.  Does this happen in production?

src/payment-flows/applepay/applepay.js
```diff
@@ -17,11 +17,41 @@ const SUPPORTED_VERSION = 4;
...
+        -> orderId
+        getCheckoutsession()
+            -> data
+
+            click ->
+                new ApplepaySession(data)
+    */
+    const { createOrder, locale, onError } = props;
+
+    return createOrder().then(id => {
```
mnicpt, line 34: @andyfen We aren't allowed to create an order without the user initiating by clicking on the button.   This solution will not work like this.

PR #420, opened 2022-06-15
Fix for approving an order during the vault capture flow

sioked:
### Description
Currently, when using the vault-capture flow, after we apply the payment (by calling the `validate-payment-method` endpoint), we have not actually approved the order. For some of our merchants, they validate that the order is approved before they will attempt to capture the order.

This PR adds in an additional call to the Orders API to `confirm-payment-method` that will set the order to `APPROVED` based on the user action

PPPLMER-104024

* Visit https://www.te-cibns.qa.paypal.com/xo-integrations/vaulting/client-token
* Set a breakpoint in the `onApprove` method prior to calling capture on the order
* Populate the customer_id of `test-1654613061212`
* Upon Setting the customer ID and waiting for the buttons to re-render, you should see the one-click buttons
* Click the one-click button
* When the breakpoint is hit, grab the order id that was created and look it up via the orders api
* You should see that the result is that the order has a status of `CREATED`. It should be `APPROVED`

```bash
curl --request GET \
  --url https://msmaster.qa.paypal.com:12326/v2/checkout/orders/2X1050494V836180N \
  --header 'Authorization: Bearer A21_A.AAcEhfq4h5VrKjkAZ_tbG7S_XD50-ncSKCkri27He_e2tLuJVgLJf74cduHbNhAPpVTQy-5A_n50dmAf-7T4XIr0l4ROGg' \
  --header 'Content-Type: application/json' \
```

❤️  Thank you!

### DIFF
--- a/src/payment-flows/vault-capture.js
+++ b/src/payment-flows/vault-capture.js
@@ -8,7 +8,7 @@ import { initiateInstallments } from '@paypal/installments/src/interface';
 
 import type { ThreeDomainSecureFlowType, MenuChoices } from '../types';
 import type { CreateOrder } from '../props';
-import { validatePaymentMethod, type ValidatePaymentMethodResponse, getSupplementalOrderInfo, deleteVault, updateButtonClientConfig, loadFraudnet } from '../api';
+import { validatePaymentMethod, type ValidatePaymentMethodResponse, getSupplementalOrderInfo, deleteVault, updateButtonClientConfig, loadFraudnet, confirmOrderAPI, buildPaymentSource } from '../api';
 import { TARGET_ELEMENT, BUYER_INTENT, FPTI_TRANSITION, FPTI_CONTEXT_TYPE, FPTI_MENU_OPTION } from '../constants';
 import { getLogger } from '../lib';
 import type { ButtonProps } from '../button/props';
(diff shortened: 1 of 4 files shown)


src/payment-flows/vault-capture.js
```diff
@@ -174,7 +174,10 @@ function initVaultCapture({ props, components, payment, serviceData, config } :
 
             const { status, body } = validate;
             return handleValidateResponse({ ThreeDomainSecure, status, body, createOrder, getParent }).then(() => {
-                return onApprove({}, { restart });
+                confirmOrderAPI(orderID, { payment_source: buildPaymentSource(paymentMethodID) }, { facilitatorAccessToken, partnerAttributionID })
```
wsbrunson, line 177: do you need a `return` here too?

wsbrunson (approved):

src/payment-flows/vault-capture.js
sioked: oh yeah, good catch, just pushed up that fix

wsbrunson (approved):

wsbrunson:
oh one last question, do we need to do anything special if `confirmOrderApi` fails? can we still call `onApprove` or should we stop processing?

sioked:
> oh one last question, do we need to do anything special if confirmOrderApi fails? can we still call onApprove or should we stop processing?

My expectation is that if the confirmOrderApi call fails we should throw that error as part of the promise chain. Currently, the `confirmOrderApi` call uses the `callRestAPI` method at `src/api/api.js` and that has a default handler that will check for any non-200 request and throw that error up the chain for us. Slightly different than the `validate` request up above where we have custom logic around this. It also does some default logging.

PR #426, opened 2022-06-22
Update onComplete data and actions

mnicpt:
### Description
Adds `capture` and `get` to actions for `onComplete` callback and also `intent` to the data attribute.  This allows merchants to capture the order if `intent=authorize` and authorization is done in the app.

### DIFF
--- a/src/button/pay.js
+++ b/src/button/pay.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { noop, stringifyError } from '@krakenjs/belter/src';
+import { noop, stringifyError, isCrossSiteTrackingEnabled } from '@krakenjs/belter/src';
 import { EXPERIENCE } from '@paypal/checkout-components/src/constants/button';
 import { ZalgoPromise } from '@krakenjs/zalgo-promise/src';
 import { FPTI_KEY } from '@paypal/sdk-constants/src';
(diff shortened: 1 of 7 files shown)


src/button/pay.js
```diff
@@ -115,8 +108,14 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 [FPTI_KEY.PAYMENT_FLOW]:      name,
                 [FPTI_KEY.IS_VAULT]:          instrumentType ? '1' : '0',
                 [FPTI_CUSTOM_KEY.INFO_MSG]:   enableNativeCheckout ? 'tester' : '',
-                [FPTI_CUSTOM_KEY.EXPERIENCE]: derivedExperience
-            }).flush();
+                [FPTI_CUSTOM_KEY.EXPERIENCE]: experience === EXPERIENCE.INLINE ? 'inline' : ''
+            });
```
sunnypatel, line 112: Not sure what this does but do we need to `.flush()` here too?

mnicpt: @sunnypatel No.  The next log is taking care of it.

src/payment-flows/card-form.js
```diff
@@ -18,18 +19,28 @@ function setupCardForm() {
...
+
+        return inlinexoExperiment.isEnabled() ? false : true;
     }
 
     if (vault) {
         return false;
     }
 
-    if (onShippingChange || onShippingAddressChange || onShippingOptionsChange) {
+    if (onShippingChange) {
```
ravishekhar, line 43: Just a query - do we need to support `isCardFormEligible` when `onShippingAddressChange` or `onShippingOptionsChange` is present but do not support when `onShippingChange` is present?

ravishekhar (approved):
lgtm 👍

PR #432, opened 2022-06-30
Fix Flow error

mnicpt:

❤️  Thank you!


src/props/onComplete.js
```diff
@@ -35,7 +35,7 @@ export type XonCompleteOrderActions = {|
 |};
 
 export type XonCompletePaymentActions = {|
-    capture : () => ZalgoPromise<AuthorizationCaptureResponse>
+    capture : (data : AuthorizationCaptureData) => ZalgoPromise<AuthorizationCaptureResponse>
```
sunnypatel, line 38: nit: Can we rename this to `authCaptureData` ?

sunnypatel (approved):
left a nit comment but otherwise lgtm!

PR #442, opened 2022-07-14
PostalCode Field for Multi Card Fields

jplukarski:
### Description

An optional component for the postal code input for Multi Card Fields

We want to ensure we have parity with the current PayPal Hosted Fields offerings. Right now, Hosted Fields has an optional Postal Code field for merchants that need to collect this information for Fraud Checks

[Jira - DTNOR-528](https://engineering.paypalcorp.com/jira/browse/DTNOR-528)

178842615-b2a8aa1b-ac1d-46ee-a14e-dc9c6142dc9f

- https://github.com/paypal/paypal-checkout-components/pull/1967

### Co-Authored By:
-@bywood
-@jplukarski 

❤️  Thank you!


nbierdeman (commented):
Looking good! 😎 Here are some screenshots with questions about input formatting and validation. Please let me know what you think. 🙏 

Screen Shot 2022-07-19 at 2 06 51 PM

Screen Shot 2022-07-19 at 1 51 25 PM

Screen Shot 2022-07-19 at 1 57 41 PM

Screen Shot 2022-07-19 at 1 49 31 PM
.

jplukarski:
> Looking good! 😎 Here are some screenshots with questions about input formatting and validation. Please let me know what you think. 🙏

Thanks for the feedback!
According to the design we were provided `Postal code`, `Card number` , `Cardholder name` should have the second word uncapitalized. We'll get that updated.

For card number, `111` is invalid since it is not a valid card number, but there is a `isPotentiallyValid` property that is set and in this case that would be true since it could be an incomplete number. Postal code is valid in the case of `122'` because we only check if the postal code is a string fewer than 3 characters in length (we use the [braintree card-validator](https://github.com/braintree/card-validator#validpostalcodevalue-string-options-object-object) module for this). The idea behind this is that we want to avoid making as many assumptions as possible on the postal code. We give the merchant control over how many characters the postal code is by allowing them to pass in a `minLength` and a `maxLength` option when rendering this field. The same thinking goes for if a customer enters a hyphen. Merchants that service locations with postal codes with hyphens can extend the `maxLength` to accommodate these postal codes

nbierdeman:
> Postal code is valid in the case of 122' because we only check if the postal code is a string fewer than 3 characters in length (we use the [braintree card-validator](https://github.com/braintree/card-validator#validpostalcodevalue-string-options-object-object) module for this).

@jplukarski, are non-alphanumeric characters ever valid input for postal codes? Does the BT card validator throw errors for invalid characters? I'm just trying to understand if it's possible for merchants to capture bogus postal codes, since they can be used for fraud detection.

jplukarski:
> are non-alphanumeric characters ever valid input for postal codes (I guess other than hyphens)? Does the BT card validator throw errors for invalid characters? I'm just trying to understand if it's possible for merchants to capture bogus postal codes, since they can be used for fraud detection.

So the way that it works on the Braintree side is the card-validator will only throw errors based on the length. When the postal code is sent to the Braintree Gateway, [we will throw a validation error on the merchant's server side.](https://developer.paypal.com/braintree/docs/reference/general/validation-errors/all/ruby#code-81813).

Our thinking was that since Hosted Card Fields uses Braintree Hosted Fields under the hood, we wanted to keep the client side check the same so that the new Multi-Card Fields will have as much parity with the current Hosted Card Fields product as possible. I am checking to see if there are similar errors returned from the orders API for postal code. If so, do we think it makes sense to keep parity, or should we update the client side check to only allow spaces, hyphens and alphanumeric characters?

nbierdeman:
>...do we think it makes sense to keep parity, or should we update the client side check to only allow spaces, hyphens and alphanumeric characters?

@crookedneighbor, any thoughts on adding additional client side validation for postal code above and beyond what BT does (see convo above for more context)? My thought was to catch as much invalid input as possible on the client side, so as to avoid any unnecessary roundtrips just to throw an error.

crookedneighbor:
In the Braintree SDK, since we version the SDK, if the validation rules need to be updated in the future, it means all the old versions of the SDK would have the incorrect rules. Since the PayPal SDK is evergreen, that aspect is less of an issue, obviously. IE, if the max length of a postal code changes in some country, y'all _should_ be able to make an update to accommodate that.

I will note that in the BT SDK, we have very limited validation on the client side to be as inclusive of the postal code rules for international customers. IE, postal codes in the US are a set length and only numbers, but postal codes in Iceland can be as little as 3 characters and many places the code is alphanumeric.

On the BT side, we chose to limit the validation we did on the client side and let the server be responsible for catching issues in the billing address rather than loading up the client with extra validation checks. Especially since we don't have that information about the context of the postal code easily available to us in the BT SDK (we typically just have the card details + postal code, not other parts of the billing address). If I recall, I believe the PayPal SDK has some detection around the customer's country, but I'd be wary of trusting those automated checks and conditionally applying more rules based on that detection. Too many edge cases where you'd prevent a customer from inputting the correct postal code for their billing location.

So anyway, that's the context. Do what y'all think is best. As long as there is an easy way to keep validations in sync with the server, and we don't accidentally prevent a customer from entering a valid postal code, I don't see why not.

nbierdeman:
Thanks, @crookedneighbor! That's some really good context.

@jplukarski, I'm thinking whether to add more client side validation or not is up to you and your team. I just wanted to point out that I'm able to enter random characters at the moment (consider this non-blocking feedback).

jplukarski:
@nbierdeman 

We are going to move forward with minimal validation for now, but we'll keep in mind additional validation for future iterations. Thanks again for your feedback here, definitely appreciate it!

PR #447, opened 2022-07-26
chore: set fraudnet id to order id for applepay

jscheinhorn:
### Description

Merchant will be directed to include paypal-client-metadata-id (CMID) in the header of their capture call. Because Applepay flow doesn't follow the same process as when the PayPal pop-up is opened (and instead uses its own sheets), the CMID is never set to the Order ID, making it difficult for Fraudnet to track the Applepay order when it comes time to do order capture.
This fix is to set the Fraudnet ID to the Order ID when the buyer has clicked the Applepay button and the Order ID is created.

Updated PR: Removes initial fraudnet script element and replaces it with new element with Order ID. This guarantees the P1 and P2 network calls made after button click use the Order ID as the correlationID. 

[DTALTPAY-986](https://engineering.paypalcorp.com/jira/browse/DTALTPAY-986)
DTAPMWALL-321

Visit https://stage-applepay-paypal-js-sdk.herokuapp.com/advanced.html
Click Applepay button
Check Network tab for the Fraudnet call (P1, P2...)
Check the CorrelationID value to see the Fraudnet ID is set to the Order ID.

Applepay Order Created - Fraudnet
Applepay P1 - Fraudnet
Applepay P2 - Fraudnet
-----
Before:
Two elements with fconfig id
After:
Applepay Fraudnet Updated 4
Applepay Fraudnet Updated 3
Applepay Fraudnet Updated 2
Applepay Fraudnet Updated 1

❤️  Thank you!

### DIFF
--- a/src/api/fraudnet.js
+++ b/src/api/fraudnet.js
@@ -64,6 +64,12 @@ export const loadFraudnet : Memoized<LoadFraudnet> = memoize(({ env, clientMetad
         window.fnCallback = resolve;
         setTimeout(resolve, timeout);
 
+        // replace existing script to update cmid
+        const fconfigElement = document.getElementById('fconfig');
+        if (fconfigElement) {
+            fconfigElement.remove();
+        }
+
         const body = getBody();
         body.appendChild(configScript);
         body.appendChild(fraudnetScript);
--- a/src/payment-flows/applepay/applepay.js
+++ b/src/payment-flows/applepay/applepay.js
@@ -348,7 +348,7 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
                         logApplePayEvent('validatemerchant', { validationURL });
 
                         orderPromise.then(orderID => {   
-                            loadFraudnet({ env, clientMetadataID: orderID }).catch(noop);        
+                            loadFraudnet({ env, clientMetadataID: orderID }).catch(noop);
                             getDetailedOrderInfo(orderID, locale.country).then(order => {
                                 const {
                                     merchant,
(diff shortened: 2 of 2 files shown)


wsbrunson (approved):

gregjopa:
@jscheinhorn please rebase this PR with `main` to fix the conflicts.

src/api/fraudnet.js
```diff
@@ -64,6 +64,12 @@ export const loadFraudnet : Memoized<LoadFraudnet> = memoize(({ env, clientMetad
         window.fnCallback = resolve;
         setTimeout(resolve, timeout);
 
+        // replace existing script to update cmid
+        const fconfigElement = document.getElementById('fconfig');
+        if (fconfigElement) {
+            fconfigElement.remove();
+        }
```
andyfen, line 71: this is removing it because its getting called before, do we know where this is getting called before ?

jscheinhorn: @andyfen that initial fraudnet call is from https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/payment-flows/wallet-capture.js#L50

andyfen: i don't think that should be getting called for applepay ? it's a different flow, might need to investigate whats going on

jscheinhorn: The smartWalletPromise is getting called. 
It looks like the eligibility is true for all payment-flows (when testing on local) so flow.setup() is called. The only flow that calls loadFraudent in its setup function is the walletCapture flow, so that's where it's coming from. 
https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/button/pay.js#L30-L33

wsbrunson (approved):

PR #452, opened 2022-08-05
applepay - oncancel 

andyfen:
oncancel

depends on 
https://github.com/paypal/paypal-checkout-components/pull/1984


wsbrunson (approved):

andyfen:
it looks like this failing test is because of the previous commit that was force merged

mnicpt (approved):

mnicpt:
@andyfen Yeah, that's odd. I removed a log and then the native test started failing.  We're good then.

PR #459, opened 2022-08-11
applepay handle onclick on reject do abort

andyfen:
calls `abort` on applepay session when `actions.reject()` is called and closes the payment sheet 
```
onClick(data, actions){
  return actions.reject()
}
```

example use case: pre payment validation 

adobe-onclick

stage test https://stage-applepay-paypal-js-sdk.herokuapp.com/onclick.html


wsbrunson (approved):

src/payment-flows/applepay/applepay.js
```diff
@@ -347,6 +324,33 @@ function initApplePay({ props, payment, serviceData } : InitOptions) : PaymentFl
                     function validateMerchant({ validationURL } : {| validationURL : string |}) {
                         logApplePayEvent('validatemerchant', { validationURL });
 
+                        const validatePromise = ZalgoPromise.try(() => {
```
mnicpt, line 327: @andyfen Wondering if we can't leave the `validate` above where its at and just do the promise inside `validateMerchant`. Will keep the promise more concise to just resolve and not do both in one.

mnicpt (approved):
@andyfen Left one comment.  Great work!

src/payment-flows/applepay/applepay.js
andyfen: i tried this initially, the issue is that `abort` needs to be in scope inside validatePromise.

here
https://github.com/paypal/paypal-smart-payment-buttons/blob/b1565ae7bb0ea2963e0743aca1ae0d85e2d82f77/src/payment-flows/applepay/applepay.js#L331

PR #460, opened 2022-08-12
remove all server code

wsbrunson:
There's a PR in our private Github to move all this code to the node service that runs smart-payment-buttons


ravishekhar (approved):
👍

PR #476, opened 2022-09-14
Remove whitelist of clientIDs

mnicpt:
### Description
https://engineering.paypalcorp.com/jira/browse/SVM-3789


wsbrunson (approved):
simple enough!

sioked:
Looks like there's a broken test that is likely related to this change. Previously we'd do payee validation on click after order creation, this would end up throwing the error in the console but allowing the order to continue.

elizabethmv (approved):

PR #478, opened 2022-09-19
Add Focus and Clear Methods to Multi Card Fields

jplukarski:
### Description

Adds clear and focus methods to Multi Card Fields. Additionally, this PR contains a refactor for the isEligible method to default to ineligible, and to report the merchant as not eligible if they are not onboarded onto unbranded payments. We also included a fix to a bug in the expiry date field where the text would not delete past the slash when the user his backspace. To solve this we refactored the expiry date format implemented the [restricted input module](https://github.com/braintree/restricted-input) used in the Braintree SDK.

[DTNOR-486](https://engineering.paypalcorp.com/jira/browse/DTNOR-486)
[DTNOR-577](https://engineering.paypalcorp.com/jira/browse/DTNOR-577)
[DTNOR-485](https://engineering.paypalcorp.com/jira/browse/DTNOR-485)
[DTNOR-587](https://engineering.paypalcorp.com/jira/browse/DTNOR-587)

[PR in Checkout Components](https://github.com/paypal/paypal-checkout-components/pull/2007)

### Co-authored-by
@bywood 
@jplukarski 
@nidhinarendra

❤️  Thank you!


src/card/components/page.jsx
```diff
@@ -77,13 +78,15 @@ function Page({ cspNonce, props } : PageProps) : mixed {
     };
 
     useEffect(() => {
-        if(typeof onChange === 'function') {
+        if ( initialRender.current ) {
+            initialRender.current = false
```
bywood, line 82: This could probably use a comment. It's not real clear to me what this is doing. It looks like `initialRender.current` is initially set to `true`, but then won't that get set to the DOM element itself which would also be truthy?

src/card/lib/card-utils.js
```diff
@@ -56,45 +56,27 @@ export function maskValidCard(number : string) : string {
...
-
-    if (prevFormat && prevFormat.includes('/')) {
-        const [ month ] = removeSpaces(prevFormat).split('/');
-        if (month.length < 2) {
-            return prevFormat;
-        }
+export function shouldUseZeroPaddedExpiryPattern(value: string, key: string) : boolean {
+    let useZeroPaddedExpiryPattern
+
+    switch (true) {
```
bywood, line 62: Using `switch (true)` here seems a little odd. I'd recommend just using `if` statements. Something like:
```js
    if (value.length === 0) {
        return false;
    }
    if (value[0] === "1" && key === "/") {
        return true;
    }
    if (value[0] !== "1" && value[0] !== "0") {
        return true;
    }
    if (value[0] === "1") {
        return false;
    }
    return false;
```

src/card/components/page.jsx
jplukarski: Good call, comment added

src/card/lib/card-utils.js
jplukarski: I agree, your refactor suggestion is much more clear.

bywood (approved):

thebastedo (approved):

PR #479, opened 2022-09-20
Query Venmo in SmartWallet

elizabethmv:
### Description
This PR adds Venmo to the SmartWallet query 

One click Venmo button with JS SDK Vault

1. Run the JS SDK locally with this branch of SPB aliased
2. Fetch an ID token using the identity API
3. Pass the ID token as a `data-user-id-token` attribute on the script
4. In the network request search for the GetSmartWallet GraphQL call
5. Confirm Venmo was in the query

Pending GraphQL release. The GraphQL changes are wrapped in a feature flag for Venmo which is set to off.

N/A


src/api/wallet.js
```diff
@@ -87,6 +87,16 @@ export const getSmartWallet : GetSmartWallet = memoize(({ clientID, merchantID,
                             oneClick
                         }
                     }
+                    venmo {
+                        instruments {
+                            type
+                            label
+                            logoUrl
+                            instrumentID
+                            tokenID
```
nbierdeman, line 96: Is there no `vendor` key for Venmo?

nbierdeman (approved):
LGTM! 🥳 Just had one quick question below.

src/api/wallet.js
elizabethmv: It's not used in the Venmo flow, so I did not include it in the query

PR #482, opened 2022-09-23
Enhance button props tests

jshawl:
### Description

This PR provides 100% code coverage for `src/button/props.js` and removes a few dead code paths.

### Why are we making these changes? Include references to any related Jira tasks or 

Increase confidence in making changes to `src/button/props.js` in the future.

```
npm run jest:client
```

Before:
Screen Shot 2022-09-23 at 11 33 39

After:

Screen Shot 2022-09-23 at 11 33 00

### DIFF
--- a/src/button/props.js
+++ b/src/button/props.js
@@ -85,19 +85,11 @@ export function getButtonProps({ facilitatorAccessToken, brandedDefault, payment
         if (!xprops.createBillingAgreement) {
             throw new Error(`Must pass createBillingAgreement with intent=tokenize`);
         }
-
-        if (xprops.createOrder || xprops.createSubscription) {
-            throw new Error(`Must not pass createOrder or createSubscription with intent=tokenize`);
-        }
     }
 
     if (intent === INTENT.SUBSCRIPTION) {
         if (!xprops.createSubscription) {
             throw new Error(`Must pass createSubscription with intent=subscription`);
-        }
-
-        if (xprops.createOrder || xprops.createBillingAgreement) {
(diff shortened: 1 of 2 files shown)


src/button/props.js
```diff
@@ -85,20 +85,12 @@ export function getButtonProps({ facilitatorAccessToken, brandedDefault, payment
         if (!xprops.createBillingAgreement) {
             throw new Error(`Must pass createBillingAgreement with intent=tokenize`);
         }
-
-        if (xprops.createOrder || xprops.createSubscription) {
-            throw new Error(`Must not pass createOrder or createSubscription with intent=tokenize`);
-        }
```
jshawl, line 91: this line (90) is unreachable because of the previous conditions. 

e.g. if you pass `createOrder` the error would have thrown on line 62. Same with `createSubscription` and line 76.

src/button/props.js
```diff
@@ -85,20 +85,12 @@ export function getButtonProps({ facilitatorAccessToken, brandedDefault, payment
...
     }
 
     if (intent === INTENT.SUBSCRIPTION) {
         if (!xprops.createSubscription) {
             throw new Error(`Must pass createSubscription with intent=subscription`);
         }
-
-        if (xprops.createOrder || xprops.createBillingAgreement) {
-            throw new Error(`Must not pass createOrder or createBillingAgreement with intent=tokenize`);
-        }
```
jshawl, line 101: Same as above comment - also worth noting the error message reference `intent=tokenize` instead of `intent=subscription`

PR #484, opened 2022-09-27
Add src/lib/ test coverage

jshawl:
### Description

This PR adds additional test cases to get 100% test coverage for the `src/lib` directory. It looks like a lot of code but note that only test files are added and/or modified.

Increase confidence while changing `src/lib/` files.

```
npm run jest:client
```

Before:
Screen Shot 2022-09-27 at 12 53 37

After:
Screen Shot 2022-09-27 at 12 52 52

I spent an inordinate amount of time trying to cover the two branch conditions showing up in yellow but I'm not sure those paths are reachable - can always refactor in the future.


westeezy (approved):
🎉

PR #489, opened 2022-10-10
DTCHKINT-1039 removed data property

aimmadisetty:
### Description

The first capture API attempt to the regular merchant API endpoint then returns a 422 INSTRUMENT_DECLINED, as expected
And the format of the response is (err?.response?.body?.details?)
And details object contains INSTRUMENT_DECLINED.

The response path and the UI path was different, so removed data property to make it same

Screen Shot 2022-10-03 at 9 45 32 AM

❤️  Thank you!

### DIFF
--- a/src/api/order.js
+++ b/src/api/order.js
@@ -121,7 +121,7 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
 
 export function isProcessorDeclineError(err : mixed) : boolean {
     // $FlowFixMe
-    return Boolean(err?.response?.body?.data?.details?.some(detail => {
+    return Boolean(err?.response?.body?.details?.some(detail => {
         return detail.issue === ORDER_API_ERROR.INSTRUMENT_DECLINED || detail.issue === ORDER_API_ERROR.PAYER_ACTION_REQUIRED;
     }));
 }


wsbrunson:
@aimmadisetty do you have a link to any public docs that explain the shape of that response?

aimmadisetty:
> @aimmadisetty do you have a link to any public docs that explain the shape of that response?

{
    "name": "UNPROCESSABLE_ENTITY",
    "details": [
        {
            "issue": "INSTRUMENT_DECLINED",
            "description": "The instrument presented  was either declined by the processor or bank, or it can't be used for this payment."
        }
    ],
    "message": "The requested action could not be performed, semantically incorrect, or failed business validation.",
    "debug_id": "ce07a7cbd4982",
    "links": [
        {
            "href": "https://developer.paypal.com/docs/api/orders/v2/#error-INSTRUMENT_DECLINED",
            "rel": "information_link",
            "method": "GET"
        },
        {
            "href": "https://www.sandbox.paypal.com/checkoutnow?token=9X668241VU621235U",
            "rel": "redirect",
            "method": "GET"
        }
    ]
}
Screen Shot 2022-10-10 at 12 08 52 PM

wsbrunson:
do we document this response anywhere on developer.paypal.com?

wsbrunson:
also, looks like there is a failing test case

```
contingency cases
    ✗ should render a button, click the button, and render checkout, then pass onApprove callback to the parent with actions.order.capture and auto restart with INSTRUMENT_DECLINED
```

src/api/order.js
```diff
@@ -121,7 +121,7 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
 
 export function isProcessorDeclineError(err : mixed) : boolean {
     // $FlowFixMe
-    return Boolean(err?.response?.body?.data?.details?.some(detail => {
```
wsbrunson, line 124: can we change this to support both response types?

I looked more into `isProcessorDeclineError`, its used in three different requests. It's possible one of those will return `response.body.data.details`.

could we instead do
```
if (Boolean(err?.response?.body?.data?.details) {
  ...
} else {
    return Boolean(err?.response?.body?.details?.some(detail => {
       ...
    )};
```

wsbrunson (approved):

PR #490, opened 2022-10-18
Remove client-side client ids

wsbrunson:
We have a bug around exposing client-ids in client-running code. This is the client-side part to remove these client ids.

The main culprits are `ORDER_VALIDATION_WHITELIST` and `LSAT_UPGRADE_EXCLUDED_MERCHANTS`.

The idea is to create a feature flag "service" in our server code that would pass an object down to `setupButtons` and `setupCardFields` that will include whether the render can upgrade the lsat token or if we should throw integration errors

I think eventually we would move more feature flags into our feature flag object that we pass down. Right now I'm keeping it simple and just passing down the two flags we need so that we can remove all the client ids.


gregjopa (approved):
💯

PR #493, opened 2022-10-25
Support CP4 vaulting with ID token

elizabethmv:
### Description
This PR updates the vault capture payment flow to support the CP4 use case with an ID token. When calling validate-payment-method and confirm-payment-source endpoints on CP4 integrations, we need to set the target subject on the access token. This will also require that the merchant ID is set or not set in the script params depending on the integration type.

The vault capture flow is used for client token returning buyer PayPal and Card and ID token returning buyer for Venmo and Card.

This change is required to support the CP4 integration for Venmo and Card using an ID token.

This change has been validated in stage for various scenarios with the PPCP team. High level repro steps are as follows with a test CP4 integration:
- Create an ID token for a vaulted Card or Venmo
- Pass the ID token to the JS SDK script attribute
- Click on the vaulted button
- Confirm Orders calls succeed

N/A

ID tokens are not enabled in production yet so this flow will remain unchanged, as the client token flow will continue to use the embedded access token.

N/A


src/payment-flows/vault-capture.js
```diff
@@ -159,6 +163,26 @@ function initVaultCapture({ props, components, payment, serviceData, config } :
         });
     };
 
+    const createAccessTokenWithTargetSubject = (): ZalgoPromise<string> => {
+        return ZalgoPromise.try(() => {
+            return createAccessToken(
+                clientID,
+                { targetSubject: merchantID[0] }
+            ).then(accessTokenWithTargetSubject => {
+                return accessTokenWithTargetSubject;
```
sioked, line 172: Shouldn't need this .then, we're not manipulating the result of the promise at all.

sioked (commented):
One minor nit, but otherwise this looks good so far

src/payment-flows/vault-capture.js
elizabethmv: Good point, had this in for some logging in stage - I will remove it.

wsbrunson (approved):

PR #494, opened 2022-10-26
fix flow types for mock setupButtons function

wsbrunson:
started doing this in this PR: https://github.com/paypal/paypal-smart-payment-buttons/pull/490

it should be its own PR at this point


dtjones404 (commented):
I'm seeing this error when I run `typecheck`, otherwise lgtm!

```
Cannot call mockSetupButton with object literal bound to options because property guestEnabled is missing in
CardEligibility [1] but exists in object literal [2] in property fundingEligibility.card. [prop-missing]

     test/client/exports.js
     246│     it('should return guestEnabled status from exports.isGuestEnabled()', async () => {
     247│         const guestEnabled = true;
     248│
     249│         await mockSetupButton({
     250│             fundingEligibility: {
 [2] 251│                 card: {
     252│                     guestEnabled,
     253│                     eligible: true,
     254│                     branded: false,
     255│                     vendors: {}
     256│                 },
     257│             }
     258│         });
     259│
     260│         let guestEnabledStatus;
     261│         try {

     node_modules/@paypal/sdk-constants/src/types.js
 [1]  68│   card?: CardEligibility,
```

wsbrunson:
> I'm seeing this error when I run `typecheck`, otherwise lgtm!
> 
> ```
> Cannot call mockSetupButton with object literal bound to options because property 
> ...
> ```

You might need to run `npm install` again, I had to update `sdk-constants` to fix the `CardEligibility` type

dtjones404:
> > I'm seeing this error when I run `typecheck`, otherwise lgtm!
> > ```
> > Cannot call mockSetupButton with object literal bound to options because property 
> > ...
> > ```
> 
> You might need to run `npm install` again, I had to update `sdk-constants` to fix the `CardEligibility` type

Yep, rm -rf node_modules and doing a fresh install fixed the error.  Thanks!

dtjones404 (approved):

PR #498, opened 2022-10-31
Add setMessage Method  to Multi Card Fields

jplukarski:
### Description

This method allows merchants to programmatically add a message to the card fields for screen readers.

[DTNOR-492](https://engineering.paypalcorp.com/jira/browse/DTNOR-492)

### Dependent Changes

[PR in Paypal-Checkout-Components](https://github.com/paypal/paypal-checkout-components/pull/2027)

@paypal/js-sdk-maintainers 

❤️  Thank you!

Co-authored-by
@jplukarski
@nidhinarendra


src/card/components/CardCVV.jsx
```diff
@@ -133,7 +135,9 @@ export function CardCVV(
     };
 
     return (
+        <Fragment>
         <input
+            aria-describedby={'card-cvv-field-description'}
```
wsbrunson, line 140: do we need this because we render these in an iframe?

or do you have an example of what types of messages the merchant would set in here?

wsbrunson (approved):

src/card/components/CardCVV.jsx
jplukarski: Yes, this gives merchants the ability to set a message on the field for screen readers.

[Here is the documentation for this method on the current version of Braintree Hosted Fields](https://braintree.github.io/braintree-web/current/HostedFields.html#setMessage)

wsbrunson: thank you for the additional info. I thought it looked strange but I figured it was because of the iframe in some way.

wsbrunson (approved):

src/card/components/CardExpiry.jsx
```diff
@@ -148,5 +153,10 @@ export function CardExpiry(
             onPaste={ onPasteEvent }
             { ...attributes }
         />
+        <AriaMessage
+            ariaMessageId={'card-expiry-field-description'}
+            ariaMessageRef={ariaMessageRef}
+        />
```
bywood, line 159: Recommend fixing the indentation.

src/card/components/CardPostalCode.jsx
```diff
@@ -119,5 +124,10 @@ export function CardPostalCode(
             minLength={ minLength }
             { ...attributes }
         />
+        <AriaMessage
+            ariaMessageId={'card-postalCode-field-description'}
+            ariaMessageRef={ariaMessageRef}
+        />
```
bywood, line 130: Recommend fixing the indentation.

src/lib/dom.js
```diff
@@ -9,3 +9,12 @@ export function getNonce() : string {
...
 }
+
+// this is admittedly a bit silly, but it makes testing
+// things where the activeElement matters much easier,
+// since we can just mock this function instead of
+// overriding properties on the document and polluting
+// other tests
+export function getActiveElement () : null | HTMLElement {
+    return document.activeElement;
+}
```
bywood, line 20: Looks like this was added for unit tests, but I don't see where it's getting used. Did the unit tests not get pushed?

bywood (approved):
Approved, but had just a couple minor comments.

src/lib/dom.js
jplukarski: We originally removed this in our [PR to add the Postal Code field due to a lint error](https://github.com/paypal/paypal-smart-payment-buttons/pull/442/files#diff-99e35f06f1318941989456309bc6679cd5dccb930d021a4a26237ea86d424ee1L13) that it was not being used.

Though it looks like it was re-added in [this PR](https://github.com/paypal/paypal-smart-payment-buttons/pull/484) a few weeks ago, so we started getting a lint error that `getActiveElement` did not exist, so we re-added it to stay aligned with the `main` branch. It is not used for this componenet.

PR #499, opened 2022-11-01
Support both client token and id-token in script tag

sioked:
### Description
Currently with a client-token integration, we automatically enable vaulting for vaultable instruments. This change prevents that from happening if an id-token was also passed in to the sdk.

We want to start promoting and transitioning vaulting functionality to use the id-token integration path, but the hosted-fields integration still requires a client token until it is replaced. This PR allows us to support both id-token and client-token data attributes so that hosted fields continues to work as expected, while SPB will use the id-token approach.

- Integrate using a vaulting test scenario with Hosted Fields and SPB
- When initializing, apply both the client-token and id-token
- Upon clicking the PayPal button, you should not see the `enableVault` mutation get called

N/A

N/A

❤️  Thank you!

### DIFF
--- a/src/middleware/vault-setup.js
+++ b/src/middleware/vault-setup.js
@@ -23,7 +23,8 @@ type VaultAutoSetupEligibleProps = {|
     intent : $Values<typeof INTENT>,
     commit : boolean,
     disableFunding : ?$ReadOnlyArray<$Values<typeof FUNDING>>,
-    disableCard : ?$ReadOnlyArray<$Values<typeof CARD>>
+    disableCard : ?$ReadOnlyArray<$Values<typeof CARD>>,
+    userIDToken: ?string,
 |};
 
 type IsFundingSourceVaultableOptions = {|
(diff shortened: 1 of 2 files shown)


src/middleware/vault-setup.js
```diff
@@ -87,13 +88,20 @@ function isFundingSourceVaultable({ accessToken, fundingSource, clientID, mercha
...
-    clientID, merchantID, buyerCountry, currency, commit, intent, disableFunding, disableCard } : VaultAutoSetupEligibleProps) : ZalgoPromise<boolean> {
+    clientID, merchantID, buyerCountry, currency, commit, intent, disableFunding, disableCard, userIDToken } : VaultAutoSetupEligibleProps) : ZalgoPromise<boolean> {
 
     return ZalgoPromise.try(() => {
         if (!clientAccessToken) {
             return false;
         }
 
+        if (userIDToken) {
+          console.log('clientAccessToken and userIDToken');
```
elizabethmv, line 99:
```suggestion
```

elizabethmv (approved):
Looks good after one small fix

src/middleware/vault-setup.js
sioked: Good catch, somehow my tired eyes missed that :-D

wsbrunson: this could be a good `calLog` though

wsbrunson (approved):

PR #502, opened 2022-11-07
Set access token for deleting vaulted card with ID token flow

elizabethmv:
### Description
Currently the deleteVault mutation is only using the client access token. We recently enabled PP issued ID tokens to enter this flow and need to correctly set the access token on the relevant requests.

We have temporarily disabled the menu on vaulted cards with an ID token, so this change technically isn't visible but it ensures that the access token is set properly for the requests.

- Create an ID token for a vaulted Card
- Pass the ID token to the JS SDK script attribute
- Click on the vaulted button menu
- Click delete card
- Confirm the deleteVault mutation has the access token set

N/A

N/A


src/payment-flows/vault-capture.js
```diff
@@ -13,13 +13,37 @@ import { TARGET_ELEMENT, BUYER_INTENT, FPTI_TRANSITION, FPTI_CONTEXT_TYPE, FPTI_
...
+
+    if (userIDToken && merchantID && merchantID[0]) {
+        return ZalgoPromise.try(() => {
+            return createAccessToken(
+                clientID,
+                { targetSubject: merchantID[0] }
+            ).then(accessTokenWithTargetSubject => {
+                vaultAccessToken = accessTokenWithTargetSubject;
+            }).catch(err => {
+                getLogger().warn('vault_access_token_with_target_subject_failure', { error: stringifyError(err) });
```
wsbrunson, line 40: should this be an `error` if we are throwing?

wsbrunson (approved):

PR #510, opened 2022-11-30
move all src test files to vitest

wsbrunson:
**NOTE** Because prettier was run over these test files, I would suggest reviewing with this link: https://github.com/paypal/paypal-smart-payment-buttons/pull/510/files?diff=split&w=1

**EDIT**
This PR has been modified to convert all tests in the `/src` folder. All test files in the `/src` folder are being run through prettier now.


gregjopa (approved):

dtjones404 (commented):
Have the tests from `order.test.js` and `sdk.test.js` been moved elsewhere?  Just wondering why they've been deleted.  Everything else LGTM 🥳

wsbrunson:
> Have the tests from `order.test.js` and `sdk.test.js` been moved elsewhere? Just wondering why they've been deleted. Everything else LGTM 🥳

I ended up deleting them both because the tests were ultimately, call a function and test that the function was called with what it was passed

dtjones404 (approved):

dtjones404:
> > Have the tests from `order.test.js` and `sdk.test.js` been moved elsewhere? Just wondering why they've been deleted. Everything else LGTM 🥳
> 
> I ended up deleting them both because the tests were ultimately, call a function and test that the function was called with what it was passed

Ah good point!

PR #511, opened 2022-12-02
PPoF

nbierdeman:
### Description

Herein lies the changes necessary to facilitate the PPoF flow. The most noteworthy change is the introduction of a new callback, `onSmartWalletEligible`. When called, it will initialize a "third" render of the PayPal button with the smart wallet instrument label(s) and logo displayed. 🔥 

### Dependent Changes

These changes need to be released to production first:

- https://github.com/paypal/paypal-checkout-components/pull/2038
- https://github.com/paypal/paypal-sdk-constants/pull/113


src/api/auth.js
```diff
@@ -127,7 +127,7 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
     });
 }
 
-export function exchangeAccessTokenForAuthCode(buyerAccessToken : string) : ZalgoPromise<string> {
+export function exchangeAccessTokenForAuthCode(buyerAccessToken : string | ZalgoPromise<string>) : ZalgoPromise<string> {
```
gregjopa, line 130: Can you share why this `ZalgoPromise<string>` change is needed for the buyerAccessToken input?

src/button/button.js
```diff
@@ -38,7 +38,10 @@ export type SetupButtonOptions = {|
     cookies : string,
     personalization : PersonalizationType,
     brandedDefault? : boolean | null,
-    featureFlags: FeatureFlags
+    featureFlags: FeatureFlags,
```
jshawl, line 41: this is the smallest thing but the other properties have a space around `:`

src/button/button.js
```diff
@@ -170,6 +181,7 @@ export function setupButton({
             getLogger()
                 .info('smart_buttons_payment_error', { err: stringifyError(err) })
                 .track({
+                    [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
```
jshawl, line 184: future refactor idea: we could create a generic `onSpbErrorHandler` and add them to `.catch` for `initiatePayment` and `initiateMenu` functions

src/button/props.js
```diff
@@ -49,12 +49,16 @@ export function getButtonProps({
...
-    featureFlags
+    featureFlags,
+    enableOrdersApprovalSmartWallet,
+    smartWalletOrderID
 } : {|
     facilitatorAccessToken : string,
     brandedDefault : boolean | null,
     paymentSource : $Values<typeof FUNDING> | null,
-    featureFlags: FeatureFlags
+    featureFlags: FeatureFlags,
```
jshawl, line 59: same comment about space around `:`

src/payment-flows/checkout.js
```diff
@@ -276,6 +278,69 @@ function initCheckout({ props, components, serviceData, payment, config, restart
                     .catch(noop);
             },
 
+            onSmartWalletEligible: ({ accessToken, eligibilityReason, locale: smartWalletLocale, orderID }) : ZalgoPromise<{| smartWalletRendered : boolean, buyerIntent : string |}> => {
+                const { country, lang } = smartWalletLocale || locale;
```
jshawl, line 282: I think `smartWalletLocale` is already destructured to `locale` in the function definition parameters, right?

so this should just be

```
const { country, lang } = locale;
```

src/payment-flows/checkout.js
```diff
@@ -276,6 +278,69 @@ function initCheckout({ props, components, serviceData, payment, config, restart
...
+                let access_token = "";
+                
+                if (accessToken) {
+                    access_token = accessToken;
+                } else if (buyerAccessToken) {
+                    access_token = buyerAccessToken;
+                }
```
jshawl, line 289: what do you think about:

```
const access_token = accessToken || buyerAccessToken || "";
```

I think that's the same... 🤔

src/payment-flows/wallet-capture.js
```diff
@@ -130,11 +130,17 @@ function initWalletCapture({ props, components, payment, serviceData, config, re
...
 
+    const planID = instrument?.planID;
+
     const createAccessToken = () => {
         if (!smartWalletPromise) {
             throw new Error(`No smart wallet found`);
         }
-                    
+
+        if(getBuyerAccessToken()) {
```
jshawl, line 140: I wasn't sure if the function call here was intentional - it looks ok per https://github.com/paypal/paypal-smart-payment-buttons/blob/387e98061c5bfc522d521ed7985839058e7c8d47/src/lib/session.js#L38-L41 but I wonder if it would be clearer to future readers to:

```js
const buyerAccessToken = getBuyerAccessToken()
if(buyerAccessToken){
  // ...
}
```

src/payment-flows/wallet-capture.js
```diff
@@ -188,7 +194,7 @@ function initWalletCapture({ props, components, payment, serviceData, config, re
             orderID:     createOrder(),
             smartWallet: smartWalletPromise
         }).then(({ orderID, smartWallet }) => {
-            const { accessToken: buyerAccessToken } = getInstrument(smartWallet, fundingSource, instrumentID);
+            const buyerAccessToken = (enableOrdersApprovalSmartWallet && getBuyerAccessToken()) || getInstrument(smartWallet, fundingSource, instrumentID).accessToken;
```
jshawl, line 197: in what situations would `getInstrument().accessToken` be different from `getBuyerAccessToken`? 

might be worth a comment about why the buyer access token changes

src/payment-flows/wallet-capture.js
```diff
@@ -287,6 +296,10 @@ function setupWalletMenu({ props, payment, serviceData, components, config, rest
                         fundingSource:     newFundingSource,
                         createAccessToken: () => {
                             return smartWalletPromise.then(smartWallet => {
+                                if(enableOrdersApprovalSmartWallet && getBuyerAccessToken()) {
```
jshawl, line 299: same comment here about setting the invocation to a variable and then using the variable in the conditional

src/props/props.js
```diff
@@ -184,19 +192,25 @@ export type Props = {|
...
-    featureFlags
+    featureFlags,
+    enableOrdersApprovalSmartWallet,
+    smartWalletOrderID
 } : {|
     facilitatorAccessToken : string,
     branded : boolean | null,
     paymentSource : $Values<typeof FUNDING> | null,
-    featureFlags: FeatureFlags
+    featureFlags: FeatureFlags,
```
jshawl, line 211: same comment about space around `:` 🙈

src/constants.js
```diff
@@ -87,9 +88,10 @@ export const FPTI_CONTEXT_TYPE = {
 };
 
 export const FPTI_STATE = {
-    BUTTON:   ('smart_button' : 'smart_button'),
-    WALLET:   ('smart_wallet' : 'smart_wallet'),
-    PXP:      ('PXP_CHECK' : 'PXP_CHECK')
+    BUTTON:            ('smart_button' : 'smart_button'),
+    WALLET:            ('smart_wallet' : 'smart_wallet'),
+    PXP:               ('PXP_CHECK' : 'PXP_CHECK'),
+    ELIGIBILITY_CHECK: ('eligibility_check' : 'eligibility_check')
```
jshawl, line 94: is this a flow thing? I've not seen `('a' : 'b')` before I think...

jshawl (commented):
LGTM I've not yet run this locally but wanted to get my comments in early. all non-blocking

wsbrunson (commented):
we should talk about where we can add tests in this repo for some of the most important new additions

src/api/auth.js
nbierdeman: I must've left this in here on accident. I just removed it. Thanks for the catch. 👍

src/button/button.js
wsbrunson: i would ignore any style stuff, we are gonna run `prettier` over this whole repo in the new year

src/button/props.js
nbierdeman: Sounds like we're gonna use prettier in the new year to standardize styles.

src/props/props.js
nbierdeman: Same comment as above re: a prettier new year. 😁

src/constants.js
nbierdeman: Yes, it's a kind of flow [type casting](https://flow.org/en/docs/types/casting/).

Screen Shot 2022-12-05 at 2 21 46 PM

src/payment-flows/checkout.js
nbierdeman: `locale` is a required `onSmartWalletEligible` parameter name (I believe its part of the billing team interface), so it gets renamed to `smartWalletLocale` to avoid a naming conflict with `initCheckout` -> `props` -> `locale`. If `smartWalletLocale` is falsey, then `country` and `lang` get destructured from `initCheckout` -> `props` -> `locale`.

nbierdeman: Here's the error I get if I do either `const { country, lang } = locale` or `const { country, lang } = locale || locale` and `onSmartWalletEligible` is called without passing in `locale`. I think this is why I renamed it `smartWalletLocale`. It's basically saying, get `country` and `lang` from `onSmartWalletEligible` -> `smartWalletLocale`, or else fall back to `initCheckout` -> `locale`. Does that make sense? I'm really having to try to remember the `why`, but I think that's it. 😆 

Screen Shot 2022-12-05 at 3 24 48 PM

src/payment-flows/wallet-capture.js
nbierdeman: Good question! @ravishekhar, do you know the difference between `getInstrument().accessToken` and `getBuyerAccessToken()`?

src/payment-flows/checkout.js
nbierdeman: Yeah, nice. Thx!

ravishekhar (changes requested):

src/props/props.js
```diff
@@ -33,6 +33,13 @@ import type { CreateOrder, XCreateOrder, CreateBillingAgreement, XCreateBillingA
 // export something to force webpack to see this as an ES module
 export const TYPES = true;
 
+export type XOnSmartWalletEligibleDataType = {|
```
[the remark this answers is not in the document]
nbierdeman: I wasn't getting any type errors without `locale`, but went ahead and added it.

src/button/logger.js
```diff
@@ -158,7 +166,19 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             logger.info(`button_render_CPL_instrumentation_not_executed`);
         }
 
-        logger.track({
+        const getFI_ID = function () : string | void {
```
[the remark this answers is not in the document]
nbierdeman: That's a much better name. 😁

src/types.js
```diff
@@ -58,6 +59,7 @@ export type CheckoutProps = {|
     onApprove : ({| accelerated? : boolean, payerID : string, paymentID : ?string, billingToken : ?string, subscriptionID : ?string, authCode : ?string |}) => ZalgoPromise<void> | void,
     onComplete : () => ZalgoPromise<void> | void,
     onAuth : ({| accessToken : string |}) => ZalgoPromise<void> | void,
+    onSmartWalletEligible? : ({| accessToken : string, eligibilityReason : string, locale? : LocaleType, orderID : string |}) => ZalgoPromise<{| smartWalletRendered : boolean, buyerIntent : string |}>,
```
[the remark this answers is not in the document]
nbierdeman: I went ahead and pushed this change.

src/payment-flows/checkout.js
```diff
@@ -276,6 +278,63 @@ function initCheckout({ props, components, serviceData, payment, config, restart
                     .catch(noop);
             },
 
+            onSmartWalletEligible: ({ accessToken, eligibilityReason, locale: smartWalletLocale, orderID }) : ZalgoPromise<{| smartWalletRendered : boolean, buyerIntent : string |}> => {
+                const { country, lang } = smartWalletLocale || locale;
+                const access_token = accessToken || buyerAccessToken || "";
+                
+                if (window.innerWidth < 300 || buyerIntent === BUYER_INTENT.PAY_WITH_DIFFERENT_FUNDING_SHIPPING || buyerIntent === BUYER_INTENT.PAY_WITH_DIFFERENT_ACCOUNT) {
```
gregjopa, line 285: For readability, can we assign `300` to a variable? Just like: https://github.com/paypal/paypal-smart-payment-buttons/blob/0317e4358a879559c99966d3eb61f88ac90e3e58/src/payment-flows/vault-capture.js#L19

src/payment-flows/checkout.js
```diff
@@ -276,6 +278,63 @@ function initCheckout({ props, components, serviceData, payment, config, restart
...
+                        [FPTI_KEY.CONTEXT_ID]:   orderID,
+                        [FPTI_KEY.CONTEXT_TYPE]: FPTI_CONTEXT_TYPE.ORDER_ID
+                    }).flush();
+                    return ZalgoPromise.resolve({
+                        smartWalletRendered: false,
+                        buyerIntent
+                    });
+                }
+
+                createOrder().then(walletOrderID => {
```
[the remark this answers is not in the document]
nbierdeman: I believe this should be a fire and forget operation, where we initialize the third render of the button and notify the caller of `onSmartWalletEligible` of that. I'll circle back if I hear differently after discussing it with @ravishekhar.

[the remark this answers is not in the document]
ravishekhar: At this point createOrder will return a memoized orderID. So this will always resolve.
We can take up an optimization here though, to return from finally block of this promise chain before reloading the buttons. Should take it up as a separate PR?

gregjopa (approved):

src/button/button.js
```diff
@@ -69,14 +72,21 @@ export function setupButton({
...
+    product
 }: SetupButtonOptions) : ZalgoPromise<void> {
     if (!window.paypal) {
         throw new Error(`PayPal SDK not loaded`);
     }
 
     const clientID = window.xprops.clientID;
 
+    if (buyerAccessToken && smartWalletOrderID) {
+        setBuyerAccessToken(buyerAccessToken);
```
sioked, line 87: cc @elizabethmv - I know we changed some logic on this recently, want to get your eyes here as well

src/payment-flows/checkout.js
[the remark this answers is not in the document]
nbierdeman: It will be called post login by checkout UI apps after determining eligibility.

src/button/button.js
elizabethmv: Sharing an update, we were able to discuss this a bit more in slack. These buyer access token cases should be specific for third render with ppof

src/button/logger.js
```diff
@@ -158,7 +166,19 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             logger.info(`button_render_CPL_instrumentation_not_executed`);
         }
 
-        logger.track({
+        const getFundingInstrumentID = function () : string | void {
+            let FI_ID;
+            
+            if (wallet?.paypal?.instruments[0]?.secondaryInstruments && wallet.paypal.instruments[0].instrumentID) {
```
elizabethmv, line 172: Is the optional chaining missing in the second part of the condition?

nbierdeman: It wasn't necessary to repeat the optional chaining for the same object and properties after `&&`, since it'll short circuit upon evaluating any falsey values in the first condition.

elizabethmv (approved):

ravishekhar (approved):

PR #518, opened 2022-12-08
Merge Multi Card Field Feature Branch into Main Branch

jplukarski:
### Description

Per conversation with @wsbrunson, we can merge our feature branch for Unbranded Multi Card Fields into the main branch and add a function to paypal-checkout-components to ensure no one that is not on a PayPal domain can access the component. 

Note: these diffs have already been reviewed in previous PRs, but feel free to review again and make any comments/suggestions on what we have so far.

Merging our feature branch into main will simplify the PR process for subsequent PRs on this component.

[PR in Checkout Components](https://github.com/paypal/paypal-checkout-components/pull/2043)

❤️  Thank you!


src/api/order.js
```diff
@@ -957,14 +958,14 @@ export function updateButtonClientConfig({ orderID, productFlow, fundingSource,
...
-type TokenizeCardOptions = {|
-    card : {|
-        number : string,
-        cvv? : string,
-        expiry? : string,
-        name? : string
-    |}
-|};
+type TokenizeCardOptions = {
+    card : {
```
nidhinarendra, line 962: These pipes meant strict types correct? is there an advantage removing them?

jplukarski: Ah, yes! We were getting a flow error since not all fields are required. The optional properties were contradicting the strict types.

wsbrunson: one trick you can do:
```
type SomeType = $Shape<{|
  field: string | void,
  otherField: string,
|}>
```

then Flow will make sure that you don't pass properties that you haven't defined but won't make you pass every property. You can nest them too:

```
type SomeType = $Shape<{|
  field: string | void,
  otherField: string,
  recordType: $Shape<{|
    howdy: string
  |}>
|}>
```

jplukarski: @wsbrunson That still leads to the error: `Cannot call tokenizeCard with object literal bound to the first parameter because inexact Card [1]
is incompatible with exact object type [2] in property card. [incompatible-exact]`

wsbrunson: Where does this type get used?

jplukarski: It gets used in the `tokenizeCard` call => https://github.com/paypal/paypal-smart-payment-buttons/blob/Unbranded-Multi-Card-Fields/src/card/interface.js#L309

PR #525, opened 2022-12-21
Use mutation during onShippingChange patch call

rjstires:
### Description
🚧 👷‍♂️ 🚧

This draft PR is proposing to use the patchShipping GQL mutation, instead of the patchOrder REST request, when updating shipping values from within the onShippingChange callback.

The exposed API does not change, but internally this brings onShippingChange more inline with the newer onShippingAddressChange and onShippingOptionsChange.

❤️  Thank you!

### DIFF
--- a/src/props/onShippingChange.js
+++ b/src/props/onShippingChange.js
@@ -3,7 +3,7 @@
 import { ZalgoPromise } from '@krakenjs/zalgo-promise/src';
 import { COUNTRY, FPTI_KEY } from '@paypal/sdk-constants/src';
 
-import { patchOrder, type OrderResponse } from '../api';
+import { patchShipping, patchOrder, type OrderResponse } from '../api';
 import { FPTI_TRANSITION, FPTI_CONTEXT_TYPE, FPTI_CUSTOM_KEY } from '../constants';
 import { getLogger } from '../lib';
 import type { OrderAmount, FeatureFlags } from '../types';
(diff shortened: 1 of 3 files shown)


src/props/onShippingChange.js
```diff
@@ -100,12 +100,12 @@ export type OnShippingChangeActionsType = {|
     reject : (string) => ZalgoPromise<void>
 |};
 
-export function buildXShippingChangeActions({ orderID, actions, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI } : {| orderID : string, actions : OnShippingChangeActionsType, facilitatorAccessToken : string, buyerAccessToken : ?string, partnerAttributionID : ?string, forceRestAPI : boolean |}) : XOnShippingChangeActionsType {
+export function buildXShippingChangeActions({ orderID, clientID, actions } : {| orderID : string, clientID : string, actions : OnShippingChangeActionsType |}) : XOnShippingChangeActionsType {
 
     const patch = (data = {}) => {
-        return patchOrder(orderID, data, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI }).catch(() => {
+        return patchShipping({ orderID, data, clientID }).catch(() => {
```
upigilam, line 106: Thanks for raising a draft pr! This specific patch functionality is already being used in legacy guest checkout app as well. So Could you please make sure test legacy guest checkout app flows as well in staging before merging this change.

rjstires:
See #538

PR #526, opened 2022-12-21
feat: Add inputEvents for merchant defined callbacks- onChange, onBlur, onFocus, onInputSubmitRequest

siddy2181:
### Description
We are adding the following merchant-defined callBacks to the New Card Fields Components:
- onFocus (fires an event payload when a user focuses a field)
- onBlur (fires an event payload when a user unfocuses a field)
- onInputSubmitRequest (fires an event payload when a user hits the enter/return key to allow merchants to listen for attempts to submit the form via a keyboard event)

[DTNOR-498]( https://paypal.atlassian.net/browse/DTNOR-498)
[DTNOR-473](https://paypal.atlassian.net/browse/DTNOR-473)
[DTNOR-747](https://paypal.atlassian.net/browse/DTNOR-747)

 paypal-checkout-components PR: https://github.com/paypal/paypal-checkout-components/pull/2059

❤️  Thank you!


wsbrunson (approved):

src/card/components/CardCVV.jsx
```diff
@@ -117,6 +119,14 @@ export function CardCVV(
     };
 
     const onKeyDownEvent : (InputEvent) => void = (event : InputEvent) : void => {
+        if (typeof onKeyDown === 'function') {
+            if(event.keyCode === 13){
```
bywood, line 123: `keyCode` has been deprecated:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

I recommend using `key` instead.
```js
if (event.key === "Enter") {
  // ...
}
```
But this could depend on what browser support is still needed:
https://caniuse.com/keyboardevent-key

src/card/components/CardExpiry.jsx
```diff
@@ -107,8 +109,15 @@ export function CardExpiry(
     }
 
     const onKeyDownEvent : (InputEvent) => void = (event : InputEvent) : void => {
-        if (allowNavigation) {
-            navigateOnKeyDown(event, navigation);
+        if (typeof onKeyDown === 'function') {
+            if(event.keyCode === 13){
```
bywood, line 113: Same as above. Recommend using `key` instead.

src/card/components/CardName.jsx
```diff
@@ -81,6 +83,14 @@ export function CardName(
     };
 
     const onKeyDownEvent : (InputEvent) => void = (event : InputEvent) : void => {
+        if (typeof onKeyDown === 'function') {
+            if(event.keyCode === 13){
```
bywood, line 87: Same as above. Recommend using `key` instead.

src/card/components/CardNumber.jsx
```diff
@@ -215,13 +217,24 @@ export function CardNumber(
...
+        }
 
         setInputState((newState) => ({ ...newState, ...updatedState }));
         
 
     };
 
     const onKeyDownEvent : (InputEvent) => void = (event : InputEvent) : void => {
+        if (typeof onKeyDown === 'function') {
+            if(event.keyCode === 13){
```
bywood, line 231: Same as above. Recommend using `key` instead.

src/card/components/CardPostalCode.jsx
```diff
@@ -80,6 +82,14 @@ export function CardPostalCode(
     };
 
     const onKeyDownEvent : (InputEvent) => void = (event : InputEvent) : void => {
+        if (typeof onKeyDown === 'function') {
+            if(event.keyCode === 13){
```
bywood, line 86: Same as above. Recommend using `key` instead.

bywood (approved):
Just one suggestion, depending on browser support.

bywood:
Looking at the existing code, it looks like `key` is already in use:
https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/card/components/CardExpiry.jsx#L95

PR #530, opened 2023-01-06
add setupCardLogger

wsbrunson:
Adding `setupCardLogger` based on `setupButtonLogger` but with only our relevant fields added.

Also moved some properties from the common `setupLogger` that don't make sense for card-fields

This will require a PR in smartcomponentnodeweb to add in the new properties for `setupCardFields` like `buyerCountry` and `metadata`


src/card/components/page.jsx
```diff
@@ -321,11 +322,36 @@ function Page({ cspNonce, props, featureFlags } : PageProps) : mixed {
...
+        clientID,
+        partnerAttributionID,
+        sdkCorrelationID,
+        locale,
+        merchantID,
+        merchantDomain,
+        cardSessionID,
+    } = props;
+
+    setupCardLogger({
```
jplukarski, line 342: I believe calling setupCardLogger here will fire an analytic event for each field. So if a merchant renders 5 fields 5 analytic events will get fired off. Is this what we want? Or do we want to fire off an analytic event once all fields have completed rendering?

wsbrunson: interesting, `setupCard` is used for every card field? do we have a top-level parent setupCard?

wsbrunson: that would explain why we see xprops.parent.action and not xprops.action working

src/card/logger.js
```diff
@@ -0,0 +1,74 @@
...
+  return ZalgoPromise.hash({
+    pageRenderTime: getPageRenderTime(),
+  }).then(({ pageRenderTime }) => {
+    logger.track({
+      [FPTI_KEY.PAGE_LOAD_TIME]: pageRenderTime
+        ? pageRenderTime.toString()
+        : "",
+    });
+
+    logger.flush();
```
gregjopa, line 72: We should be careful of when we call `logger.flush()`. I think this always sends a network request. Ideally we want to batch up logs before sending them over the wire since merchant's often complain about the amount of network requests our sdk sends.

src/card/components/page.jsx
jplukarski: Yep, that's right. `setupCard` is called in every iFrame, and then depending on which field needs to be rendered it renders that page ([see lines 230-320](https://github.com/paypal/paypal-smart-payment-buttons/pull/530/files#diff-24115103b1e7103c0640569b02826371d3954c1d70dfef09ba5dab1f5df3f625R230-R320))

@siddy2181 and I have been trying to track down the top-level parent that calls `setupCard` but have been unable to find it so far.

src/card/constants.js
```diff
@@ -304,4 +304,9 @@ export const ALLOWED_ATTRIBUTES = [
 
 export const DEFAULT_EXPIRY_PATTERN = "{{99}} / {{9999}}";
 
-export const ZERO_PADDED_EXPIRY_PATTERN = "0{{9}} / {{9999}}";
\ No newline at end of file
+export const ZERO_PADDED_EXPIRY_PATTERN = "0{{9}} / {{9999}}";
+
+export const FPTI_HCF_KEYS = {
+  HCF_SESSION_ID: ("hcf_session_id": "hcf_session_id"),
+  HCF_CORRELATION_ID: ("hcf_correlation_id": "hcf_correlation_id"),
```
siddy2181, line 311: Why do we call it `hcf_correlation_id` and not just `correlation_id` and add `sdk component` tag to distinguish between different components, as the general PayPal convention is? Isn't it easier to build analytics on?

wsbrunson: we have two different correlation ids for the sdk:
`button|hcf_correlation_id` - the correlation id for second render
`context_correlation_id` - the correlation id for both renders (1st and 2nd)

i went with `hcf_correlation_id` because that's what we currently do with the button

src/props/props.js
```diff
@@ -91,7 +91,8 @@ export type XProps = {|
     allowBillingPayments : boolean,
 
     applePay : XApplePaySessionConfigRequest,
-    paymentRequest: ?PaymentRequest
+    paymentRequest: ?PaymentRequest,
+    hcfSessionID : string
```
wsbrunson, line 95: we should move this to `CardXProps` in `card/props.js`

jplukarski: Done

src/card/components/page.jsx
```diff
@@ -321,11 +322,40 @@ function Page({ cspNonce, props, featureFlags } : PageProps) : mixed {
     );
 }
 
-export function setupCard({ cspNonce, facilitatorAccessToken, featureFlags } : SetupCardOptions) {
+export function setupCard({ cspNonce, facilitatorAccessToken, featureFlags, buyerCountry, metadata } : SetupCardOptions) {
```
bywood, line 325: Does the merchant have to specify the `buyCountry`?

src/card/logger.js
```diff
@@ -0,0 +1,87 @@
...
+    [FPTI_HCF_KEYS.SDK_CORRELATION_ID] : sdkCorrelationID,
+    [FPTI_DATA_SOURCE.PAYMENTS_SDK] : clientID,
+    [FPTI_KEY.SELLER_ID] : merchantID?.[0]
+  }));
+
+  const tracking = {
+    [FPTI_KEY.STATE]:                           FPTI_STATE.CARD,
+    [FPTI_KEY.TRANSITION]:                      `hcf_${type}_field_rendered`,
+    [FPTI_KEY.EVENT_NAME]:                      `hcf_${type}_field_rendered`,
+  }
```
bywood, line 73: Some whitespace issues. I'd recommend sticking to a consistent use of whitespace.

src/card/components/page.jsx
jplukarski: Looks like `buyerCountry` is returned by `smartcomponentnodeweb`, though I am not sure exactly how it gets this value => https://github.paypal.com/Checkout-R/smartcomponentnodeweb/blob/develop/src/apps/card/controller.ts#L170

I am not specifying anything in my setup and getting `US` sent along so it's possible that this is based off the buyer's location reported by their browser. @wsbrunson can you confirm that's where that's coming from?

wsbrunson: `smartcomponentnodeweb` interprets buyerCountry based on headers that it gets from paypal's router. Its best guess and not always accurate. You can pass buyerCountry in the sandbox environment to simulate what your integration looks like in different countries but we ignore that parameter in production

src/card/logger.js
jplukarski: Ran prettier on this file to fix whitespace issues.

src/card/logger.js
```diff
@@ -0,0 +1,87 @@
...
+
+  return ZalgoPromise.hash({
```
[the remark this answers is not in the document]
jplukarski: Looks like we followed the pattern for Buttons, and that has a second promise to wait for called `queriedEligibleFunding`.

I'm not sure if we'll be adding this or any other additional promises this will have to wait for. @wsbrunson , thoughts?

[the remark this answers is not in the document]
wsbrunson: it's possible that we'd want to add that if we start running into eligibility issues. I'd say stick with `.hash` so its easier for any developer to come in later and put more promises in there

wsbrunson (commented):
I can't approve since this is my PR but if you are ready @jplukarski  & @siddy2181 I will merge it

jplukarski (approved):

jplukarski:
> I can't approve since this is my PR but if you are ready @jplukarski & @siddy2181 I will merge it

Added an approval on your behalf, @wsbrunson . LGTM

PR #531, opened 2023-01-06
 feat: cardfields accepts action input

cgdibble:
### Description

DTNOR-541

CardFields, in order to support Standalone Vaulting, needs to support the passing of an `action`.

TODO:
- Verify that we validate the inputs to Save correctly per these paradigms, too. Does it initialize the logger for `onApprove`?

n/a
n/a
n/a

❤️  Thank you!


src/card/props.js
```diff
@@ -84,49 +85,82 @@ export type CardProps = {|
...
 
+/**
+ * These CardFields props are disallowed when an action is also provided. 
+ * This is to prevent confusion between which flow is being used at runtime.
+ */
+const disallowedPropsWithAction = ["onApprove", "onCancel", "createOrder", "intent"]
+/**
+ * When CardFields is used with an Action, the required properties change. This is for validating the arguments in that use-case.
+ */
+function validateActionProps (xprops): CardProps { // TODO: this shouldn't be the exact return type since they are different
```
wsbrunson, line 105: does it make sense to separate `validate` from `get`? or maybe make it clearer in the name that we are getting the props if they are validated?

src/card/props.js
```diff
@@ -84,49 +85,82 @@ export type CardProps = {|
...
+      placeholder,
+      minLength,
+      maxLength,
+      cardSessionID,
+      fundingEligibility,
+      inputEvents,
+      inlinexo: experience === EXPERIENCE.INLINE,
+      export:   parent ? parent.export : xport,
+      facilitatorAccessToken,
+      action // TODO: parent.props?? Or just `action` off xprops? parent.props.action
```
wsbrunson, line 162: it should just be `xprops.action` but @thebastedo and I noticed that `xprops` might be referring to one of the child components? Something strange is happening

cgdibble: Ok, maybe I'll keep it as `xprops.action` and then once it's in the testbed take a look.

package.json
```diff
@@ -49,8 +49,8 @@
     "deploy": "npm run cdnify",
     "flow": "flow",
     "flow-typed": "rm -rf ./flow-typed && flow-typed install --ignoreDeps peer && flow-typed install jest@24 mocha@4 && flow-typed install express@4.16 && sed -i -e 's/http$IncomingMessage,/http$IncomingMessage<>,/g' ./flow-typed/npm/express_v4.16.x.js",
-    "jest:client": "jest src/ --collectCoverage --collectCoverageFrom='src/' --no-cache --runInBand --env=jsdom",
-    "karma": "cross-env NODE_ENV=test babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/karma start",
+    "jest:client": "jest src/{button,card,props}/{props,legacyProps}.test.js --runInBand --env=jsdom",
```
cgdibble, line 52: TODO: oops yeah I gotta not commit these as these are me just simplifying and speeding up local dev.

src/button/props.test.js
```diff
@@ -22,74 +25,96 @@ describe('getButtonProps', () => {
...
+      const result = getButtonProps(defaultArgs)
+      expect(result.onError).toEqual(expect.any(Function))
+      expect(getOnErrorSpy).toBeCalled()
+    })
+    
     it('should fail if createBillingAgreement & createOrder are both passed in', () => {
         window.xprops.createBillingAgreement = jest.fn();
         window.xprops.createOrder = jest.fn();
-        expect(() => getButtonProps(defaultArgs)).toThrowError();
+        expect(() => getButtonProps(defaultArgs)).toThrowError("Do not pass both createBillingAgreement and createOrder");
```
cgdibble, line 60: Note: It was not entirely related, but since these were just looking for nondescript errors being thrown, I wanted to be specific. In assertions around thrown errors providing a specific expectation really helps reduce the change of false positives that might occur from _some_ other error occurring. e.g. You expect it to throw a validation error but there's something else broken so it throws a network error

src/card/interface.js
```diff
@@ -278,9 +278,11 @@ export function submitCardFields({
   extraFields,
   featureFlags
 }: SubmitCardFieldsOptions): ZalgoPromise<void> {
-  const { intent, createOrder, onApprove, onError } = getCardProps({
+  // TODO: For next ticket, take these card props and act on them based on action flow or not
```
cgdibble, line 281: The next ticket is for HCF to _use_ these, so adjusted this accordingly to keep this PR narrower in scope. The validation refactor already expanded things enough.

src/card/interface.test.js
```diff
@@ -86,4 +86,8 @@ describe("interface", () => {
       );
     });
   });
+
+  describe('submitCardFields', () => {
+    // TODO: test drive updates to handling actions here
```
cgdibble, line 91: Felt prudent to put in for next ticket 🤷 Not sure why ha

src/props/props.js
```diff
@@ -10,25 +10,10 @@ import type { XApplePaySessionConfigRequest } from '../payment-flows/types';
 import { getStorageID, isStorageStateFresh } from '../lib';
 
 import { getOnInit } from './onInit';
-import { getCreateOrder } from './createOrder';
```
cgdibble, line 13: We ran into an issue here because Buttons/HCF all need this stuff. Which, until now, was _everything_. But, in the context of actions, that is no longer the case. So we had to refactor these properties to a separate function that otherwise does the same work, but allows us to selectively get the props in question when we do need them.

The action functionality dictates this because we want to error early on if merchants are providing things like `onApprove`, `createOrder`, etc. _along with_ an `action`. Those are mutually exclusive and different flows, so we wanted things to be clearer.

src/card/props.js
```diff
@@ -84,49 +96,136 @@ export type CardProps = {|
...
+      type,
+      branded,
+      style,
+      placeholder,
+      minLength,
+      maxLength,
+      cardSessionID,
+      fundingEligibility,
+      inputEvents,
+      inlinexo: experience === EXPERIENCE.INLINE,
```
wsbrunson, line 224: i think we can delete this

wsbrunson (approved):
this is looking great!

jplukarski (approved):

src/card/props.js
cgdibble: Consider it deleted! To clarify, though, delete as a prop entirely within the context of `cards`?

thebastedo (approved):

PR #532, opened 2023-01-09
feat: hosted card fields handle a provided action

thebastedo:
### Description

Hosted Card Fields will need to call the provided `createVaultSetupToken` and then update that token with the card details to complete the vaulting flow. This PR aims to add that functionality and enable this behavior in `hosted card fields`

❤️  Thank you!


src/card/interface.js
```diff
@@ -273,13 +274,72 @@ export function reformatExpiry(expiry: ?string): ?string {
...
+  action: SaveAction,
+  lowScopedAccessToken: string,
+  paymentSourceDetails: Card
+|}
+
+export async function vaultPaymentSource ({ action, lowScopedAccessToken, paymentSourceDetails }: VaultPaymenSourceOptions) {
+  const { createVaultSetupToken, onApprove } = action;
+
+  const { vaultSetupToken } = await createVaultSetupToken();
+  await updateVaultSetupToken({ vaultSetupToken, lowScopedAccessToken, paymentSourceDetails });
```
cgdibble, line 332: Ahh, this clicked weirdly late, but we're updating the _same_ token with new data and thus just return the original vault token once it is updated with the paymentSourceDetails. 👍

wsbrunson (approved):

tests/card/interface.test.js
```diff
@@ -0,0 +1,197 @@
...
+      fields.cardExpiryField.isValid = true;
+      fields.cardCvvField.isValid = true;
+      // $FlowFixMe
+      expect(getFieldErrors(fields).sort()).toEqual(
+        ["INVALID_NUMBER", "INVALID_NAME", "INVALID_POSTAL"].sort()
+      );
+    });
+  });
+
+  describe("updateVaultSetupToken", () => {
```
cgdibble, line 125: Can we include a test that specifies the failure scenario? How do we expect/assert `updateVaultSetupToken` to handle failures?

cgdibble (changes requested):
Some changes to expand testing to cover error scenarios/sad paths as well, plus `vaultPaymentSource` not being resolved in submit.

src/card/props.js
```diff
@@ -218,6 +219,9 @@ export function getCardProps({ facilitatorAccessToken, featureFlags } : GetCardP
       inputEvents,
       export:   parent ? parent.export : xport,
       facilitatorAccessToken,
+      // $FlowFixMe[incompatible-use]
+      // $FlowFixMe[prop-missing]
+      ...(action?.props?.action && { action: parent.props.action })
```
cgdibble, line 224: This had me wondering from before, but why `parent.props.action`? What is that in this context and why?

wsbrunson: @cgdibble we chatted about this in the planning meeting, we are about to write a ticket about why. Basically, the parent component uses the same initialization method as the children. When we get the card props here, we are actually getting them from the child components (NameField, NumberField, etc) instead of the parent CardField component.

We have a plan to stop doing this though and then this can be `props.action` instead of `parent.props.action` but we didn't want to stop this work to do it

src/card/interface.js
```diff
@@ -302,6 +359,20 @@ export function submitCardFields({
       throw new Error(`Restart not implemented for card fields flow`);
     };
 
+    if (cardProps.action !== undefined) {
+      switch(cardProps.action.type) {
+        case 'save':
+          vaultPaymentSource({
```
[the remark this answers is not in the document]
thebastedo: I couldn't get that to work with flow. the major challenge here from a flow perspective if that whole function does not have a consistent return pattern, as further down in the logic we have different places where we return void. As well is there something upstream waiting for this that we need to return the async promise?

src/card/interface.js
```diff
@@ -273,13 +274,69 @@ export function reformatExpiry(expiry: ?string): ?string {
...
+  await updateVaultSetupToken({ vaultSetupToken, lowScopedAccessToken, paymentSourceDetails });
+  await onApprove({ vaultSetupToken });
+}
+
 export function submitCardFields({
   facilitatorAccessToken,
   extraFields,
   featureFlags
 }: SubmitCardFieldsOptions): ZalgoPromise<void> {
   // TODO: For next ticket, take these card props and act on them based on action flow or not
```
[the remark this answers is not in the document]
thebastedo: Good catch, I cleaned those out but musta snuck back in in one of my merges

src/card/interface.js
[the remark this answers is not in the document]
cgdibble: Honestly Zalgo just borks the ease of use around async here and I don't care what it says haha

I guess returning the async call was my hope/attempt to try and keep it consistent. Having `vaultPaymentSource` resolve undefined/void, and returning that seemed the best way to do it. 

> As well is there something upstream waiting for this that we need to return the async promise?

Good question, I'll need to take a look locally for ease of navigation, but I think we can at least determine that the happy paths are currently doing so in here already. You're right that the `onApprove` is the thing the merchant waits on, but we should be consistent within this submit function.

> I couldn't get that to work with flow

What about the flow broke when returning here?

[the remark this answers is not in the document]
thebastedo: > I guess returning the async call was my hope/attempt to try and keep it consistent.

I agree, I'm revisiting this at the moment I'll see if I can't be smarter than flow here.

> What about the flow broke when returning here?

Mainly was the fact that at the end we return `onApprove`, which is merchant provided so we can't expect it to return a `ZalgoPromise` which is the expected return. I think I just need to figure out how to set the return as `Promise<void> | ZalgoPromise<void>`

tests/card/interface.test.js
thebastedo: Errors should bubble up the promise chain and should now be testing this in the `submitCardFields` tests

tests/card/interface.test.js
```diff
@@ -0,0 +1,197 @@
...
+            'card': {
+              ...options.paymentSourceDetails,
+            }
+          }
+        }
+      });
+    });
+  });
+
+  describe("vaultPaymentSource", () => {
```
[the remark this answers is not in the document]
thebastedo: Should be testing for those now in the `submitCardFields` unit tests

jplukarski (approved):

src/card/interface.js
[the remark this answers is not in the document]
thebastedo: I don't know exactly what I changed but I think this should be addressed, take a look here https://github.com/paypal/paypal-smart-payment-buttons/blob/a75d8ec50ba5639dfa7bed0399983e9e9a0681af/src/card/interface/submitCardFields.js#L67

tests/card/interface/gql.test.js
```diff
@@ -0,0 +1,9 @@
+import { describe, it, expect } from 'vitest';
+
+import { resetGQLErrors, emitGqlErrors } from '../../../src/card/interface';
+
+describe('card/interface/gql', () => {
+  it('should be true', () => {
+    expect(true).toBe(true);
```
cgdibble, line 7: Just noting a TODO so this doesn't fly under the radar.

tests/card/interface/submitCardFields.test.js
```diff
@@ -0,0 +1,273 @@
...
+    await expect(() => submitCardFields(options)).rejects.toThrowError('Card not available to submit');
+  })
+
+  it('should use the provided save action', async () => {
+    const onCreateVaultSetupToken = vi.fn();
+    const onApprove = vi.fn();
+
+    const mockGetCardPropsReturn = {
+      action: {
+        type: 'save',
```
cgdibble, line 119: FWIW I believe it was decided that this is capitalized. Probably worth trying to be consistent for test expectations. Do we have that defined somewhere we can even reference?

tests/card/interface/submitCardFields.test.js
```diff
@@ -0,0 +1,273 @@
...
+
+    expect(vaultPaymentSource).toHaveBeenCalledOnce();
+    expect(vaultPaymentSource).toHaveBeenCalledWith({
+      action: mockGetCardPropsReturn.action,
+      lowScopedAccessToken: 'test-access-token',
+      paymentSourceDetails: mockGetCardFieldsReturn
+    });
+  });
+
+  it('should throw an error when given an unsuported action', async () => {
```
cgdibble, line 141:
```suggestion
  it('should throw an error when given an unsupported action', async () => {
```

tests/card/interface/submitCardFields.test.js
```diff
@@ -0,0 +1,273 @@
...
+
+    getCardProps.mockReturnValueOnce(mockGetCardPropsReturn);
+
+    const options = {
+      facilitatorAccessToken: 'test-access-token',
+    };
+
+    expect.assertions(3);
+    await expect(() => submitCardFields(options)).rejects.toThrowError('Action of type testing is not supported by Card Fields')
+    expect(mockInfo).toHaveBeenCalledOnce();
```
cgdibble, line 156: Does this logger represent analytics? I want to clarify because IMO testing against general logs is flaky and not a good idea, but if this is an analytics call and thus inherently more stable as a testable value, then that works for me!

tests/card/interface/submitCardFields.test.js
```diff
@@ -0,0 +1,273 @@
...
+      facilitatorAccessToken: 'test-access-token',
+    };
+
+    expect.assertions(3);
+    await expect(() => submitCardFields(options)).rejects.toThrow('testing')
+    expect(mockInfo).toHaveBeenCalledOnce();
+    expect(mockInfo).toHaveBeenCalledWith('card_fields_vault_payment_source_failed')
+  });
+
+  it('should tokenize', async () => {
```
cgdibble, line 187: What if we organized these tests in to describes for "standalone vault" and "checkout" or something? It might help organize them better and keep mocks/etc. scoped more cleanly

tests/card/interface/updateVaultSetupToken.test.js
```diff
@@ -0,0 +1,69 @@
...
+vi.mock('@paypal/sdk-client/src', async () => {
+  const actaul = await vi.importActual('@paypal/sdk-client/src')
+
+  return {
+    ...actaul,
+    getPayPalDomain: () => mockPayPalDomain
+  }
+});
+
+describe("card/interface/updateVaultSetupToken", () => {
```
cgdibble, line 30: Can we be consistent with the top-level describes in each test file? Personally I think just the function name
 under test is the way to go, but open to others

tests/card/interface/updateVaultSetupToken.test.js
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("should call request with the correct", async () => {
```
cgdibble, line 35: What happens if this call to the API fails? What is our expectation here? Can we provide a test + assertion for this?

tests/card/interface/vaultPaymentSource.test.js
```diff
@@ -0,0 +1,51 @@
...
+vi.mock('../../../src/card/interface/updateVaultSetupToken', () => ({
+  updateVaultSetupToken: vi.fn(() => Promise.resolve()),
+}));
+
+describe("vaultPaymentSource", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("should call the provided functions", async () => {
```
cgdibble, line 15: Can we get assertions/tests around the failure cases, too?

src/card/interface/hasCardFields.js
```diff
@@ -0,0 +1,18 @@
...
+    cardNumberFrame,
+    cardCVVFrame,
+    cardExpiryFrame,
+  } = getCardFrames();
+
+  if (cardFrame || (cardNumberFrame && cardCVVFrame && cardExpiryFrame)) {
+    return true;
+  }
+
+  return false;
```
wsbrunson, line 17:
```suggestion
  return cardFrame || (cardNumberFrame && cardCVVFrame && cardExpiryFrame)
```

wsbrunson (approved):

src/card/interface/updateVaultSetupToken.js
```diff
@@ -0,0 +1,43 @@
...
+
+export function updateVaultSetupToken({
+  vaultSetupToken,
+  lowScopedAccessToken,
+  paymentSourceDetails,
+}: UpdateVaultSetupTokenOptions): ZalgoPromise<string> {
+  const vaultUrl = `${ getPayPalDomain() }/v3/vault/setup-tokens/${vaultSetupToken}/update`
+  return request({
+    method: 'post',
+    url: vaultUrl,
```
wsbrunson, line 23:
```suggestion
  return request({
    method: 'post',
    url: `/v3/vault/setup-tokens/${vaultSetupToken}/update`,
```

tests/card/interface/vaultPaymentSource.test.js
thebastedo: Those errors should bubble up through the promise chain, those get tested by the catch and error handling inside the `submitCardFields` function

tests/card/interface/updateVaultSetupToken.test.js
thebastedo: this function isn't concerned with handling that failure case. that promise failure will bubble up and get caught in the `submitCardFields` function

src/card/interface/updateVaultSetupToken.js
```diff
@@ -0,0 +1,42 @@
+/* @flow */
+
+import { getPayPalDomain } from "@paypal/sdk-client/src";
```
bywood, line 3: `getPayPalDomain` doesn't appear to be used.

src/card/interface/submitCardFields.js
```diff
@@ -0,0 +1,131 @@
...
+        case 'save':
+          try {
+            return vaultPaymentSource({
+              action: cardProps.action,
+              lowScopedAccessToken: facilitatorAccessToken,
+              paymentSourceDetails: card,
+            })
+          } catch (error) {
+            getLogger().info("card_fields_vault_payment_source_failed")
+            throw error;
```
bywood, line 74: I'm not as familiar with `ZalgoPromise`. Is this thrown error catchable? Who should be catching it?

elizabethmv (approved):

PR #534, opened 2023-01-11
HCF Vault with Purchase documentation

wickerpopstar:
### Description
Initial docs for Vault with Purchase functionality with new HCF. General flow outlined, details for specific card types coming with those specific tickets. No functionality changes required for Vault with Purchase for HCF.

[DTNOR-551](https://paypal.atlassian.net/browse/DTNOR-551)

@paypal/js-sdk-maintainers 

❤️  Thank you!

### DIFF
--- a/src/card/docs/Vaulting/README.md
+++ b/src/card/docs/Vaulting/README.md
@@ -1,0 +1,129 @@
+# Vaulting with HCF
+
+## Outline of Vaulting Flow
+1. Calling `POST /create-order` creates the order specifying the card as payment method, and indicates vaulting for the specified customer.
+2. The order is successfully created and the `orderId` is returned along with vaulting details such as the `id` of the vaulted payment method.
+3. The returned `orderId` is passed along to the `POST /capture` call and processing continues as usual.
+
+## Script tag modifications
+Include `vault=true` in the script tag to include vaulting options in the SDK. For example:
+
+```html
+<script src="https://www.paypal.com/sdk/js?client-id=<YOUR_CLIENT_ID>&components=card-fields&<OTHER_PARAMETERS>&vault=true" />
+```
+
+## Modifications to `createOrder`
+In order to vault the card, the request body of the `POST create-order` call must include the `customerId` and the instruction to vault the payment method. For example:
+```json
+{
+  payment_source: {
+    card: {
+      attributes: {
+        customer: {
+          id: "vwxj123",
+        },
+        vault: {
+          store_in_vault: "ON_SUCCESS"
+        },
+      },
+    },
+  },
+}
+```
+
+Depending on the type of card being used, several different responses may be returned to indicate the vaulting status. Generally, if vaulting occurred successfully, you will see the `details.payment_source.card.attributes.vault` field populated with an `id` field. This is the id of the vaulted payment method that can be stored and used to retrieve the vaulted payment method for subsequent transactions. Example response:
+```json
+{
+  details: {
+    payment_source: {
+      card: {
+        attributes: {
+          vault: {
+            customer : {
+              id: "vwxj123"
+            },
+            id: "7dyps8w",
+            status: "VAULTED"
+          }
+        },
+        brand: "VISA",
+        expiry: "2024-11",
+        last_digits: "1111",
+      },
+    },
+  },
+}
+```
+
+### Card Specific Requirements
+
+#### Visa
+
+#### MasterCard
+Vaulting with MasterCard uses an asynchronous process. Instead of a vault `id` being returned, a `setupToken` is returned. More details TBD.
+
+#### American Express
+Vaulting with American Express requires an additional field `partner` to be added to the `POST /create-order` payload with a value of `true`. More details TBD.
+
+### Example of vaulting modifications:
+In this example you can see the `createOrder` callback defining a payload to include a customer id and the vault command, and then included in the `fetch()` call.
+```js
+const cardField = paypal.CardFields({
+            style: styleObject,
+            createOrder: function (data, actions) {
+              const customerId = "vwxj123"
+              const payload = {
+                payment_source: {
+                  card: {
+                    attributes: {
+                      customer: {
+                        id: customerId,
+                      }
+                      vault: {
+                        store_in_vault: "ON_SUCCESS",
+                      },
+                    },
+                  },
+                };
+              };
+              return fetch("/api/paypal/order/create/", {
+                method: "post",
+                payload,
+                })
+                .then((res) => {
+                    return res.json();
+                })
+                .then((orderData) => {
+                    return orderData.id;
+                });
+            },
+            onApprove: function (data, actions) {
+                const { orderID } = data;
+                return fetch(`/api/paypal/orders/${orderID}/capture/`, {
+                method: "post",
+                })
+                .then((res) => {
+                    return res.json();
+                })
+                .then((orderData) => {
+                    // Redirect to success page
+                });
+            },
+            inputEvents: {
+                onChange: function (data) {
+                    // Handle a change event in any of the fields
+                },
+                onFocus: function(data) {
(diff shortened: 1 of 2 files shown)


wsbrunson (approved):

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,129 @@
...
+  },
+}
+```
+
+### Card Specific Requirements
+
+#### Visa
+
+#### MasterCard
+Vaulting with MasterCard uses an asynchronous process. Instead of a vault `id` being returned, a `setupToken` is returned. More details TBD.
```
bywood, line 63: `setupToken` -> `setup_token`

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### American Express
+Vaulting with American Express requires an additional field `partner` to be added to the `POST /create-order` payload with a value of `true`. More details TBD.
```
bywood, line 66: The `partner` field is specific to `xotestbednodeweb`. For American Express cards, my understanding is that the merchant needs to be subscribed to their OptBlue feature.

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,129 @@
+# Vaulting with HCF
+
+## Outline of Vaulting Flow
+1. Calling `POST /create-order` creates the order specifying the card as payment method, and indicates vaulting for the specified customer.
+2. The order is successfully created and the `orderId` is returned along with vaulting details such as the `id` of the vaulted payment method.
```
bywood, line 5: I think the vaulting details are returned in the capture call, not the create order call.

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,129 @@
...
+3. The returned `orderId` is passed along to the `POST /capture` call and processing continues as usual.
+
+## Script tag modifications
+Include `vault=true` in the script tag to include vaulting options in the SDK. For example:
+
+```html
+<script src="https://www.paypal.com/sdk/js?client-id=<YOUR_CLIENT_ID>&components=card-fields&<OTHER_PARAMETERS>&vault=true" />
+```
+
```
elizabethmv, line 14: For HCF vault with purchase using orders V2 we want to avoid using the vault param. This will enabled the auto vaulting flow and override the vault attributes in the order
```suggestion
```

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,129 @@
+# Vaulting with HCF
+
+## Outline of Vaulting Flow
+1. Calling `POST /create-order` creates the order specifying the card as payment method, and indicates vaulting for the specified customer.
```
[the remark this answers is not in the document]
wickerpopstar: Definitely—my understanding was that the testbed was effectively emulating the perspective of a merchant using SPB. I did a pass and converted some of the routes to the general PPaaS Order API routes instead of the proxies set up in the TB, but I clearly missed some. Is that the correct approach to us the Order API routes or do you think it would be better to just refer to them as the generic work being done?

src/card/docs/Vaulting/README.md
wickerpopstar: Thanks, I did not know this. This note was going off of the early approach we were using in hard coding on the TB, so that makes sense.

src/card/docs/Vaulting/README.md
wickerpopstar: It looks to me like `partner` does still get passed in the body of the `/checkout/orders` and used as part of the auth token that gets generated for that call. I thought the `/checkout/orders` route in the Orders API was the same route any merchant would need to use, is that not the case?

src/card/docs/Vaulting/README.md
wickerpopstar: I thought I had seen them returned in both. The PPaaS API reference has them in the response for both the `POST /checkout/orders` call and the `POST /checkout/orders/{id}/capture` calls.

bywood: Hmm...I don't recall seeing it in the `orders` response. I've only seen it in the `capture` response.

src/card/docs/Vaulting/README.md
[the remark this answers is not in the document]
bywood: The test bed does emulate the perspective of the merchant, but both the client and server side of the merchant's app. This only covers the client side integrating with the test bed's server side, which would also be owned by the merchant.

src/card/docs/Vaulting/README.md
bywood: My understanding is that `partner` is a construct of the `xotestbednodeweb` app to allow us to switch between different client IDs and secrets. For example: https://github.paypal.com/Checkout-R/xotestbednodeweb/blob/develop/lib/Request.js#L75-L78

So `xotestbednodeweb`'s `/xo-integrations/api/create-order` uses it to get the desired client ID and secret needed to create the access token used by the actual call to  PayPal's `/v2/checkout/orders`.

src/card/docs/Vaulting/README.md
wickerpopstar: You're definitely right about this. I must have been confusing the documentation with our tests.

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,144 @@
+# Vaulting with HCF
+
+## Outline of Vaulting Flow
+1. Define the `CardFields` `createOrder` callback to pass relevant payment information (excluding card details) and desire to vault payment method to server-side API.
+2. API calls `POST /checkout/orders` with vault specific field.
+3. The order is successfully created and the `orderId` is returned along with vaulting details such as the `id` of the vaulted payment method.
```
bywood, line 6: The vaulting details are not returned in the `/checkout/orders` response. So I recommend removing: "along with vaulting details such as the `id` of the vaulted payment method".

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,144 @@
+# Vaulting with HCF
+
+## Outline of Vaulting Flow
+1. Define the `CardFields` `createOrder` callback to pass relevant payment information (excluding card details) and desire to vault payment method to server-side API.
+2. API calls `POST /checkout/orders` with vault specific field.
```
bywood, line 5: Should we specify the full path? `/v2/checkout/orders`

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,144 @@
...
+3. The order is successfully created and the `orderId` is returned along with vaulting details such as the `id` of the vaulted payment method.
+4. Define the `CardFields` `onApprove` callback to pass the returned `orderId` to server-side API.
+5. API calls `POST /checkout/capture`, vault details are returned and processing continues as usual.
```
bywood, line 8: Path should be: `/v2/checkout/orders/${orderId}/capture`

bywood (changes requested):

src/card/docs/Vaulting/README.md
wickerpopstar: I was considering that. I was originally thinking to keep it more generic, but I suppose we're really only building this for latest functionality. I'll add it in now, and it can be removed in the future if need be.

bywood (approved):

src/card/docs/Vaulting/README.md
```diff
@@ -0,0 +1,154 @@
...
+            {
+              "href": "https://api-m.paypal.com/v2/checkout/orders/5O190127TN364715T",
+              "rel": "up",
+              "method": "GET"
+            }
+          ]
+        }
+      }
+    }
+  },
```
wsbrunson, line 72:
```suggestion
  }
```

wsbrunson (approved):

PR #539, opened 2023-01-25
Second Render Initialization for HCF

jplukarski:

[Jira](https://paypal.atlassian.net/jira/software/c/projects/DTNOR/boards/22228?modal=detail&selectedIssue=DTNOR-764)

❤️  Thank you!

### DIFF
--- a/src/card/components/page.jsx
+++ b/src/card/components/page.jsx
@@ -1,16 +1,12 @@
 /* @flow */
 /** @jsx h */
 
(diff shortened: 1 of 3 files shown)


jplukarski:
After looking into this further, we realized this does not change the behavior of the sdk in the way we thought it would, so we are going to close this PR and move forward with each field reporting individually that they have completed setting up.

PR #540, opened 2023-01-26
Standalone Card Vault - Round 2

wsbrunson:
This was a PR to add callback-ify the merchant provided callback functions but it turned into something more:

* treat `createVaultSetupToken` and `onApprove` in the save action the way we do other merchant-provided callbacks in the sdk
* `/api/vault` - these are the helpers we will need to get and update a vault setup token from the api
* `/card/interface/vault` - this will be where we create the helpers for our components to do vault actions. You can think of these helpers as the combination between merchant-provided callbacks and our api calls on their behalf. currently we have `vault.create`. In the future, we will most likely include `vault.update`, `vault.delete`, and possibly `vault.list`
* Moved some existing test files to vitest + prettiered them. This repo will eventually be moved fully over to vitest and prettier, its easier if we make sure all of our new files are vitested and prettiered from the start


src/card/props.js
```diff
@@ -133,8 +135,24 @@ function validateActionProps (xprops) : ActionProps {
...
+        createVaultSetupToken: getCreateVaultSetupToken({
+          createVaultSetupToken: action.createVaultSetupToken,
+        }),
+        onApprove: getSaveActionOnApprove({
+          onApprove: action.onApprove,
+          onError: baseProps.onError,
+        }),
+      };
+    }
+    default: {
```
thebastedo, line 151: In card fields i throw an error for the default case here stating that the action isn't supported. Do we want to have consistent behavior here? If so should the behavior be to throw errors for unknown actions or just return it?

src/config.js
```diff
@@ -9,6 +9,8 @@ export const ORDERS_API_URL = '/v2/checkout/orders';
 export const PAYMENTS_API_URL = '/v1/payments/payment';
 export const CREATE_SUBSCRIPTIONS_API_URL = '/v1/billing/subscriptions';
 export const VALIDATE_PAYMENT_METHOD_API = 'validate-payment-method';
+export const VAULT_API_URL = '/v3/vault'
+export const VAULT_SETUP_TOKENS_API_URL = `${VAULT_API_URL}/setup-tokens`
```
thebastedo, line 13: We should add a todo to use these in card fields 👍

src/props/createVaultSetupToken.js
```diff
@@ -0,0 +1,31 @@
...
+  ({
+    createVaultSetupToken,
+  }: {|
+    createVaultSetupToken: XCreateVaultSetupToken,
+  |}): ?CreateVaultSetupToken =>
+  () =>
+    createVaultSetupToken(
+      // $FlowIssue empty object
+      {}
+    ).then((vaultSetupToken) => {
```
thebastedo, line 23: I think currently in hosted card field the expected return/destructure from `createVaultSetupToken()` is `{ vaultSetupToken }`, we still need to align that with the response handling needs as well.

src/card/props.js
wsbrunson: yeah might as well, since this is exposed to the merchant we could get literally anything here

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+  status: "APPROVED" | "CREATED" | "PAYER_ACTION_REQUIRED" | "TOKENIZED" | "VAULTED",
+  payment_source: Object,
+  links: $ReadOnlyArray<{|
+    href: string,
+    rel: "approve" | "confirm" | "self",
+    method: "GET" | "POST",
+  |}>,
+|};
+
+export const getSetupVaultToken = ({
```
elizabethmv, line 25: Can we swap this name? getVaultSetupToken

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
+import { stringifyError } from "@krakenjs/belter/src";
+
+import { HEADERS, PREFER } from "../constants";
+import { getLogger } from "../lib";
+import { VAULT_SETUP_TOKENS_API_URL } from "../config";
+
+import { callRestAPI, getErrorResponseCorrelationID } from "./api";
+
+type SetupVaultTokenResponse = {|
```
elizabethmv, line 11:
```suggestion
type VaultSetupTokenResponse = {|
```

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+  payment_source: Object,
+  links: $ReadOnlyArray<{|
+    href: string,
+    rel: "approve" | "confirm" | "self",
+    method: "GET" | "POST",
+  |}>,
+|};
+
+export const getVaultSetupToken = ({
+  setupVaultToken,
```
elizabethmv, line 26:
```suggestion
  vaultSetupToken,
```

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+  facilitatorAccessToken,
+  partnerAttributionID,
+}: {|
+  setupVaultToken: string,
```
elizabethmv, line 30:
```suggestion
  vaultSetupToken: string,
```

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+  facilitatorAccessToken: string,
+  partnerAttributionID: ?string,
+|}): ZalgoPromise<SetupVaultTokenResponse> =>
```
elizabethmv, line 33:
```suggestion
|}): ZalgoPromise<VaultSetupTokenResponse> =>
```

src/api/vault.js
```diff
@@ -0,0 +1,49 @@
...
+  callRestAPI({
+    accessToken: facilitatorAccessToken,
+    url: `${VAULT_SETUP_TOKENS_API_URL}/${setupVaultToken}`,
```
elizabethmv, line 36:
```suggestion
    url: `${VAULT_SETUP_TOKENS_API_URL}/${vaultSetupToken}`,
```

elizabethmv (commented):
A few more instances where the name is swapped, but looks good otherwise!

wsbrunson:
> A few more instances where the name is swapped, but looks good otherwise!
Good eye, thank you!

src/card/interface/vault.js
```diff
@@ -0,0 +1,76 @@
...
+const updateVaultSetupToken = ({
+  vaultSetupToken,
+  lowScopedAccessToken,
+  paymentSourceDetails,
+}: UpdateVaultSetupTokenOptions): ZalgoPromise<string> =>
+  request({
+    method: "post",
+    url: `/v3/vault/setup-tokens/${vaultSetupToken}/update`,
+    headers: {
+      Authorization: `Basic ${lowScopedAccessToken}`,
```
elizabethmv, line 23:
```suggestion
      Authorization: `Bearer ${lowScopedAccessToken}`,
```

wsbrunson: whoops, i forgot to delete this and import from `/api`

src/api/vault.js
```diff
@@ -0,0 +1,107 @@
...
+    | "VAULTED",
+  payment_source: Object,
+  links: $ReadOnlyArray<{|
+    href: string,
+    rel: "approve" | "confirm" | "self",
+    method: "GET" | "POST",
+  |}>,
+|};
+
+export const getVaultSetupToken = ({
```
cgdibble, line 71: This is scoped to "Vault", but is there anyway you see to organize these by Standalone v Vault w/ Purchase? I feel like it will get confusing as hell down the road if we intermingle it too much.

cgdibble (approved):

elizabethmv (approved):

src/api/vault.js
wsbrunson: the vault w/ purchase flow doesn't actually use any of the vault apis. These would only be for standalone vault. vault w/ purchase uses orders v2 and passes a flag in the request body that triggers the backend to vault to the funding instrument used for purchase

PR #543, opened 2023-02-08
Standalone Vault Pt 3

wsbrunson:
### Summary
* new graphql endpoint for updating vault setup token
* remove `action` in favor of new interface change
* format payment source for vault api
* rename `interface/vault` to `interface/standalone-vault`
* Misc
  * update types to reflect actual values
  * remove unneeded `INTENT` checks (we don't allow client-side capture)
  * update/fix/add tests (and remove some that didn't matter anymore)

### New graphql endpoint for updating vault setup token
The vault api team has an endpoint for updating a vault setup token with payment fields. Unfortuantely the api is only available internally. We created a new mutation in graphql to be a 1:1 proxy for this internal endpoint while we wait for the vault to to externalize it.

### Remove `action` in favor of new interface change
We decided against the proposed `paypal.actions.Save` feature because we are no longer supporting a UI-less save action in the first release. This will have to be used in conjunction with `paypal.Buttons` or `paypal.CardFields`. In the future, we may explore a way for merchants to perform actions like Save outside of our paypal-hosted iframes but for now we are sticking with the current patterns.

### Format payment source for vault api
We need to match the expected object structure for a payment source for the vault v3 api, including the right date format and calling fields `security_code` instead of `cvv`

### Rename `interface/vault` to `interface/standalone-vault`
it was pointed out that using terms like `vault` without the standalone modifier can be confusing when there's two ways to vault: vault with purchase and vault without purchase. Standalone vault is vault without purchase, our hope is to make it clear to any future engineers exactly what flow they are dealing with.


src/card/interface/submitCardFields.js
```diff
@@ -33,6 +33,17 @@ type SubmitCardFieldsOptions = {|
   |}
 |};
 
+const convertMMYYToPaymentSourceDate = (dateString = "") => {
+  const dateArray = dateString.split('/');
+  const month = parseInt(dateArray[0] | "", 10);
```
bywood, line 38: Was this meant to be `dateArray[0] || ""`. However, `parseInt("", 10)` would return `NaN`, which would create an invalid date further down.

src/card/interface/submitCardFields.js
```diff
@@ -33,6 +33,17 @@ type SubmitCardFieldsOptions = {|
...
+  const year = parseInt(dateArray[1] | "", 10);
```
bywood, line 39: Same as above.

bywood (changes requested):

src/api/vault.js
```diff
@@ -94,23 +96,52 @@ export const getVaultSetupToken = ({
     },
   });
 
+export type PaymentSourceInput = {|
```
cgdibble, line 99: I find it hard to believe this isn't defined somewhere already that could be re-used...but hell either way it looks good.

src/api/vault.js
```diff
@@ -94,23 +96,52 @@ export const getVaultSetupToken = ({
...
+          paymentSource: $paymentSource
+        ) {
+          id,
+          status
+        }
+      }`,
+    variables: {
+      clientID,
+      idToken,
+      vaultSetupToken,
```
cgdibble, line 144: Given this is Standalone-specific, we could name this local var to be more explicitly `emptyVaultSetupToken`? Though, tokens are so complex maybe calling it empty is a potentially confusing assumption ha

src/card/interface/vault.js
```diff
@@ -12,27 +16,37 @@ type SaveAction = {|
...
+  idToken,
   paymentSource,
 }: VaultPaymenSourceOptions): ZalgoPromise<void> => {
   const { createVaultSetupToken, onApprove } = action;
 
   // happy path for now
   // need to add error cases with fpti events for each .then
   return createVaultSetupToken().then((vaultSetupToken) => {
-    return updateVaultSetupToken({
+    return getVaultSetupToken({
```
cgdibble, line 37: I'm a little confused on this ordering now. What is `getVaultSetupToken` doing? I thought standalone just had to make the one API call and it was the step done in `updateVaultSetupToken`.

src/card/interface/submitCardFields.js
wsbrunson: I'm going to write tests for this function and I can experiment with using strings vs numbers. I think the `Date` api expects `Date(number, number, number)` if you are going that route instead of passing a date string but i'm not sure

src/api/vault.js
wsbrunson: you and me both brother

wsbrunson: @gregjopa and I keep talking about putting all of our Types (api request/response, general sdk types) in a central location. Not sure what that package is, most likely it will be TypeScript though

src/api/vault.js
wsbrunson: yeah at this point it is a `string` so while we know its empty because it was just created, it doesn't change as we go through the flow, its always the same `string`

src/card/interface/vault.js
wsbrunson: I think we will understand this more when we do the PayPal wallet. For the PayPal Wallet, we would get the contingency when the token is created, we'd need to do a `get` to check for said contingency. I think with Cards, we don't get a contingency until we update it with the card details. At that point, we'd need to display the 3DS screen to approve.

I think if this ends up being the case, we could have separate functions for vaulting wallets vs cards

src/card/interface/vault.js
```diff
@@ -12,27 +16,37 @@ type SaveAction = {|
 
 type VaultPaymenSourceOptions = {|
   action: SaveAction,
+  clientID: string,
+  idToken: string,
```
elizabethmv, line 20: Should all of these `idToken` be `userIDToken`?

src/card/interface/vault.js
```diff
@@ -12,27 +16,37 @@ type SaveAction = {|
...
   // happy path for now
   // need to add error cases with fpti events for each .then
   return createVaultSetupToken().then((vaultSetupToken) => {
-    return updateVaultSetupToken({
+    return getVaultSetupToken({
       vaultSetupToken,
       facilitatorAccessToken,
-      paymentSource,
-      partnerAttributionID: ''
+      partnerAttributionID: "",
```
elizabethmv, line 40: Should we take this `partnerAttributionID` off until we know if we need it?

src/card/interface/submitCardFields.js
```diff
@@ -33,6 +33,17 @@ type SubmitCardFieldsOptions = {|
   |}
 |};
 
+const convertMMYYToPaymentSourceDate = (dateString = "") => {
```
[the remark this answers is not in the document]
wsbrunson: oof i did not know that

src/card/interface/submitCardFields.js
wsbrunson: oh wait i see what you are saying now. we should just throw an error if we don't have a date string here

bywood: Yeah, if a thrown error would be handled correctly, then I think you could just do something like:
```js
  const month = parseInt(dateArray[0], 10);
  if (isNaN(month)) {
    // throw some error
  }
```

src/api/vault.js
cgdibble: Yeah that's a good perspective, I'm for it staying for sure. That's a great example of the `empty` verbiage being confusing because I was meaning the token itself doesn't have an associated payment source. But, yeah, carry on bad suggestion is bad lol

src/card/interface/vault.js
cgdibble: I'm confused more then...what's the new(ish) need of `getVaultSetupToken`? We get the token from the merchant via `createVaultSetupToken`, so what is `getVaultSetupToken` doing against the API and why?

src/api/vault.js
wsbrunson: not a bad suggestion at all! we are all continuously updating our mental models about what all this means

src/card/props.js
```diff
@@ -123,69 +123,77 @@ type BaseCardProps = {|
...
+ * These CardFields props are disallowed when save is also provided.
  * This is to prevent confusion between which flow is being used at runtime.
  */
-const disallowedPropsWithAction = [
+const disallowedPropsWithSave = [
   "onApprove",
   "onCancel",
   "onComplete",
   "createOrder",
-  "intent",
```
cgdibble, line 150: Does `intent` have a role in this now? Something might well have changed, but wanna make sure so my understanding stays up to date. I'm adding these disallowed props to the docs now.

cgdibble: Resolved offline. To try and paraphrase, intent wouldn't even come in from this means anyway, but we at least can and want to scope the disallowed list to callbacks and things the merchant passes directly to CardFields.

src/card/interface/vault.js
wsbrunson: Charlie and I talked about this in a meeting, for some vault cases we have to do a GET on the id token to check for contingencies. it's possible this will go away though once we better understand vaulting for cards and wallets

src/card/interface/submitCardFields.js
wsbrunson: ended up getting rid of this completely

src/card/interface/index.js
```diff
@@ -8,4 +8,4 @@ export * from './hasCardFields';
 export * from './isEmpty';
 export * from './reformatExpiry';
 export * from './submitCardFields';
-export * from './vault';
+export * from './standalone-vault';
```
elizabethmv, line 11: Non blocking, `vault-without-purchase` might be even more clear on the flow

jplukarski (approved):

src/card/props.js
```diff
@@ -125,86 +125,98 @@ type BaseCardProps = {|
...
 ];
 /**
  * When CardFields is used with an Action, the required properties change. This is for validating the arguments in that use-case.
```
elizabethmv, line 160: Missing an update in this comment I think? Action -> Save

src/card/interface/standalone-vault.js
```diff
@@ -0,0 +1,46 @@
...
+  save,
+  facilitatorAccessToken,
+  clientID,
+  userIDToken,
+  paymentSource,
+}: VaultPaymenSourceOptions): ZalgoPromise<void> => {
+  const { createVaultSetupToken, onApprove } = save;
+
+  // happy path for now
+  // need to add error cases with fpti events for each .then
```
elizabethmv, line 30: Bumping a reminder here for a follow up PR to check on the token statuses, etc.

elizabethmv (approved):
Some minor comments. Awesome work!

src/card/interface/standalone-vault.js
wsbrunson: yeah i should turn this into a jira ticket

PR #545, opened 2023-02-14
Fix - Reset the LSAT upgrade error state on retry

ravishekhar:
### Description
This change will fixes the Upgrade LSAT error state and resets the error state on retry.  

This change is needed as on Upgrade LSAT retries the error state still indicates it has errored. 

❤️  Thank you!


gregjopa (approved):

src/api/auth.js
```diff
@@ -78,7 +78,7 @@ let lsatUpgradeCalled : boolean = false;
 let lsatUpgradeError : ?mixed;
 
 export const onLsatUpgradeCalled = () => {
-    lsatUpgradeCalled = false;
+    lsatUpgradeCalled = true;
```
ravishekhar, line 81: This line is covered when investigating the added test case manually. However the coverage reporter is not counting this line covered. I will look into it separately.

src/api/auth.js
```diff
@@ -99,6 +99,7 @@ export const clearLsatState = () => {
 };
 
 export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, { buyerAccessToken, orderID } : {| buyerAccessToken : string, orderID : string |}) : ZalgoPromise<void> {
+    clearLsatState();
```
ravishekhar, line 102: Same comment as above.
> This line is covered when investigating the added test case manually. However the coverage reporter is not counting this line covered. I will look into it separately.

PR #546, opened 2023-02-17
Remove env logic from isNativeEligible()

spencersablan:
**Jira**:  [LI-4619](https://paypal.atlassian.net/browse/LI-4619)

On stage right now, if you use the Venmo button when trying to create a billing agreement, it defaults to the fallback.  We don't need this logic anymore. 😄

1. On stage, try to create a billing agreement with Venmo.
2. Before this change: You will see the fallback modal saying "We weren't able to open Venmo".
3. After this change: You should see the normal popup

@elizabethmv 

❤️  Thank you!

### DIFF
--- a/src/payment-flows/native/eligibility.js
+++ b/src/payment-flows/native/eligibility.js
@@ -176,10 +176,6 @@ export function isNativeEligible({ props, config, serviceData } : IsEligibleOpti
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


elizabethmv (approved):
Some minor fixes for CI, but looks good after!!

wsbrunson (approved):

PR #550, opened 2023-02-23
fix: bug where postal code was not being send to confirm payment sour…

jplukarski:
### Description

Fix bug where postal_code was not being sent to confirm_payment_source

[DTNOR-830](https://paypal.atlassian.net/browse/DTNOR-830)

❤️  Thank you!


wsbrunson (approved):

gregjopa:
@jplukarski I'm seeing the following failure in CI:
```
 ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  tests/card/interface/submitCardFields.test.js > submitCardFields > should checkout
AssertionError: expected "spy" to be called with arguments: [ 'test-order-id', …(2) ]expected "spy" to be called with arguments: [ 'test-order-id', …(2) ]

Received: 
    1st spy call:

    Array [
      "test-order-id",
      Object {
        "payment_source": Object {
  +       "billing_address": Object {
  +         "postal_code": "91210",
  +       },
          "card": Object {
            "expiry": "24-01",
            "name": "John Doe",
            "number": "4111111111111111",
```

wsbrunson (approved):

src/api/vault.js
```diff
@@ -14,11 +14,11 @@ export type PaymentSourceCardDetails = {|
   type?: "CREDIT" | "DEBIT" | "PREPAID" | "STORE" | "UNKNOWN",
   brand?: string,
   billing_address?: $Shape<{|
```
wsbrunson, line 16: I would remove the `$Shape` helper now that there are required fields and optional fields

src/api/vault.js
```diff
@@ -87,9 +87,14 @@ export type PaymentSourceInput = {|
     number: string,
     expiry: string,
     name?: string,
-    securityCode?: string,
+    securityCode: string,
     billingAddress?: $Shape<{|
```
wsbrunson, line 91: same here about Shape

src/card/interface/submitCardFields.js
```diff
@@ -66,24 +56,21 @@ export function submitCardFields({
...
-            ...extraFields,
-          };
+          let payment_source
 
-          if (card.name) {
-            cardObject.name = card.name;
+          if (extraFields) {
+            payment_source = convertCardToPaymentSource(card, extraFields) 
+          } else {
+            payment_source = convertCardToPaymentSource(card)
```
wsbrunson, line 64: i'm not following why we would need to call this with different arguments? Would `convertCardToPaymentSource(card: Card, extraFields: ?ExtraFields)` work?

src/card/types.js
```diff
@@ -16,12 +16,31 @@ export type SetupCardOptions = {|
...
+    countryCode: string,
+|}
+
+export type ReformattedBillingAddress = {|
+    address_line_1?: string,
+    address_line_2?: string,
+    admin_area_2?: string,
+    admin_area_1?: string,
+    postal_code?: string,
+    country_code?: string,
```
wsbrunson, line 34: should this be?
```suggestion
    country_code: string,
```

oscarleonnogales: I believe so, great catch!

src/card/interface/submitCardFields.js
siddy2181: Good catch!

wsbrunson (approved):

PR #552, opened 2023-02-24
Set paymentSource to paypal for PPoF change FI flow

nbierdeman:
### Description

There's a bug where merchants' `onApprove()` `data.paymentSource` is being set to `null` when a buyer clicks the menu button -> "Pay with a different funding source" -> "Pay Now" during a PPoF flow.

This PR fixes the aforementioned bug by properly setting `paymentSource` to `paypal` during a "change FI" PPoF flow.

### Reproduction Steps

- Set up a test page where you log `data` in the `onApprove` callback
- Walk through a PPoF flow on the `main` branch, but instead of clicking the PayPal button after the smart wallet renders, click the menu button and "Pay with a different funding source"
- Click "Pay Now" in the popup
- Notice that `data.paymentSource` is set to `null` in the browser console
- Repeat the above steps, but on the `ppof-paymentSource-fix` branch
- Notice that `data.paymentSource` is now set to `paypal`


nbierdeman:
@ravishekhar @jshawl, marking this as a draft while I doublecheck the changes and look into adding a test, but feel free to try it out if you have some bandwidth.

wsbrunson (commented):
What if we took the same approach we take for initiating a payment?

```
const paymentProps = getButtonProps({ paymentSource: paymentFundingSource, ...otherProps });
const payPromise = initiatePayment({ payment, props: paymentProps });
```

we have this inside the `if (menuToggle)` under `onElementClick`

```
const menuProps = ... get menu-specific props ...
const menuPromise = initiateMenu({ payment, props: menuProps });
```

is that overkill though? idk what the implications are of adding a payment source to `getProps` vs keeping it as null

wsbrunson:
> > What if we took the same approach we take for initiating a payment?
> 
> @wsbrunson, `getOnApproveOrder()` `data.paymentSource` is getting set on line 107 in `src/button/button.js` during the PPoF change FI -> Pay Now flow, bypassing `onElementClick()` for both `button` and `menuToggle`. Please let me know though if I'm not understanding what you're suggesting.

that makes perfect sense, thank you for entertaining me!

wsbrunson (approved):

jshawl (approved):
:shipit: changes look good! 🚢

PR #554, opened 2023-02-27
Removed item category check for billing agreement integration with donate label

arielxin:
### Description

Item category `DONATION` validation is in placed when `donate` button label is being used, to ensure that it is being used for donation use case only. Recently we got merchant reported issue saying `donate` label is not working with billing agreement integration. Since for billing agreement, item category can be passed at the capture time, we're removing this validation for billing agreement flow.

Screen Shot 2023-02-27 at 1 51 31 PM

❤️  Thank you!


jshawl (commented):
what do you think about disabling the `throw` here instead of skipping the validation error?

https://github.com/paypal/paypal-smart-payment-buttons/pull/554/files#diff-216163d3983a51c09c007334d1942f4569c530384ef8f2cdcf62cf2880a28351R217

In the PR description - I'm not sure where I would pass the category - do you have a snippet available where developers can set the category? thank you!

arielxin:
@jshawl we should still throw error and block the payment from being initiated for orders V2 and subscription. 

Category could be set as part of the order V2 payload in purchase_unit. 

Our expectation would be as follows:
donate label + Orders V2: run validation and block payment if item category is not passed as DONATION

donate label + Subscription: block completely since donation is not yet enabled for Subscriptions. There is no category in the subscription payload so this validation will always fail and throw error.

donate label + Billing Agreement: skip this validation

jshawl (approved):
Thank you for the details about the validation logic! changes look good to me. 

I also took a look to see if it would be a quick win to add a test case but that might not be feasible for this change - https://github.com/paypal/paypal-smart-payment-buttons/blob/main/test/client/validation.js - I'm ok with leaving out the test for now but will request review from one other member hang tight!

arielxin:
@jshawl One more question after reading other validations in the same file. Do you know if the check I added covers all billingagreement use cases? I saw `!cartBillingType` for some other validations. Will billing agreement be supported without `createBillingAgreement`?

ravishekhar:
> @jshawl One more question after reading other validations in the same file. Do you know if the check I added covers all billingagreement use cases? I saw `!cartBillingType` for some other validations. Will billing agreement be supported without `createBillingAgreement`?

Billing Agreement flow can be triggered even with createOrder

arielxin:
@ravishekhar do you have any suggestion on what would be the sufficient check to exclude billing agreement for this validation? Also do you recommend to add `!cartBillingType` as well?

Can we still have this PR merged today to unblock the merchant reporting the issue? And we can add other required checked in a separate PR.

ravishekhar:
> @ravishekhar do you have any suggestion on what would be the sufficient check to exclude billing agreement for this validation? Also do you recommend to add `!cartBillingType` as well?
> 
> Can we still have this PR merged today to unblock the merchant reporting the issue? And we can add other required checks in a separate PR.

Yes you can send the fix for other billing use cases as a separate PR. Checking the cart data would be the best way.

PR #555, opened 2023-02-27
refactored reformatPaymentSource function

oscarleonnogales:

❤️  Thank you!


jplukarski (approved):

src/card/lib/card-utils.test.js
```diff
@@ -699,5 +699,43 @@ describe('card utils', () => {
 
             expect(reformatPaymentSource(paymentSource)).toStrictEqual(reformatted_payment_source)
         })
+
+        it.only('should remove billingAddress from payment source object when undefined', () => {
```
wsbrunson, line 703:
```suggestion
        it('should remove billingAddress from payment source object when undefined', () => {
```

src/card/lib/card-utils.test.js
```diff
@@ -699,5 +699,43 @@ describe('card utils', () => {
...
+                number: '4111111111111111',
+                expiry: '2024-01',
+                name: 'John Doe',
+                security_code: '123'
+            }
+
+            expect(reformatPaymentSource(paymentSource)).toStrictEqual(reformatted_payment_source)
+        })
+
+        it.only('should remove billingAddress from payment source object when empty', () => {
```
wsbrunson, line 722:
```suggestion
        it('should remove billingAddress from payment source object when empty', () => {
```

src/card/lib/card-utils.js
```diff
@@ -340,11 +340,18 @@ export function reformatPaymentSource(paymentSource: PaymentSourceCardInput) : R
...
 
-        if (key === 'billingAddress' && paymentSource['billingAddress']) {
-           // $FlowIssue
-           newObj[transformedKey] = reformatPaymentSource(paymentSource[key])
+        if (key === 'billingAddress') {
+            if (paymentSource.billingAddress && Object.keys(paymentSource.billingAddress).length !== 0) {
+                Object.keys(paymentSource.billingAddress).forEach((billingKey) => {
+                    const snakeCaseBillingKey = camelToSnakeCase(billingKey);
+                    paymentSource.billingAddress[snakeCaseBillingKey] = paymentSource.billingAddress[billingKey];
+                    delete paymentSource.billingAddress.billingKey;
```
wsbrunson, line 348: do we need to delete keys from the input if we are ultimately returning a new object (`newObj`)?

oscarleonnogales: We're essentially replacing the nested keys inside of billingAddress (addressLine1, addressLine2, etc.) after adding the camel_case conversion. It'll only go into this snippet when billingAddress exists and it's non-empty.

PR #558, opened 2023-02-28
(feat): Loggin for Submit card fields

siddy2181:
### Description
We want to track events for the usage of Hosted Card Fields so we can detect any issues during the transaction through Card Fields 
#### Logs covered in this PR
- Error on submitCardFields
- createOrder, onApprove, confirm-card-payment
- Successful transaction: onApprove

JIRA:
https://paypal.atlassian.net/browse/DTNOR-732
https://paypal.atlassian.net/browse/DTNOR-735

hcf_error_logs

#### Error
[Search link](https://github.com/paypal/paypal-smart-payment-buttons/pull/558/files#diff-84bf1f3b16b25fd5ae8707244a976f19382c981728d3e234d78f25bf57cf3446R82)
Screenshot 2023-03-13 at 4 17 51 PM

#### Success
[Search link](https://github.com/paypal/paypal-smart-payment-buttons/pull/558/files#diff-84bf1f3b16b25fd5ae8707244a976f19382c981728d3e234d78f25bf57cf3446R82)
Screenshot 2023-03-13 at 4 22 54 PM

https://github.com/paypal/paypal-checkout-components/pull/2138

❤️  Thank you!


src/card/constants.js
```diff
@@ -310,5 +310,10 @@ export const FPTI_HCF_KEYS = {
   HCF_SESSION_ID: ("hcf_session_id": "hcf_session_id"),
   HCF_CORRELATION_ID: ("hcf_correlation_id": "hcf_correlation_id"),
   SDK_CORRELATION_ID: ("sdk_correlation_id": "sdk_correlation_id"),
-  VAULT_TOKEN: ("vault_token": "vault_token")
+  VAULT_TOKEN: ("vault_token": "vault_token"),
+  HCF_ORDER_ID: ('hcf_order_id' : 'hcf_order_id'),
```
wsbrunson, line 314: we should use `order_id` because it is an existing tag. We'd have to create a new tag for `hcf_order_id`

src/card/props.js
```diff
@@ -272,6 +272,8 @@ export function getCardProps({
       branded,
       style,
       placeholder,
+      onApprove: xprops.onApprove,
+      createOrder: xprops.createOrder,
```
siddy2181, line 276: @wsbrunson createOrder is what overrides the CONTEXT_TYPE. Can we bypass legacy props for [createOrder](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/createOrder.js#L163)?

wsbrunson: good find! yes, i would bypass this

src/card/interface/submitCardFields.js
```diff
@@ -63,49 +62,47 @@ export function submitCardFields({
...
+          payment_source: {
+            card: cardObject,
+          },
+        };
+        return confirmOrderAPI(orderID, data, {
+          facilitatorAccessToken,
+          partnerAttributionID: ""
+        })
+        .then(() => {
+          hcfTransactionSuccess({ orderID });
```
wsbrunson, line 92: We'll want to do this after the `onApprove` call, like we do here with vault-without-purchase: https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/card/interface/vault-without-purchase.js#L60

src/card/interface/submitCardFields.js
```diff
@@ -63,49 +62,47 @@ export function submitCardFields({
...
-          // $FlowFixMe
-          return cardProps.onApprove(
-            { payerID: uniqueID(), buyerAccessToken: uniqueID(), ...orderData },
-            {
-              restart: () => {
-                throw new Error(`Restart not implemented for card fields flow`);
-              },
-            }
-          );
+        .catch(error => {
```
wsbrunson, line 95: we'll also want a `catch` for the `createOrder` call on line 67 in case the merchant passes us a bad callback or it fails to create an order

src/api/order.js
```diff
@@ -354,14 +354,12 @@ export function confirmOrderAPI(orderID : string, data : ConfirmData, { facilita
...
         eventName:   'order_confirm_payment_source',
-        url:         `${ ORDERS_API_URL }/${ orderID }/confirm-payment-source`,
+        url:         `${ ORDERS_API_URL }/${ orderID}/confirm-payment-source`,
         data,
         headers:     {
             [HEADERS.PARTNER_ATTRIBUTION_ID]: partnerAttributionID || '',
             [HEADERS.PREFER]:                 PREFER.REPRESENTATION
         }
-    }).then(({ data: orderData }) => {
-        return orderData;
```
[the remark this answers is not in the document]
bywood: The `callRestAPI` call returns the response body. It doesn't add the body to a data property. Looking at other places where `confirmOrderAPI` is called, I don't see them using the returned data, so I don't think this returning `undefined` would have caused an issue before. The `confirm-payment-source` API isn't returning a body with a `data` property for our usage and I"m not sure why it would be different for SPB.

src/card/interface/submitCardFields.js
bywood: If we do it after the `onApprove`, how do we know if it was successful or not? We'd have to rely on the merchant to throw an error or return a Promise, but those aren't guarantees.

src/card/interface/submitCardFields.js
```diff
@@ -52,42 +51,44 @@ export function submitCardFields({
...
-          }
+    return cardProps
+      .createOrder()
+      .then((id) => {
 
-          return confirmOrderAPI(orderID, data, {
-            facilitatorAccessToken,
-            partnerAttributionID: "",
-          }).catch((error) => {
-            getLogger().info("card_fields_payment_failed");
```
siddy2181, line 74: Do we need this catch block and log here? we have a [catch block](https://github.com/paypal/paypal-smart-payment-buttons/pull/558/files#diff-84bf1f3b16b25fd5ae8707244a976f19382c981728d3e234d78f25bf57cf3446R82) that would capture this event as `hcf_transaction_error` .

src/card/lib/card-focus.test.js
siddy2181, line 1: I ran `npm run lint` locally and these changes happened.

src/lib/logger.test.js
siddy2181, line 1: Same for this file. 
I ran `npm run lint` locally and these changes happened.

src/lib/register-service-worker.test.js
siddy2181, line 1: Same for this file

wsbrunson (approved):

PR #567, opened 2023-03-07
Tests refactor

oscarleonnogales:

❤️  Thank you!


src/card/lib/card-utils.js
```diff
@@ -319,6 +320,61 @@ export function getContext(win : Object) : string {
     return win.xprops?.parent?.uid || win.xprops?.uid;
 }
 
+export function cardExpiryToPaymentSourceExpiry(dateString: string): string {
```
oscarleonnogales, line 323: Moved these two functions and their tests out of the `index.js` file. Other `index.js` files don't have any logic in them.

src/card/interface/gql.test.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+import { describe, it, expect } from "vitest";
+
```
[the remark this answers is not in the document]
wsbrunson: I'd say delete it now, otherwise you'll be looking at this file in 3 years wondering why its still here 😂

src/card/interface/reformatExpiry.js
```diff
@@ -1,9 +0,0 @@
-/* @flow */
-
-// Reformat MM/YYYY to YYYY-MM
-export function reformatExpiry(expiry: ?string): ?string {
```
wsbrunson, line 4: Are we not using this anywhere?

wsbrunson (commented):
just one question otherwise this looks great

wsbrunson (approved):

PR #569, opened 2023-03-08
preventing onSubmit if cardholder name is invalid

oscarleonnogales:
### Description

Preventing `submitCardFields()` to complete if the cardholder name is invalid. This change is for all optional fields. The expected behaviour is to prevent any API calls to happen if we know there's a mistake on the forms, and log them for the merchant.

https://paypal.atlassian.net/browse/DTCURIOSTY-8477
https://paypal.atlassian.net/browse/DTNOR-863

Render the cardholderName field, and fill it out using only digits. The payment should not be going through, but it's currently ignoring this field as it's an optional field. This fix will prevent the flow from continuing.

❤️  Thank you!


wsbrunson (approved):
great work!

PR #572, opened 2023-03-14
LI-3650: Add logging for invalid resource ID  

spencersablan:
### Description

Added logging for when there is an invalid resource ID (clientID mismatch).  
This will eventually turn into an error, but we want to check the impact of this before we implement the error.

**JIRA issue:** https://paypal.atlassian.net/browse/LI-3650
A merchant noticed they were able to load the SDK with one client ID, but create an order with another.  This is due to the Smart API being able to process the order with an invalid client ID.  So hopefully next week, this code will turn into a check before we call the Smart API, that will catch any requests with an invalid client ID & throw an error instead of calling the Smart API.

For now, this is just to add logging to see impact.

@gregjopa @xuanwan05 @ravishekhar @dtjones404 @elizabethmv @wsbrunson 

❤️  Thank you!


spencersablan:
The logging feels a bit redundant right now, but it will be easier to find out how often the `INVALID_RESOURCE_ID` error is hit.

wsbrunson (approved):

westeezy (approved):

wsbrunson (approved):

PR #576, opened 2023-03-15
remove unneeded get setup token call

wsbrunson:
`GET /v3/vault/setup-tokens/:id` needs special permission to call, and we don't need to use it on the sdk anyway. We can just use `update` and check contingencies that happen from the update call

EDIT:
We are also adding back the passing of the id token to the graphql update token endpoint.

We don't think its required for the vault without purchase flow but we still need to pass it until we can deploy xobuyernodeserv with a fix to make it optional.

### DIFF
--- a/src/card/interface/vault-without-purchase.js
+++ b/src/card/interface/vault-without-purchase.js
@@ -33,6 +33,7 @@ type VaultPaymenSourceOptions = {|
   onError: XOnError,
   clientID: string,
   paymentSource: PaymentSourceInput,
+  idToken: string,
 |};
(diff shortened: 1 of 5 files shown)


src/card/interface/vault-without-purchase.js
wsbrunson, line 1: this file had changes that weren't already run through `prettier` hence why it looks like there were a lot of changes.

the only actual change starts on line 47

gregjopa (approved):

elizabethmv (approved):

PR #579, opened 2023-03-20
chore: no longer marking ineligible cards as invalid

oscarleonnogales:
### Description

When the merchant account is ineligible for certain card types(Discover,Amex),the form is able to submit successfully even though the field is marked as invalid.
The orders API team confirmed that the payment would decline on their end after the call is made.

We'll no longer mark ineligible cards as invalid and will allow the call to orders API to go through. The response will be an error and we can surface this to the merchant so they can decide and handle what to do on their site when a buyer attempts to make a purchase with an ineligible card type.

Additionally, when we detect that a card is ineligible for either the purchase or vaulting flow, we'll throw an error for the merchant to know and handle any UI changes on their end.

https://paypal.atlassian.net/browse/DTCURIOSTY-8609

Type in a Discover card number in the card number field. If the merchant does not have Discover enabled as a part of their funding eligibility then it will be marked as invalid, but still allow the form to submit.

Screenshot 2023-03-17 at 15 20 20

❤️  Thank you!

### DIFF
--- a/src/card/lib/card-checks.test.js
+++ b/src/card/lib/card-checks.test.js
@@ -138,7 +138,7 @@ describe("card-checks", () => {
       window.xprops = {};
     });
 
-    it("should find the card eligible when the vendor is eligible", () => {
+    it("should find the card eligible when the vendor is eligible for purchaese flow", () => {
       window.xprops.fundingEligibility = {
         card: {
           eligible: true,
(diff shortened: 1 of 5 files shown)


wsbrunson (approved):

src/card/lib/card-checks.test.js
```diff
@@ -178,7 +222,9 @@ describe("card-checks", () => {
       };
       const cardNumber = "4111111111111111";
       const cardType = detectCardType(cardNumber)[0];
-      expect(checkCardEligibility(cardNumber, cardType)).toBe(false);
+      expect.assertions(2);
```
wsbrunson, line 225: you don't really need these assertions unless you are working with promises/async code. 

with async/promises, the test can exit before executing all of your `expect`s and give you a false positive. For synchronous code, that should never happen

PR #580, opened 2023-03-22
URL sanitization and alphanumeric check for service worker integration

Simmba:
### Description
Changes requested for a previously submitted feature

PR's
https://github.com/paypal/paypal-smart-payment-buttons/pull/382
Requests
Move asset load to button click from button render. -> Needs a new PR
Add validation for URL sanity and alphanumeric check in SPB - [Code Link](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/lib/register-service-worker.js#L102) -> Needs a new PR

❤️  Thank you!


src/lib/register-service-worker.js
```diff
@@ -98,14 +98,25 @@ const startRegistration = (swUrl) => {
         });
 }
 
+export function sanitizedUrlGenerator (releaseHash: string, serviceWorker: string): string {
```
gregjopa, line 101: nit: could we prefix this with `get` to make it clear that it returns a string?
```suggestion
export function getSanitizedUrl (releaseHash: string, serviceWorker: string): string {
```

PR #582, opened 2023-03-23
separated submitCardFields into 2 flows

oscarleonnogales:
### Description

Splitting up the purchase and vault-without-purchase flow in the `submitCardFields()` function. The `getCardProps()` method will validate that only 1 of either `createVaultSetupToken` or `createOrder` methods is being passed in.

❤️  Thank you!

### DIFF
--- a/src/card/interface/submitCardFields.js
+++ b/src/card/interface/submitCardFields.js
@@ -2,12 +2,13 @@
 
 import { ZalgoPromise } from "@krakenjs/zalgo-promise/src"
 
-import { getCardProps } from "../props"
+import { getCardProps, type SaveCardFieldsProps, type LegacyCardProps } from "../props"
 import { confirmOrderAPI } from "../../api"
 import { hcfTransactionError, hcfTransactionSuccess } from "../logger"
 import type { FeatureFlags } from "../../types"
-import type { BillingAddress } from '../types'
+import type { BillingAddress, Card, ExtraFields } from '../types'
 import {convertCardToPaymentSource, reformatPaymentSource} from '../lib'
+import { SUBMIT_ERRORS } from "../constants"
 
 import { resetGQLErrors } from "./gql"
 import { hasCardFields } from "./hasCardFields"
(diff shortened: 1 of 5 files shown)


bywood (commented):
I like the changes. Just one suggestion.

src/card/interface/submitCardFields.js
```diff
@@ -32,70 +87,23 @@ export function submitCardFields({
     featureFlags,
   });
 
+  // $FlowIssue
+  const [isPurchaseFlow, isVaultFlow] = [Boolean(cardProps.createOrder), Boolean(cardProps.createVaultSetupToken)];
```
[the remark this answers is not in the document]
oscarleonnogales: I originally had it that way, but both lines were throwing the same type of flow issue. Do you suggest splitting it up again and commenting `FlowIssue` above both lines? Or is okay to keep it this way for a case like this?

elizabethmv, line 91:
```suggestion
  const [isPurchaseFlow, isWithoutPurchaseFlow] = [Boolean(cardProps.createOrder), Boolean(cardProps.createVaultSetupToken)];
```

Nit: I think the `isVaultFlow` naming could be a little misleading - the `isPurchaseFlow` supports a one time checkout and also a vault with purchase. One idea is `isWithoutPurchase` curious on thoughts from others cc: @bywood @wsbrunson @wickerpopstar

wsbrunson: yeah, specifically `isVaultWithoutPurchase` because there are also other non-purchase flows

src/card/props.js
```diff
@@ -159,11 +159,11 @@ function validateVaultWithoutPurchaseSetup(xprops, baseProps): {| createVaultSet
 
   // this if check is really for Flow so that later it knows this function is defined
   if (!xprops?.createVaultSetupToken) {
-      throw new Error("createVaultSetupToken is required when saving card fields");
+      throw new Error(SUBMIT_ERRORS.MISSING_CREATE_VAULT_SETUP);
```
wsbrunson, line 162: let me know if you have time to talk today (Monday 27 march). There are more changes we need to make to this file

src/card/interface/submitCardFields.js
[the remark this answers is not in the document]
bywood: @oscarleonnogales What's the Flow issue? If it doesn't like the `Boolean` function, then maybe something like:
```js
const isPurchaseFlow = (typeof cardProps.createOrder === "function");
const isVaultFlow = (typeof cardProps.createVaultSetupToken === "function");
```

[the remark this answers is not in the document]
oscarleonnogales:
```
Cannot get cardProps.createOrder because property createOrder is missing in SaveCardFieldsProps [prop-missing]
Cannot get cardProps.createVaultSetupToken because property createVaultSetupToken is missing in LegacyCardProps [prop-missing]
```

`createOrder` should only be defined in `LegacyCardProps`, and `createVaultSetupToken` should only be defined in `SaveCardFieldsProps`

src/card/props.js
```diff
@@ -223,63 +184,51 @@ export function getCardProps({
...
+    createBillingAgreement: xprops.createBillingAgreement,
+  }) : null;
+
+  // $FlowFixMe
+  return {
+    ...baseProps,
+    ...props,
+    ...returnData,
+    // $FlowIssue
+    createVaultSetupToken: createVaultSetupToken ? getCreateVaultSetupToken({ createVaultSetupToken: xprops.createVaultSetupToken }) : null,
```
wsbrunson, line 228:
```suggestion
    createVaultSetupToken: xprops.createVaultSetupToken,
    createOrder: xprops.createOrder,
```

src/card/props.js
```diff
@@ -223,63 +184,51 @@ export function getCardProps({
...
+    // $FlowIssue
+    onApprove: createOrder ? props.onApprove : getSaveActionOnApprove({ onApprove: xprops.onApprove, onError: baseProps.onError }),
```
wsbrunson, line 230:
```suggestion
    onApprove: xprops.onApprove,
```

wsbrunson (approved):

wsbrunson (approved):

PR #587, opened 2023-03-29
add feature flag check to hosted card fields

wsbrunson:
the `experiments` type is duplicated. We'll come back and figure out how to share that. There's two types, one for paypal and one for card fields.


src/card/types.js
```diff
@@ -13,6 +13,9 @@ export type SetupCardOptions = {|
     metadata: {|
         correlationID: string,
         spbVersion: string
+    |},
+    experiments: {|
```
siddy2181, line 17: should we make this field optional?

src/card/props.js
```diff
@@ -136,12 +136,20 @@ export type CardProps = LegacyCardProps | SaveCardFieldsProps
 type GetCardPropsOptions = {|
   facilitatorAccessToken: string,
   featureFlags: FeatureFlags,
+  experiments: {
```
siddy2181, line 139: same here?

src/card/interface/submitCardFields.js
```diff
@@ -21,6 +21,9 @@ type SubmitCardFieldsOptions = {|
   extraFields?: {|
     billingAddress?: BillingAddress,
   |},
+  experiments: {|
```
siddy2181, line 24: and here

src/card/components/page.jsx
```diff
@@ -18,10 +18,13 @@ import { CardField, CardNumberField, CardCVVField, CardExpiryField, CardNameFiel
 type PageProps = {|
     cspNonce : string,
     props : CardProps,
-    featureFlags: FeatureFlags
+    featureFlags: FeatureFlags,
+    experiments: {|
```
siddy2181, line 22: should we make this field optional?
`experiments?: ?{|...`

siddy2181 (approved):

PR #595, opened 2023-04-17
fix: apply feature flag cco blocking check to checkout flow

gregjopa:
### Description

Follow up to #594. This PR implements the feature flag into the Checkout flow which calls the CCO mutation when eligible.


ravishekhar (approved):
💯

PR #597, opened 2023-04-20
chore: Remove idToken for hcf

siddy2181:
### Description

Removes IdToken requirement for hosted card fields

Previously we wanted to make ID token required to use this flow, however upon implementation we decided that we only want ID token when we need buyer identity passed in the SDK and don't need to make it required on the first time for some of our more complex partner flows.
At this time the SDK does not support return buyer experience for the new HCFs so it makes sense for us to remove it altogether until it is needed for any return buyer.

https://github.paypal.com/Checkout-R/xobuyernodeserv/pull/3181
https://github.paypal.com/Checkout-R/xobuyernodeserv/pull/3125

❤️  Thank you!


elizabethmv (approved):

gregjopa:
@siddy2181 @elizabethmv is this PR blocked until the graphql contract is updated? Or are we good to merge?

elizabethmv:
> @siddy2181 @elizabethmv is this PR blocked until the graphql contract is updated? Or are we good to merge?

@gregjopa this PR is good to go! The GQL changes have been released for some time now.

PR #598, opened 2023-04-21
fix: opt out of native flow for blocking cco call

gregjopa:
### Description

The native app switch flow is not designed to support a blocking updateClientConfig() call. This PR makes the native flow ineligible when `featureFlags.isButtonClientConfigCallBlocking` is set to true.


ravishekhar (approved):
lgtm 👍

gregjopa:
Closing this out without merging. We have met with internal teams and decided against this approach.

PR #600, opened 2023-04-24
Add logging for updateButtonConfigClient

dtjones404:
### Description

Added additional logs for new feature flag behavior.

Screenshot 2023-04-24 at 2 45 43 PM

❤️  Thank you!


src/api/order.js
```diff
@@ -566,10 +567,15 @@ type ClientConfig = {|
...
     productFlow : string,
-    buttonSessionID : ?string
+    buttonSessionID : ?string,
+    featureFlags?: FeatureFlags
 |};
 
-export function updateClientConfig({ orderID, fundingSource, integrationArtifact, userExperienceFlow, productFlow, buttonSessionID } : ClientConfig) : ZalgoPromise<void> {
+export function updateClientConfig({ orderID, fundingSource, integrationArtifact, userExperienceFlow, productFlow, buttonSessionID, featureFlags } : ClientConfig) : ZalgoPromise<void> {
+    if (featureFlags && featureFlags.isButtonClientConfigCallBlocking) {
+      getLogger().info('update_button_client_config_called_as_blocking', {time: String(new Date().getTime()), ecToken: orderID, buttonSessionID, fundingSource})
```
wsbrunson, line 576:
```
time: String(new Date().getTime())
```

that should already be included by our log ingestion tool

src/button/pay.js
```diff
@@ -153,14 +153,16 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
...
+                    return updateButtonClientConfig({ orderID, fundingSource, inline, userExperienceFlow, featureFlags }).catch(err => {
                         getLogger().error('update_client_config_error', { err: stringifyError(err) });
                     });
                 }
 
                 // feature flag to control blocking/non-blocking behavior
                 if (featureFlags.isButtonClientConfigCallBlocking) {
-                    return updateButtonClientConfigWrapper();
+                    return updateButtonClientConfigWrapper().then(() => {
+                        getLogger().info('update_button_client_config_resolved', {time: String(new Date().getTime()), fundingSource, ecToken: orderID, buttonSessionID})
```
wsbrunson, line 164: same comment about `time`

src/button/pay.js
```diff
@@ -221,7 +226,11 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 return ZalgoPromise.try(close).then(() => {
                     throw err;
                 });
-            }).then(noop);
+            }).then((noop)).finally(() => {
+              if (featureFlags.isButtonClientConfigCallBlocking) {
+                  getLogger().info('redirect_to_xorouter', {time: String(new Date().getTime()), fundingSource, ecToken: confirmedOrderID, buttonSessionID})
```
wsbrunson, line 231: same comment about `time`

should `confirmedOrderID` be `orderID` for the sake of consistency? We could include a prop like `status: 'confirmed'`

src/api/order.js
dtjones404: It is, but I think there was some doubt about whether the existing timestamps are created when the log is created, or when the logger gets flushed.  I thought it might make sense to add these additional timestamps to get some visibility about that.

src/button/pay.js
dtjones404: Yeah this is kinda awkward.  I wanted to include orderID in this log but it wasn't available inside this scope before.  I added lines 197 and 200 as a (kinda hacky) workaround.  I didn't want to reuse the name `orderID` because I was worried about variable shadowing so I chose `confirmedOrderID`.  Definitely open to a better solution here!

wsbrunson: the best thing would be to use `async/await` but since we cant, i think this is okay

wsbrunson (approved):

src/button/pay.js
```diff
@@ -192,7 +194,10 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 });
             });
 
+            let confirmedOrderID;
+
             const confirmOrderPromise = createOrder().then((orderID) => {
+                confirmedOrderID = orderID
```
gregjopa, line 200:
```suggestion
                confirmedOrderID = orderID;
```

src/button/pay.js
```diff
@@ -153,14 +153,16 @@ export function initiatePaymentFlow({ payment, serviceData, config, components,
                 }
 
                 function updateButtonClientConfigWrapper() : ZalgoPromise<void> {
-                    return updateButtonClientConfig({ orderID, fundingSource, inline, userExperienceFlow }).catch(err => {
+                    return updateButtonClientConfig({ orderID, fundingSource, inline, userExperienceFlow, featureFlags }).catch(err => {
```
gregjopa, line 156: Can you also pass `featureFlags` into this updateButtonClientConfig() call? https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/payment-flows/checkout.js#L471

gregjopa (approved):
Looks great! 💯 

@ravishekhar please review when you get a chance.

src/api/order.js
ravishekhar: The timestamps recorded by logger pool are actually server side timestamps and its getting impacted by the order received in server side rather than actually created.

PR #603, opened 2023-04-26
Fix validation on payee for Vault without Purchase

elizabethmv:
### Description
In a vault without purchase transaction there is currently a validation error being thrown in a partner scenario if a merchant ID is passed to the SDK.

The SDK receives a setup token and exchanges that for an order ID. It will then validate that order ID and is checking to make sure the payees match. We are removing this validation error specifically for `createVaultSetupToken`.

We don't need to validate this on a vault without purchase transaction though, as there is no amount or way to set the payee in the current integration. The payee will be set in the order once there is a subsequent transaction.

Another option is for the partner to not pass a merchant ID to the SDK, however, if the partner needs to pass it for other eligibility checks (MCC, for example) they would be seeing this error.

The payee validation will still be present on the other checkout scenarios.

Before:
- Local setup with the vault without purchase flow and the main branch
- Pass a merchant ID in the SDK query params and a partner client ID
- Click button
- Observe error

After:
- Local setup with the vault without purchase flow and this branch
- Pass a merchant ID in the SDK query params and a partner client ID
- Click button
- Observe no error

The error before & the error that will remain for a checkout scenario:
Screen Shot 2023-04-26 at 1 47 01 PM

N/A

N/A


wsbrunson (approved):

elizabethmv:
The `validation` test file has some tests that are skipped due to flakiness. I'm seeing the test I added behaving flaky as well. I'm removing it for now and will send a new PR for a test. Manual testing looked great in the meantime!

spencersablan (approved):

gregjopa (approved):

gregjopa:
This change looks good. Let's make a note to revisit this `src/button/validation.js` file in the future and clean up the conditional logic. I'm thinking we can get rid of a lot of the `else` statements and it out by doing `return triggerIntegrationError()` (assuming we only want one error/warning returned back). And/or break out the complex validation parts into their own functions.

PR #604, opened 2023-05-01
Ubcc logging

dtjones404:
### Description

Some tweaks to the logging added with #600.

* Removed ecToken properties, as our logger already adds these.
* Moved `resolved` log to capture missing payment flows
* Moved `redirect` log from `finally` to `then` to avoid logging on failed calls
* Added logs for the non-blocking use-case as well

❤️  Thank you!


gregjopa (approved):

gregjopa:
@ravishekhar please review when you get a chance.

PR #606, opened 2023-05-02
pass paymentSource to createVaultSetupToken

wsbrunson:
Currently `createVaultSetupToken` only supports PayPal. In the future, this callback will be supported in other payment methods so we want to make sure we support other payment methods like the `createOrder` callback does.

Also did some cleanup with the types of `XCreateVaultSetupToken` and `CreateVaultSetupToken`


src/card/interface/vault-without-purchase.js
```diff
@@ -33,7 +34,7 @@ export const savePaymentSource = ({
   paymentSource,
 }: VaultPaymenSourceOptions): ZalgoPromise<void> => {
   let vaultToken;
-  return createVaultSetupToken()
+  return createVaultSetupToken({ paymentSource: FUNDING.CARD })
```
wsbrunson, line 37: This is another instance where sharing props between spb and card fields makes the code confusing. `createVaultSetupToken` for CardFields should really be its own thing, there is not concept of multiple payment sources for card fields.

elizabethmv: It makes sense for buttons, but does it make sense to set the `paymentSource` for CardFields? The merchant would know to set the payment source as `card` in the request payload. 

On the other had the consistency is always nice. 🤷‍♀️

elizabethmv (approved):
Lgtm! One question on payment source for card fields

src/card/interface/vault-without-purchase.js
wsbrunson: I guess lets get rid of it now, I don't think we want it for CardFields and it could be a bigger headache to go back and remove it

elizabethmv (approved):

PR #607, opened 2023-05-03
log product action on render

wsbrunson:
This started as making the "payment_flow" available for FPTI events on render but grew into something a bit bigger.

I added a new card prop called `productAction` that will represent what action the card fields component should take. For now, these actions map directly to `with_purchase` and `vault_with_purchase` because that's what analytics want us to track. We could change these names later as long we we set up some type of map to send the right tag name to FPTI. The constant is called `PAYMENT_FLOWS` right now because it maps directly to FPTI but we can change it later.

I spent awhile thinking of the name `productAction`. `flow`, `payment_flow`, and `product_flow` are all taken. `buyerFlow` makes it sound like a purchase is happening. `action` by itself is a little overloaded after our architecture pivot from standalone actions. If anyone has suggestions or thoughts I'd love to hear them.

I also made sure to update some places where we were checking for createOrder or createVaultSetupToken and replace them with checks for `productAction`. There's a larger refactor we could do when we have time


siddy2181 (approved):

oscarleonnogales:
I like it! Looks like it'll be a lot easier to add future payment flows with these changes as well.

oscarleonnogales (approved):

Wanlc888 (approved):

siddy2181 (approved):

PR #608, opened 2023-05-09
update props for card-fields; memoize create order

wsbrunson:
* memoize createOrder and createVaultSetupToken
* move legacy props to callback props and stop using them in CardFields
* re-work some of the types associated with createOrder and createVaultSetupToken


src/card/props/createOrder.js
```diff
@@ -0,0 +1,43 @@
...
+}: CreateOrderXProps): ?CreateOrder {
+  if (!createOrder) {
+    return;
+  }
+
+  return memoize(() => {
+    const queryOrderID = getQueryParam("orderID");
+    if (queryOrderID) {
+      return ZalgoPromise.resolve(queryOrderID);
+    }
```
wsbrunson, line 25: we should check the dev docs. if we don't document this query param we should remove support for it

wickerpopstar:
General question: what's the plan for the the `$FlowFixMe` issues? Are we planning on investing any time to resolve those, or just wait until we can convert this to TS and resolve things then?

wickerpopstar (approved):

wsbrunson:
> General question: what's the plan for the the $FlowFixMe issues? Are we planning on investing any time to resolve those, or just wait until we can convert this to TS and resolve things then?

I think waiting for the move to TS. Some of these types need an overhaul (and the code itself is probably in need of a refactor) and it seems like just waiting until we are on TS would be better. IDK though, maybe i'm just thinking all our problems will be solved by TS which, out loud, sounds like magically thinking

PR #613, opened 2023-05-19
log metrics and fpti data on click and create order errors

wsbrunson:


src/button/button.js
```diff
@@ -164,14 +163,9 @@ export function setupButton({
...
-            getLogger()
-                .info('smart_buttons_payment_error', { err: stringifyError(err) })
-                .track({
-                    [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
-                    [FPTI_KEY.ERROR_CODE]: 'smart_buttons_payment_error',
-                    [FPTI_KEY.ERROR_DESC]: stringifyErrorMessage(err)
-                });
+            // we don't think this code path is actually hit but we
+            // are too afraid to remove it right now
+            getLogger().info('smart_buttons_payment_error', { err: stringifyError(err) });
```
cgdibble, line 168: 🤔 Can we add a log prefix or something to this to be able to detect this code path in logs over time? Or is this already it's own error?

src/button/button.js
```diff
@@ -164,14 +163,9 @@ export function setupButton({
...
-            
-            getLogger()
-                .info('smart_buttons_payment_error', { err: stringifyError(err) })
-                .track({
-                    [FPTI_KEY.STATE]:      FPTI_STATE.BUTTON,
-                    [FPTI_KEY.ERROR_CODE]: 'smart_buttons_payment_error',
-                    [FPTI_KEY.ERROR_DESC]: stringifyErrorMessage(err)
-                });
+            // we don't think this code path is actually hit but we
+            // are too afraid to remove it right now
```
wickerpopstar, line 167: Would it be helpful to add a comment indicating what we need to feel comfortable to remove it? Conditions, date, what have you?

wickerpopstar (approved):

src/button/button.js
```diff
@@ -268,7 +276,14 @@ export function setupButton({
             const { onError } = paymentProps;
 
             payPromise.catch(err => {
-                getLogger().error('prerender_initiate_payment_reject', { err: stringifyError(err) }).flush();
+                getLogger()
```
dtjones404, line 279: Would it be worth sending another metric here?  Or possibly the same metric with a new dimension representing whether or not the button is in prerender?

src/props/createOrder.js
```diff
@@ -177,6 +177,10 @@ type CreateOrderProps = {|
 export function getCreateOrder({ createOrder, intent, currency, merchantID, partnerAttributionID, paymentSource } : CreateOrderXProps, { facilitatorAccessToken, createBillingAgreement, createSubscription, enableOrdersApprovalSmartWallet, smartWalletOrderID, createVaultSetupToken, flow } : CreateOrderProps) : CreateOrder {
     const data = buildXCreateOrderData({ paymentSource });
     const actions = buildXCreateOrderActions({ facilitatorAccessToken, intent, currency, merchantID, partnerAttributionID });
+    // this is purely for analytics purposes. We'd like to know whether
+    // create order was called by a client side integration (actions.order)
+    // or a server-side (all other callbacks only allow server-side)
```
cgdibble, line 182: Clarifying my understanding on this fine Monday morning:
All the other callbacks involve making a request to the merchant server in order to make the call, correct? Where as `actions.order` is the only path to a clientside integration?

cgdibble (approved):
Just a couple questions/comments otherwise all good

src/button/button.js
wsbrunson: this is already its own error and i think ideally we'd just delete this once we can verify that our metrics are doing a better job here

src/props/createOrder.js
wsbrunson: you are spot on

PR #618, opened 2023-06-07
Deprecate VPM in vault capture flow

elizabethmv:
### Description
This PR adds an experiment to the vault capture flow to slowly deprecate the use of the validate payment method endpoint. `vault-capture` handles quite a few flows, so we want to ramp it slowly before completely removing it.

Validate payment method in this flow was also handling 3DS, which we had to make some updates to.

We want to only use the confirm payment source endpoint.

Reach out internally for test cases.

N/A

GraphQL changes will be coming as well, but can go out independently.

N/A


wsbrunson:
@elizabethmv do we still need this PR?

elizabethmv:
> @elizabethmv do we still need this PR?

We can leave this a branch. Thanks!

PR #621, opened 2023-06-22
chore: add log for render version mismatch

dtjones404:
### Description

* Added warning log when SDK version used for first render doesn't match SDK version used for server-side render.

To increase visibility of fallback rendering behaviors in SCNW

Screenshot 2023-06-21 at 8 28 55 PM

❤️  Thank you!

### DIFF
--- a/src/button/logger.js
+++ b/src/button/logger.js
@@ -131,15 +131,15 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             native_device = 'android_chrome';
         }
 
-        const renderVersion = getTemplateVersion();
-        const dotSeparatedRenderVersion = renderVersion.split('_').join('.');
+        const serverRenderVersion = getTemplateVersion();
+        const dotSeparatedRenderVersion = serverRenderVersion.split('_').join('.');
 
         if (dotSeparatedRenderVersion !== sdkVersion) {
-          logger.warn('server_render_version_mismatch', {sdkVersion, renderVersion})
+          logger.warn('server_render_version_mismatch', {sdkVersion, serverRenderVersion})
         }
 
         logger.info(`button_render`);
-        logger.info(`button_render_template_version_${ renderVersion }`);
+        logger.info(`button_render_template_version_${ serverRenderVersion }`);
         logger.info(`button_render_client_version_${ getClientVersion() }`);
         logger.info(`button_render_color_${ color }`);
         logger.info(`button_render_shape_${ shape }`);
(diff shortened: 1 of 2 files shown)


src/button/logger.js
```diff
@@ -131,8 +131,15 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             native_device = 'android_chrome';
         }
 
+        const renderVersion = getTemplateVersion();
+        const dotSeparatedRenderVersion = renderVersion.split('_').join('.');
+
+        if (dotSeparatedRenderVersion !== sdkVersion) {
+          logger.warn('server_render_version_mismatch', {sdkVersion, renderVersion})
```
wsbrunson, line 138: this would be a great metric to send too

wsbrunson (approved):

src/button/logger.js
```diff
@@ -131,8 +131,15 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             native_device = 'android_chrome';
         }
 
+        const renderVersion = getTemplateVersion();
```
gregjopa, line 134: Wondering if we should be as explicit as possible with our variable names that this is the version of the JS SDK used to server render the buttons.
```suggestion
        const serverRenderVersion = getTemplateVersion();
```

dtjones404: Good call!  I like that much better

src/button/logger.js
dtjones404: Good idea!  Just added one, hopefully correctly 😅

src/button/logger.js
```diff
@@ -131,8 +132,19 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
...
         }
 
+        const serverRenderVersion = getTemplateVersion();
+        const dotSeparatedRenderVersion = serverRenderVersion.split('_').join('.');
+
+        if (dotSeparatedRenderVersion !== sdkVersion) {
+          logger.warn('server_render_version_mismatch', {sdkVersion, serverRenderVersion})
+          sendMetric({
+            name: 'pp.app.paypal_sdk.buttons.server_render_version_mismatch',
+            dimensions: {sdkVersion, serverRenderVersion}
```
wsbrunson, line 142: we may want to omit this for now, the cardinality of these fields could be too high

dtjones404: Ah good point.  Should we just drop the dimensions object entirely then?

gregjopa (approved):
💯

PR #626, opened 2023-06-26
[DTPPSDK-1266]Add disableSetCookie flag to fb.js call

xuanwan05:
### Description
Add disableSetCookie flag to fb.js call

Screenshot 2023-06-26 at 12 43 04 PM

❤️  Thank you!


wsbrunson (commented):
similar to what @gregjopa is suggesting here: https://github.com/paypal/paypal-smart-payment-buttons/pull/625/files#r1242561921

we should modify `loadFraudnet` to accept a new property, something like `fraudnetUrlParams` and in there we would include 

```
{
  disableSetCookie: true
}
```

PR #632, opened 2023-07-03
chore: make render version mismatch log info in prod

dtjones404:
### Description

* set log level for `server_render_version_mismatch` based on environment- `info` in production and `warn` otherwise

❤️  Thank you!


src/button/logger.js
```diff
@@ -136,7 +136,8 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
         const dotSeparatedRenderVersion = serverRenderVersion.split('_').join('.');
 
         if (dotSeparatedRenderVersion !== sdkVersion) {
-          logger.warn('server_render_version_mismatch', {sdkVersion, serverRenderVersion})
+          const logLevel = env === "production" ? "info" : "warn";
+          logger[logLevel]('server_render_version_mismatch', {sdkVersion, serverRenderVersion})
```
gregjopa, line 140: What do you think about just using `info` level for all environments?

dtjones404: Probably a better idea,  I'm just worried that an info-level log might not stand out during stage testing

PR #636, opened 2023-07-07
Use vitest for menu tests

jshawl:
### Description

This PR replaces the menu karma test with vitest.

The existing karma tests use `wrapPromise` and `expect` and `avoid` which is imported from the internal [kraken/belter](https://github.com/krakenjs/belter/blob/6b054d83d2be457dd5c652549ee9f3c4a5a4b42d/src/test.js#L24) repo.

Using vitest will make it easier for other contributors to write tests using publicly documented methods and patterns.

While migrating the test, I noticed a few things:

1. some of the `expect` and `avoid`s were not being called correctly and silently passing. 
  - I removed these from the tests
3. we did not have happy path coverage for opening a popup from the menu
  - I added coverage for the happy path 😺 

```
npx vitest menu
```

### Reviewer Notes

I realized we are relying on karma for the test coverage % report so this change will make it look like we have no coverage for the menu now :( 

❤️  Thank you!


src/menu/menu.test.js
```diff
@@ -0,0 +1,62 @@
+/* @flow */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const warn = vi.fn();
+const openPopup = vi.fn();
+const onSelect = vi.fn();
+
+vi.mock("../lib", async () => {
```
westeezy, line 8: I wonder how we can start sharing this as I'm sure we will need to mock this in more places. Maybe we can explore that in a future unit of work

westeezy (approved):
nice job!

dtjones404 (approved):

PR #645, opened 2023-07-25
track button click success count

jshawl:
### Description

📈 This pr increments `pp.app.paypal_sdk.buttons.click.success.count` right before the merchant's `onApprove` callback is invoked.

🚨 We would like to be able to set up alerts for when the success count drops significantly and be able to identify which payment flow is having trouble.

```
npm run vitest onApprove
```

☝️ i added test coverage for the `getOnApproveOrder` implementation. I started to add other flows to these tests but realized that the tests would be tightly coupled to the lower level implementation details, and while that increases code coverage, does not leave a lot of room for refactoring in the future, so I decided to leave it out. I'm happy to add more tests in though if anyone thinks that would be useful!

❤️  Thank you!


wsbrunson (approved):
great!

nbierdeman (approved):
Nice! 😲 🥇

PR #647, opened 2023-08-03
remove .only from popup cases

jshawl:
### Description

This PR fixes an issue where only the popup cases ran in CI.

Run all of the tests, instead of just the popup tests.

```
npm run karma
```

❤️  Thank you!


nbierdeman (approved):
Looks good! Let's address failing tests in a separate PR. 🕺

PR #650, opened 2023-08-09
add browser eligiblity check for app shell service worker registration

SuchirAgarwal:
### Description

We currently use the BroadcastChannel browser API in the App Shell service worker. However, we do not currently have checks for browser support of this API, so we add that check prior to service worker registration.

DTLITEPLAT-1147

Run the App Shell flow locally (internal documentation). Test on browser versions that support and don't support BroadcastChannel. Ensure that the service worker only gets registered when the browser version supports service workers and BroadcastChannel.

[https://caniuse.com/broadcastchannel](https://caniuse.com/broadcastchannel)
[https://caniuse.com/serviceworkers](https://caniuse.com/serviceworkers)

❤️  Thank you!

### DIFF
--- a/src/lib/register-service-worker.js
+++ b/src/lib/register-service-worker.js
@@ -38,7 +38,6 @@ const requestSwLogs = () => {
 };
 
 const startBroadCastChannel = () => {
-    // eslint-disable-next-line compat/compat
     broadcastChannel = new BroadcastChannel(LOGS_CHANNEL_NAME);
 };
 
--- a/src/lib/register-service-worker.js
+++ b/src/lib/register-service-worker.js
@@ -72,7 +72,7 @@ const unRegisterButtonHandlers = () => {
 const startRegistration = (swUrl) => {
     // eslint-disable-next-line compat/compat
     navigator.serviceWorker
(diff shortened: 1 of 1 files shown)


src/lib/register-service-worker.js
```diff
@@ -120,7 +120,8 @@ const executeServiceWorker = (releaseHash: string, serviceWorker: string) => {
 }
 
 export function registerServiceWorker({ dumbledoreCurrentReleaseHash, dumbledoreServiceWorker }: RegisterServiceWorkerParams) {
-    if ('serviceWorker' in navigator === false) {
+    // Browser eligibility criteria for SW registration
+    if ('serviceWorker' in navigator === false || 'BroadcastChannel' in window === false) {
         getLogger().info(`${ LOG_PREFIX }NOT_SUPPORTED`);
```
wsbrunson, line 125: should we log which features were available?

```suggestion
        getLogger().info(`${ LOG_PREFIX }NOT_SUPPORTED`, {
          serviceWorker: Boolean('serviceWorker' in navigator),
          BroadcastChannel: Boolean('BroadcastChannel' in window),
        });
```

wsbrunson (approved):

wsbrunson (approved):

PR #654, opened 2023-08-11
validate id token is present to enable wallet capture

ravishekhar:
### Description
* At present the wallet capture payment flow validates `serviceData.wallet` to be truthy to enable wallet capture payment flow
* The `serviceData.wallet` object is always an empty object, thus it always results in truthy object. 
* Ensure the payment flow checks for id-token to enable wallet capture payment flow. 

❤️  Thank you!


wsbrunson (approved):
very cool 😎 i still owe you those charts so we can monitor this

wsbrunson (approved):

PR #657, opened 2023-08-18
logging improvments for spb

wsbrunson:
### Remove `button_render_*` logs
We make a lot of `info` calls just to send up basic configuration data about the buttons. We're changing this to a single info call, with data included.

If we find that we were using any of these info logs for charts in signalFX, we can create a metric instead

### `sendMetric` interface change
We initially were not using the `metricEvent` property for metrics but this is a recommended approach. In the near future, we'll change our metrics like: `pp.app.paypal_sdk.buttons.click.count` to `pp.app.paypal_sdk.buttons.transaction.count` and provide events like `click`, `error`, and `success`.

### `unhandled_exception` metric
Start sending counts for `unhandled_exception` so we can track them as a chart

### small cleanup in `button/logger.js`
some snake_casing variables were annoying me so I changed them


src/button/logger.js
```diff
@@ -210,10 +206,10 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
             [FPTI_CUSTOM_KEY.SHIPPING_CALLBACK_PASSED]: onShippingChange ? '1' : '0'
         }
 
-        const fiID = getFundingInstrumentID();
+        const fundingInstrumentId = getFundingInstrumentId();
```
SancSalix, line 209: I would quibble that this is some yak shaving but this overall PR is such lovely cleanup overall

src/lib/logger.js
```diff
@@ -39,20 +39,24 @@ export function getLogger(): LoggerType {
...
   },
   // $FlowFixMe beaver-logger types need to be updated
 |}) => getLogger().metric({
-    metricNamespace: name,
-    metricEventName: 'unused',
-    metricValue: 1,
     // $FlowFixMe beaver-logger types need to be updated
     dimensions,
+    metricEventName: event,
+    metricNamespace: name,
```
SancSalix, line 58: would it be impossible to write a small test for this? i know there's not a lot in this codebase but i get antsy still

SancSalix: And by this I mean the overall changes, not metricNamespace in particular

src/button/logger.js
```diff
@@ -141,21 +141,21 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
           });
         }
 
-        logger.info(`button_render`);
-        logger.info(`button_render_template_version_${ serverRenderVersion }`);
-        logger.info(`button_render_client_version_${ getClientVersion() }`);
-        logger.info(`button_render_color_${ color }`);
```
SancSalix, line 147: 🌟

src/lib/logger.js
wsbrunson: I'm not sure if tests for this particular function would be that useful. Its just a wrapper around a different function. The type system here is providing us with safety.

Also with changes to metrics/logs, its really easy to verify they are working correctly as you deploy. If the charts suddenly stop displaying info, you messed up

src/button/logger.js
wsbrunson: sometimes the casing in these codebases make my eyes itch

src/button/logger.js
```diff
@@ -141,21 +141,21 @@ export function setupButtonLogger({ env, sessionID, buttonSessionID, clientID, p
...
+            shape,
+            fundingInstruments: fundingSources.join(","),
+            fundingInstrumentsCount: fundingSources.length.toString(),
+            js_sdk_version: getClientVersion() ,
+            storageState: isStorageStateFresh() ? 'fresh' : 'not_fresh',
+            tagline: tagline.toString(),
+            version: serverRenderVersion,
+            walletInstruments: walletInstruments.join(","),
+            walletInstrumentsCount: walletInstruments.length.toString(),
+        });
```
jshawl, line 158: 🤩

src/lib/logger.js
```diff
@@ -39,20 +39,24 @@ export function getLogger(): LoggerType {
...
+  name,
+  value = 1,
 }: {|
+  event?: string,
   name: string,
+  value?: number,
   dimensions: {
     [string]: mixed
   },
   // $FlowFixMe beaver-logger types need to be updated
```
jshawl, line 53: I wonder if there is another quick win here aside from updating the `beaver-logger` types? I realized you weren't even editing this line, but could be a solid addition to this PR!

src/lib/logger.js
```diff
@@ -122,6 +126,14 @@ export function setupLogger({
       err: stringifyError(err),
     });
 
+    sendMetric({
+      event: 'error',
+      name: 'pp.app.paypal_sdk.buttons.unhandled_exception.count',
+      dimensions: {
+        errorType: "payments_sdk_error"
+      }
+    });
```
jshawl, line 135: I think one cool enhancement would be constants and/or types for these properties

jshawl (approved):
Looks great! Left a few non-blocking suggestions about types but that can be addressed in the future (if wanted)! :shipit:

src/lib/logger.js
wsbrunson: do you imagine they'd be local to this module or something shared across the entire SDK?

src/lib/logger.js
SancSalix: That's very fair. I assume we'd know immediately and could roll back?

wsbrunson: yeah absolutely. its one of those things that would fail catastrophically so we'd see it happen in sandbox before we even get to production. instead of a single chart or a couple of charts looking strange, all of them would or just be missing data completely

src/lib/logger.js
jshawl: maybe local to start? because this function is defined in this repo.

eg from https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/lib/logger.js#L45

```ts
export const sendMetric = ({
  name,
  dimensions,
}: {|
  name: 'unused' | 'error' | 'pp.app.paypal_sdk.buttons.click.error.count' | 'pp.app.paypal_sdk.buttons.server_render_version_mismatch',
  dimensions: {
    [string]: mixed
  },
```

idk there are a lot of names - I don't think I like my suggestion anymore 🙈

src/lib/logger.js
wsbrunson: fuck it: https://github.com/krakenjs/beaver-logger/pull/77

jshawl: 🙌

PR #661, opened 2023-08-23
(feat): 🔥 Add logs for invalid shipping change patches

Zoghayyer:

We previously attempted to migrate onShippingChange traffic from `xoon` to `weasley` but we found out that there are some merchants who are using unsupported patch paths, such as `/purchase_units/@reference_id=="default"`. Hence, we had to ramp down. Now, we'd like to add some logs to identify the count of invalid patches that come from merchants with the client integration.

For more info about the issues that we encountered, see https://engineering.paypalcorp.com/confluence/display/Checkout/Shipping+callback%28v1%29+migration+blockers

`Xoon` -> legacy app for guest and signup flows 
`Weasley` -> modern app for guest and signup flows

### Screenshots (if a
Screenshot 2023-08-22 at 8 17 21 PM

Screenshot 2023-08-22 at 8 17 57 PM
pplicable)

❤️  Thank you!


src/props/onShippingChange.js
```diff
@@ -100,11 +101,86 @@ export type OnShippingChangeActionsType = {|
...
 
     const patch = (data = {}) => {
-        if(useShippingChangeCallbackMutation){
+        const shouldUsePatchShipping = Boolean(useShippingChangeCallbackMutation && !buyerAccessToken && isWeasley(appName));
+        logInvalidShippingChangePatches({ appName, buyerAccessToken, data, shouldUsePatchShipping });
+
+        // Be aware when using `patchShipping` with LAST; Checkoutsellerplatserv(CSP) will throw a PERMISSION_DENIED error for any patch
+        // that contains unsupported paths. There are some merchants with a client integration who are using unsupported paths, such as /purchase_units/@reference_id=='default'.
+        // See https://github.paypal.com/Checkout-R/checkoutsellerplatserv/blob/0822237cd1c8c6dbe01a10f0be576c481d9fcb18/checkoutsellerplatservCore/src/main/java/com/paypal/checkout/orders/core/util/v2/PatchUtility.java#L486-L490
+        // This isn't an issue with `patchOrder` because there is a fallback call to smart API which will use the buyer access token to patch the order with the disallowed paths.
```
jshawl, line 182: Thank you for the helpful comment! are you open to removing the internal details here (but only because this is a public repo) maybe something like:

```
// patchShipping is only supported for legacy integrations
// see LI-1234 for more details
```

or something to that effect?

Zoghayyer: Yes, good point 👍

src/props/onShippingChange.js
```diff
@@ -100,11 +101,86 @@ export type OnShippingChangeActionsType = {|
...
+        }
+    } catch(err) {
+        getLogger()
+        .error('button_shipping_change_patch_data_logging_failed', {
+            appName,
+            errMessage: JSON.stringify(err),
+            hasBuyerAccessToken: String(Boolean(buyerAccessToken)),
+            shouldUsePatchShipping: String(shouldUsePatchShipping)
+        })
+        .flush();
```
wsbrunson, line 169: Can we remove these `.flush()`s? This should be handled later in the flow.

We get a lot of push back from merchants because of multiple logging requests
```suggestion
                });
            }
        } else {
            getLogger()
            .info('button_shipping_change_patch_data_is_object', {
                appName,
                hasBuyerAccessToken: String(Boolean(buyerAccessToken)),
                shouldUsePatchShipping: String(shouldUsePatchShipping)
            });
        }
    } catch(err) {
        getLogger()
        .error('button_shipping_change_patch_data_logging_failed', {
            appName,
            errMessage: JSON.stringify(err),
            hasBuyerAccessToken: String(Boolean(buyerAccessToken)),
            shouldUsePatchShipping: String(shouldUsePatchShipping)
        });
```

src/props/onShippingChange.js
```diff
@@ -176,10 +177,7 @@ export function buildXShippingChangeActions({ orderID, actions, facilitatorAcces
         const shouldUsePatchShipping = Boolean(useShippingChangeCallbackMutation && !buyerAccessToken && isWeasley(appName));
         logInvalidShippingChangePatches({ appName, buyerAccessToken, data, shouldUsePatchShipping });
 
-        // Be aware when using `patchShipping` with LAST; Checkoutsellerplatserv(CSP) will throw a PERMISSION_DENIED error for any patch
-        // that contains unsupported paths. There are some merchants with a client integration who are using unsupported paths, such as /purchase_units/@reference_id=='default'.
-        // See https://github.paypal.com/Checkout-R/checkoutsellerplatserv/blob/0822237cd1c8c6dbe01a10f0be576c481d9fcb18/checkoutsellerplatservCore/src/main/java/com/paypal/checkout/orders/core/util/v2/PatchUtility.java#L486-L490
-        // This isn't an issue with `patchOrder` because there is a fallback call to smart API which will use the buyer access token to patch the order with the disallowed paths.
+        // For more details about this change, see DTOPPOR-1620
```
jshawl, line 180: :100:

src/props/onShippingChange.js
Zoghayyer: Sure; will update

src/props/onShippingChange.js
```diff
@@ -92,19 +93,92 @@ export type OnShippingChangeData = {|
...
+
+const sanitizePatch = (rejected: $ReadOnlyArray<string>, patch: Query): $ReadOnlyArray<string> => {
+    const { path } = patch;
+
+    if (!pathPattern.test(path)) {
+        // $FlowFixMe
+        rejected.push(path);
+    }
+    return rejected;
+};
```
rjstires, line 131: My code lives!

rjstires (approved):

src/props/onShippingChange.js
```diff
@@ -92,19 +93,92 @@ export type OnShippingChangeData = {|
...
+    }
+}
+
+export function buildXShippingChangeActions({ orderID, actions, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, clientID, experiments, appName } : {| orderID : string, actions : OnShippingChangeActionsType, facilitatorAccessToken : string, buyerAccessToken : ?string, partnerAttributionID : ?string, forceRestAPI : boolean, experiments: Experiments; clientID: string; appName: string; |}) : XOnShippingChangeActionsType {
     const { useShippingChangeCallbackMutation } = experiments;
 
     const patch = (data = {}) => {
-        if(useShippingChangeCallbackMutation){
+        const shouldUsePatchShipping = Boolean(useShippingChangeCallbackMutation && !buyerAccessToken && isWeasley(appName));
+        // $FlowFixMe
```
jshawl, line 177: it would be nice to remove the `$FlowFixMe` here - is there a way to get what you need out of `data` before calling `logInvalidShippingChangePatches`?

src/props/onShippingChange.test.js
```diff
@@ -5,17 +5,29 @@ import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
...
-import { getLogger } from "../lib";
 
-import { getOnShippingChange } from "./onShippingChange";
+import {
+  getOnShippingChange,
+  logInvalidShippingChangePatches,
+  isWeasley,
+} from "./onShippingChange";
 
 vi.mock("../api");
```
jshawl, line 15: are you open to refactoring line 29 `const mockPatchOrder = patchOrder;` to take advantage of a named spy? something like:

```
const mockPatchOrder = vi.fn();
vi.mock("../api", async () => {
  const actual = await vi.importActual("../api");
  return {
    ...actual,
    mockPatchOrder
  }
})
```

src/props/onShippingChange.test.js
```diff
@@ -218,4 +262,91 @@ describe("onShippingChange", () => {
...
+      expect(logger.info).toHaveBeenCalledWith(
+        "button_shipping_change_patch_data_is_object",
+        {
+          appName: "weasley",
+          hasBuyerAccessToken: "false",
+          shouldUsePatchShipping: "true",
+        }
+      );
+    });
+  });
```
jshawl, line 351: nice test coverage!!!

jshawl (commented):
Lookin good! I left a few comments about removing the `$FlowFixMe`s - lmk if you want a second set of eyes on fixing them - some might not be worth the effort but maybe there's a quick win in there somewhere!

src/props/onShippingChange.js
Zoghayyer: It was throwing type errors because the `data` param(L175) has a default value of an `object` whereas in `LogInvalidShippingChangePatchesPayload` its type is an array of objects. I believe the default `object` value is incorrect(it should be array of objects) but I didn't want to change it since it was written 4 years ago.

jshawl: oh you're right https://developer.paypal.com/docs/api/orders/v2/#orders_patch it _should_ be an array hmmm

jshawl: looks like this might work:

```diff
diff --git a/src/props/onShippingChange.js b/src/props/onShippingChange.js
index e1c932dc..25bfed53 100644
--- a/src/props/onShippingChange.js
+++ b/src/props/onShippingChange.js
@@ -70,7 +70,7 @@ export type XOnShippingChangeActionsType = {|
     resolve : () => ZalgoPromise<void>,
     reject : (string) => ZalgoPromise<void>,
     order : {|
-        patch : (data: $ReadOnlyArray<string>) => ZalgoPromise<OrderResponse>
+        patch : (data: $ReadOnlyArray<Query>) => ZalgoPromise<OrderResponse>
     |}
 |};

@@ -104,7 +104,7 @@ export type OnShippingChangeActionsType = {|

 export type LogInvalidShippingChangePatchesPayload = {|
     appName : string,
-    buyerAccessToken? : string,
+    buyerAccessToken? : string | null,
     data : $ReadOnlyArray<Query>,
     shouldUsePatchShipping : boolean,
 |};
@@ -172,9 +172,8 @@ export const logInvalidShippingChangePatches = ({ appName, buyerAccessToken, dat
 export function buildXShippingChangeActions({ orderID, actions, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, clientID, experiments, appName } : {| orderID : string, actions : OnShippingChangeActionsType, facilitatorAccessToken : string, buyerAccessToken : ?string, partnerAttributionID : ?string, forceRestAPI : boolean, experiments: Experiments; clientID: string; appName: string; |}) : XOnShippingChangeActionsType {
     const { useShippingChangeCallbackMutation } = experiments;

-    const patch = (data = {}) => {
+    const patch = (data = ([]: $ReadOnlyArray<Query>)) => {
         const shouldUsePatchShipping = Boolean(useShippingChangeCallbackMutation && !buyerAccessToken && isWeasley(appName));
-        // $FlowFixMe
         logInvalidShippingChangePatches({ appName, buyerAccessToken, data, shouldUsePatchShipping });
```

though I'm a little nervous about defaulting to `[]` instead of `{}` because it has been there for a while. I think your idea of leaving it as is works here.

Zoghayyer: Yes, I'd hesitant to update it before we can verify that merchants are not passing objects(I don't think that's the case). That's why I added the [button_shipping_change_patch_data_is_object](https://github.com/paypal/paypal-smart-payment-buttons/pull/661/files#diff-6ff70a9568d157b130cfcc04c11465fa803312d358b057769e5a9f8df8505397R155) log.

I removed `$FlowFixMe` and extended the data type to accept both array of objects(queries) and an empty object. I think having some type check is better than not at all.

Zoghayyer:
@jshawl is it possible to test those changes in a `testenv` before we release them? OR what is the verification process before releasing the changes?

wsbrunson (approved):

jshawl (approved):
:shipit: looks great!

PR #664, opened 2023-08-31
feat: add fraud net call for card field render

siddy2181:
### Description

- Add fraudnet call before card field renders
- DTPPCPSDK-732

❤️  Thank you!


src/card/components/page.test.js
```diff
@@ -0,0 +1,108 @@
...
+    }),
+  };
+});
+
+vi.mock("../../api", () => ({
+  // $FlowIssue jest mock
+  loadFraudnet: vi.fn().mockResolvedValue(),
+}));
+
+describe.skip("setupCard", () => {
```
wsbrunson, line 56: we can create a jira ticket at the top of the backlog to come back to this

cgdibble (approved):
Codecoverage failing is acceptable because we are at the very least we are adding tests where once there were none.

PR #666, opened 2023-08-31
Add disable smart api experiment

jshawl:
### Description

This PR leverages a new experiment for skipping calls to smart apis when the REST API calls are not successful.

This is the beginning of the end for the smart APIs. There are several incorrect integration patterns that we have supported over time by falling back to smart API calls. With this experiment, we can add specific merchants to an allow list, and  everyone else will receive the REST error instead of falling back. 

### Reproduction Steps

One way to hit this code path is to create the order server-side (using a client id & access token) and then capture the order on the client-side (using `actions.order.capture`.

### Impact of Change

Only buyers who encounter a 403 from the REST API will be impacted. To start, all of them will continue to fallback to the smart APIs.

mermaid-diagram-2023-08-28-154746

❤️  Thank you!


tests/eligibility.test.js
```diff
@@ -24,7 +24,7 @@ describe("eligibility cases", () => {
     });
   });
 
-  it("should call xprops.remember with venmo if venmo is eligible", async () => {
+  it.skip("should call xprops.remember with venmo if venmo is eligible", async () => {
```
nbierdeman, line 27: Are you thinking we just skip this until the experiment is done?

nbierdeman (approved):
Really nice work, @jshawl! 🤘

tests/eligibility.test.js
jshawl: Good question! I think there's some odd caching going on with the dependencies, because I got the same error in this PR https://github.com/paypal/paypal-smart-payment-buttons/pull/665/files#diff-2d06430dd251e43a1af4130e6fe1a254298d58541a85fa44e0a0dce9558ea3ec and ended up just moving the test  to https://github.com/paypal/paypal-smart-payment-buttons/pull/665/files#diff-9522ce6d5be1d8f889b753e0dca6b22b0948e548805207304a74c23dc4b01056

src/api/order.js
```diff
@@ -146,205 +137,161 @@ export function isUnprocessableEntityError(err : mixed) : boolean {
     }));
 }
 
-export function captureOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
+export function captureOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, experiments } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
     getLogger().info(`capture_order_lsat_upgrade_${ getLsatUpgradeCalled() ? 'called' : 'not_called' }`);
     getLogger().info(`capture_order_lsat_upgrade_${ getLsatUpgradeError() ? 'errored' : 'did_not_error' }`, { err: stringifyError(getLsatUpgradeError()) });
```
jshawl, line 142: In a future PR, these two lines for each for the `*Order` functions can be removed / used in https://github.com/paypal/paypal-smart-payment-buttons/blob/38761ad6d9d3b08557bb4ca8f0ef15b49a92a4cd/src/api/auth.js#L104

src/api/order.js
```diff
@@ -79,55 +79,46 @@ export function isInvalidResourceIDError(err: mixed) : boolean {
...
 
-    return callSmartAPI({
-        accessToken: buyerAccessToken,
-        url:         `${ SMART_API_URI.ORDER }/${ orderID }`,
-        eventName:   'order_get',
-        headers:     {
-            [HEADERS.CLIENT_CONTEXT]:         orderID
+        if(experiments.disableSmartAPI){
+            throw err;
         }
```
jshawl, line 104: ☝️ it was! this message is getting fixed with part 1 here: https://github.com/paypal/paypal-smart-payment-buttons/pull/665 and then in a future part 2, codecov can read the output of `vitest --coverage`

src/api/order.js
```diff
@@ -100,6 +101,10 @@ export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccess
                 getLogger().warn(`get_order_invalid_resource_id_error`, { restCorrID, orderID, err: stringifyError(err) })
             }
 
+            if(experiments.disableSmartAPI){
+                throw err;
+            }
```
jshawl, line 106: ☝️ this is covered! fix incoming in https://github.com/paypal/paypal-smart-payment-buttons/pull/670

src/payment-flows/vault-capture.js
```diff
@@ -146,7 +146,7 @@ function initVaultCapture({ props, components, payment, serviceData, config } :
 
             const { status, body } = validate;
             return handleValidatePaymentMethodResponse({ ThreeDomainSecure, status, body, createOrder, getParent }).then(() => {
-                return confirmOrderAPI(orderID, { payment_source: buildPaymentSource(paymentMethodID) }, { facilitatorAccessToken: accessToken, partnerAttributionID })
+                return confirmOrderAPI(orderID, { payment_source: buildPaymentSource(paymentMethodID) }, { facilitatorAccessToken: accessToken, partnerAttributionID, experiments: {} })
```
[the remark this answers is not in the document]
wsbrunson: I don't think `confirmOrderAPI` falls back to the smart api though?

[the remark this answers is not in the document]
wsbrunson: but also yes, we should definitely make different types for these experiements since they come from different routes `/smart/button` vs `/smart/card-field`

wsbrunson (approved):

PR #667, opened 2023-09-01
feat(checkout): change size of popup in webview

mnicpt:
### Description
https://paypal.atlassian.net/browse/DTMEMBER-3002

❤️  Thank you!


src/payment-flows/checkout.js
```diff
@@ -134,7 +134,18 @@ function getContext({ win, isClick, merchantRequestedPopupsDisabled } : {| win :
 }
 
 export const getDimensions = (fundingSource : string, popupIncreaseDimensions? : boolean) : {| width : number, height : number |} => {
-    if (APM_LIST.indexOf(fundingSource) !== -1) {
+    if (isWebView()) {
+        getLogger().info(`popup_dimensions_${ fundingSource }`);
+        sendCountMetric({
```
wsbrunson, line 139: do we need the `info` log if we have a metric?

wsbrunson (approved):

src/payment-flows/checkout.js
mnicpt: @wsbrunson It's a great question! Just don't know who depends on what from analytics so mirrored what was done with the rest.

PR #671, opened 2023-09-05
Move vitest tests from test/ to src/

jshawl:
### Description

This PR moves the remaining unit tests inside the `tests` folder next to their implementation files.

I was wondering why there were both a `test` and `tests` folder but realized that there were only two files inside of the plural version, so I thought moving the remaining files would help future contributors know which folder to look in.

### Screenshots

**Before**:
Screenshot 2023-09-06 at 09 08 59

**After**:

Screenshot 2023-09-06 at 09 09 57

❤️  Thank you!


src/native/fallback/utilities.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
+/* eslint-disable promise/no-native, no-restricted-globals */
+/* @flow */
+import { vi } from "vitest";
+
```
jshawl, line 5: this file was renamed and I deleted the unused `setupButtonWithDefault` function

src/native/fallback/utilities.js
wsbrunson, line 1: we should name this `test-utilities` or `testUtilities`

wsbrunson (approved):

nbierdeman (approved):
💯

PR #674, opened 2023-09-07
(feat): Update sanitization rules for patchShipping

Zoghayyer:
### Description

- Update regex sanitization pattern for `patchShipping` mutation to allow changing the shipping name.
- We discovered that `checkoutsellerplatserv` allows patching the shipping name with a low access token; hence, we're modifying the sanitization rules to allow it. 

- Update regex sanitization pattern for `patchShipping` mutation to allow changing the shipping name.
- We discovered that `checkoutsellerplatserv` allows patching the shipping name with a low access token; hence, we're modifying the sanitization rules to allow it.

❤️  Thank you!


jshawl (approved):
:shipit:

PR #675, opened 2023-09-07
Pass the ID token through card fields VwoP

elizabethmv:
### Description
This PR passes the ID token through the vault without purchase flow into the update mutation.

We are making these changes to better support different partner scenarios.

- Run SDK/SPB locally
- Set up a card fields vault without purchase flow and pass the ID token
- See the ID token passed in the update mutation

N/A

GQL changes are required to enable the behavior we want, but can go out independently of each other.

@cgdibble @siddy2181 @wsbrunson


wsbrunson (approved):

cgdibble (approved):
Approved! But I'd like to look at adding unit tests. It's probably going to be a frustrating hurdle to get started, but gets easier down the line. Approving as I don't want it to be a blocker, but something to aim towards and try to do for now

wsbrunson:
What part of these changes would you write a unit test for?

src/card/interface/vault-without-purchase.test.js
```diff
@@ -141,4 +141,32 @@ describe("savePaymentSource", () => {
       vaultToken: defaultVaultSetupToken,
     });
   });
+
+  test("should pass the ID token to updateVaultSetupToken if present", async () => {
```
elizabethmv, line 145: @cgdibble @wsbrunson here is what I view as a unit test translates to from the changes directly in the PR. Let me know if you see value in it

PR #677, opened 2023-09-07
Update onShippingOptionsChange and onShippingAddressChange actions interface

jshawl:

❤️  Thank you!

### DIFF
--- a/src/props/onShippingAddressChange.js
+++ b/src/props/onShippingAddressChange.js
@@ -31,12 +31,15 @@ export type XOnShippingAddressChangeDataType = {|
     errors : typeof SHIPPING_ADDRESS_ERROR_MESSAGES
 |};
 
+type BuildPatchPayloadArgs = {|
+    options?: $ReadOnlyArray<CheckoutShippingOption>,
+    discount?: string,
+    tax?: string
+|}
+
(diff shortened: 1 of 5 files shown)


src/props/onShippingAddressChange.js
```diff
@@ -95,84 +98,75 @@ export function buildXOnShippingAddressChangeActions({ data, actions: passedActi
...
-                }
-            };
-
-            return actions;
-        },
-
-        query: () => {
+        buildPatchPayload: ({tax, options, discount}) => {
+            if (tax) {
+                breakdown = buildBreakdown({ breakdown, updatedAmounts: { tax_total: tax } });
```
mnicpt, line 103: @jshawl I love where you are heading with this. Wondering if we can eliminate all of the if-conditions and pass in all params into `updatedAmounts` for both `buildBreakdown` and `calculateTotalFromShippingBreakdownAmounts`? Then we would just have a condition for options to to due things unique to it.

jshawl: @mnicpt we are opening a new pr on a non-fork but have taken this comment into consideration (nate and jesse worked on it together so you will see it soon!)

PR #680, opened 2023-09-12
Experimentally handle the vault-capture payment flow

elizabethmv:
### Description
This PR updates the vault capture flow to support experimentally handling the payment. The experiment flow eliminates the validate payment method call and handles 3DS through confirm payment source instead.

We we to use confirm payment source only, but slowly monitor and ramp traffic before deprecating completely.

Available internally

N/A

SCNW experiment changes will need to go out to get this working properly but can be depolyed independently of each other.


test/client/vault.js
```diff
@@ -1604,36 +1588,14 @@ describe('vault cases', () => {
...
-                    }
-
-                    return {};
-                })
-            }).expectCalls();
-
             const confirmCall = getConfirmOrderApiMock({
                 extraHandler: expect('confirmOrder', ({ headers }) => {
-                    if (headers.authorization !== `Bearer ${accessTokenWithTargetSubject}`) {
+                    if (headers.authorization !== `Bearer ${userIDToken}`) {
```
cgdibble, line 1598: I think this is largely me not being familiar with Kraken's testing, but what's exactly being tested here? Is it just asserting the auth token passed and that the function was called (since it has `expectCalls` chained to it)?

Is there a reason to have explicit `throw` statements rather than using assertions specifically? 

e.g.
```
expect(headers.authorization).not.toBe(`Bearer ${userIDToken}`)
```

cgdibble (approved):

siddy2181 (approved):

PR #681, opened 2023-09-12
change default amount to remove decimals

wsbrunson:
https://paypal.atlassian.net/browse/LI-18516

`0.00` is not a valid amount for currencies like JPY which don't support decimals. Backend team has assured us this should still work for decimal currencies


jshawl (approved):
TIL! :shipit:

elizabethmv (approved):

PR #686, opened 2023-09-25
create new access token

xuanwan05:
### Description
Create a new access token (with ignoreCache) & upgrade it when merchant is making a call to any of actions.order.* methods like actions.order.get(), actions.order.patch(), actions.order.capture() .. etc. 

Also added the test case to cover all the added code.

❤️  Thank you!


src/api/auth.js
```diff
@@ -128,6 +128,34 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
...
+            ) {
+                createUpgradedLowScopeAccessToken(
+                    token: $orderID
+                    buyerAccessToken: $buyerAccessToken
+                    merchantLSAT: $facilitatorAccessToken
+                )
+            }
+        `,
+            variables: { facilitatorAccessToken, buyerAccessToken, orderID }
+        }).then(noop).catch(err => {
```
ravishekhar, line 153: We need to return the new value from returned from Graphql API to the callers.

src/props/onApprove.js
```diff
@@ -131,6 +131,18 @@ const handleProcessorError = <T>(err : mixed, restart : () => ZalgoPromise<void>
     throw err;
 };
 
+export const createNewAccessToken = async({ orderID, facilitatorAccessToken, buyerAccessToken }) : ?string  => {
```
ravishekhar, line 134: This seems redundant method. Do we really need this abstraction? Can we use `upgradeFacilitatorAccessTokenWithIgnoreCache(...)` directly in below callbacks?

src/api/auth.js
```diff
@@ -128,6 +128,34 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
     });
 }
 
+export function upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken : string, { buyerAccessToken, orderID } : {| buyerAccessToken : string, orderID : string |}) : ZalgoPromise<void> {
```
ravishekhar, line 131: Lets keep a flat params for this method siganture as all params for mandatory for this. 
```suggestion
export function upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken : string, buyerAccessToken : string, orderID : string) : ZalgoPromise<void> {
```

src/props/onApprove.js
```diff
@@ -147,15 +159,30 @@ type OrderActionOptions = {|
 |};
 
 function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
-    const get = memoize(() => {
+    const get = memoize(async() => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
+            } catch (error) {
+                console.error("Cannot create new buyer access token");
```
ravishekhar, line 167: Since we would update upgradeFacilitatorAccessTokenWithIgnoreCache(...) to not throw any errors. We can avoid this try catch block.

src/props/onApprove.js
```diff
@@ -147,15 +159,30 @@ type OrderActionOptions = {|
...
+    const capture = memoize(async() => {
         if (intent !== INTENT.CAPTURE) {
             throw new Error(`Use ${ SDK_QUERY_KEYS.INTENT }=${ INTENT.CAPTURE } to use client-side capture`);
         }
 
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
+            } catch (error) {
+                throw new Error("Cannot create new buyer access token");
```
ravishekhar, line 182: Since we would update upgradeFacilitatorAccessTokenWithIgnoreCache(...) to not throw any errors. We can avoid this try catch block.

src/props/onApprove.js
```diff
@@ -164,18 +191,33 @@ function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, b
...
+    const authorize = memoize(async() => {
         if (intent !== INTENT.AUTHORIZE) {
             throw new Error(`Use ${ SDK_QUERY_KEYS.INTENT }=${ INTENT.AUTHORIZE } to use client-side authorize`);
         }
 
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
+            } catch (error) {
+                throw new Error("Cannot create new buyer access token");
```
ravishekhar, line 203: Since we would update upgradeFacilitatorAccessTokenWithIgnoreCache(...) to not throw any errors. We can avoid this try catch block.

src/props/onApprove.js
```diff
@@ -147,15 +159,30 @@ type OrderActionOptions = {|
...
 
-    const capture = memoize(() => {
+    const capture = memoize(async() => {
         if (intent !== INTENT.CAPTURE) {
             throw new Error(`Use ${ SDK_QUERY_KEYS.INTENT }=${ INTENT.CAPTURE } to use client-side capture`);
         }
 
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
```
ravishekhar, line 180: Lets use upgradeFacilitatorAccessTokenWithIgnoreCache(...) directly 

Also mutation returns the new `facilitatorAccessToken` which should be passed when calling `captureOrder` on line 186.

src/props/onApprove.js
```diff
@@ -164,18 +191,33 @@ function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, b
...
 
-    const authorize = memoize(() => {
+    const authorize = memoize(async() => {
         if (intent !== INTENT.AUTHORIZE) {
             throw new Error(`Use ${ SDK_QUERY_KEYS.INTENT }=${ INTENT.AUTHORIZE } to use client-side authorize`);
         }
 
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
```
ravishekhar, line 201: Same comment as above. Lets use upgradeFacilitatorAccessTokenWithIgnoreCache(...) directly 

Also mutation returns the new `facilitatorAccessToken` which should be passed when calling `authorizeOrder` on line 207.

src/props/onApprove.js
```diff
@@ -164,18 +191,33 @@ function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, b
...
             .finally(get.reset)
             .finally(authorize.reset)
             .catch(err => handleProcessorError<OrderResponse>(err, restart, onError));
     });
 
-    const patch = (data = {}) => {
+    const patch = async(data = {}) => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
```
ravishekhar, line 216: Same comment as above. Lets use upgradeFacilitatorAccessTokenWithIgnoreCache(...) directly 

Also mutation returns the new `facilitatorAccessToken` which should be passed when calling `patchOrder` on line 221.

src/api/auth.js
```diff
@@ -128,6 +128,34 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
...
+                createUpgradedLowScopeAccessToken(
+                    token: $orderID
+                    buyerAccessToken: $buyerAccessToken
+                    merchantLSAT: $facilitatorAccessToken
+                )
+            }
+        `,
+            variables: { facilitatorAccessToken, buyerAccessToken, orderID }
+        }).then(noop).catch(err => {
+            throw err;
```
ravishekhar, line 154: Lets not throw error as the clients can use it in path of capturing or getting order details, lets just log it using beaver logger. 

Also pls follow the above sample to set the lsat upgrade error flag `onLsatUpgradeError(err);`

src/api/auth.js
```diff
@@ -128,6 +128,34 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
     });
 }
 
+export function upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken : string, { buyerAccessToken, orderID } : {| buyerAccessToken : string, orderID : string |}) : ZalgoPromise<void> {
+    return inlineMemoize(upgradeFacilitatorAccessTokenWithIgnoreCache, () => {
```
ravishekhar, line 132: Can you pls test and ensure the `inlineMemoize` will not return the memoized value of any of params like `facilitatorAccessToken`, `buyerAccessToken` or `orderID` changes.

src/api/auth.js
```diff
@@ -128,6 +128,34 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
...
+        return callGraphQL({
```
ravishekhar, line 133: Before calling graphql, set the appropriate flags.
```
clearLsatState();
onLsatUpgradeCalled();
```

src/props/onApprove.js
```diff
@@ -147,15 +159,30 @@ type OrderActionOptions = {|
 |};
 
 function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
-    const get = memoize(() => {
+    const get = memoize(async() => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
```
ravishekhar, line 165: Lets use upgradeFacilitatorAccessTokenWithIgnoreCache(...) directly 

Also mutation returns the new `facilitatorAccessToken` which should be passed when calling `getOrder` on line 170

src/props/onApprove.js
```diff
@@ -164,18 +191,33 @@ function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, b
...
             .catch(err => handleProcessorError<OrderResponse>(err, restart, onError));
     });
 
-    const patch = (data = {}) => {
+    const patch = async(data = {}) => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
+            } catch (error) {
+                throw new Error("Cannot create new buyer access token");
```
ravishekhar, line 218: Since we would update upgradeFacilitatorAccessTokenWithIgnoreCache(...) to not throw any errors. We can avoid this try catch block.

src/props/onApprove.js
```diff
@@ -147,15 +159,30 @@ type OrderActionOptions = {|
 |};
 
 function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
-    const get = memoize(() => {
+    const get = memoize(async() => {
```
ravishekhar, line 162: is `async` supported with `memoize` ? can we use the Promise, then/catch syntax if async is not supported?

src/props/onShippingChange.js
```diff
@@ -9,6 +9,7 @@ import { getLogger } from '../lib';
 import type { OrderAmount, Experiments, FeatureFlags, ShippingOption } from '../types';
 
 import type { CreateOrder } from './createOrder';
+import {createNewAccessToken} from './onApprove';
```
ravishekhar, line 12: Pls use the `upgradeFacilitatorAccessTokenWithIgnoreCache(...) ` from `api/auth` instead of createNewAccessToken.

src/props/onShippingChange.js
```diff
@@ -170,6 +171,14 @@ export function buildXShippingChangeActions({ orderID, actions, facilitatorAcces
                 throw new Error('Order could not be patched');
             });
         }
+
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            try {
+                buyerAccessToken = await createNewAccessToken({ orderID, facilitatorAccessToken, buyerAccessToken });
```
ravishekhar, line 177: Same comment as above. Lets use upgradeFacilitatorAccessTokenWithIgnoreCache(...) directly

Also mutation returns the new facilitatorAccessToken which should be passed when calling patchOrder on line 182.

src/props/onShippingChange.js
```diff
@@ -170,6 +171,14 @@ export function buildXShippingChangeActions({ orderID, actions, facilitatorAcces
...
+            } catch (error) {
+                throw new Error("Cannot create new buyer access token");
```
ravishekhar, line 179: Since we would update upgradeFacilitatorAccessTokenWithIgnoreCache(...) to not throw any errors. We can avoid this try catch block.

ravishekhar (changes requested):

src/props/onApprove.js
```diff
@@ -148,6 +147,9 @@ type OrderActionOptions = {|
 
 function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
     const get = memoize(() => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            facilitatorAccessToken = upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken,buyerAccessToken,orderID);
```
wsbrunson, line 151: I think we need to `await` any call to `upgradeFacilitatorAccessTokenWithIgnoreCache`

src/props/onApprove.js
```diff
@@ -148,6 +147,13 @@ type OrderActionOptions = {|
 
 function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
     const get = memoize(() => {
+        if (experiments?.upgradeLSATWithIgnoreCache) {
+            return upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken, buyerAccessToken, orderID)
+                .then((res) => {
+                    facilitatorAccessToken = res;
+                    return getOrder(orderID, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, experiments });
```
wsbrunson, line 154: I'd suggest the following to make it a little clearer what is happening:
```suggestion
                .then((upgradedFacilitatorAccessToken) => {
                    return getOrder(orderID, { facilitatorAccessToken: upgradedFacilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, experiments });
```

src/api/auth.js
```diff
@@ -128,6 +128,45 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
...
 }
 
+export function upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken : string, buyerAccessToken : ?string, orderID : string) : ZalgoPromise<string> {
+    return inlineMemoize(upgradeFacilitatorAccessTokenWithIgnoreCache, () => {
+        clearLsatState();
+        onLsatUpgradeCalled();
+        const headers = {};
+        if (buyerAccessToken) {
+            headers[HEADERS.ACCESS_TOKEN] = buyerAccessToken;
+          }
```
ravishekhar, line 138: At this point the buyer access will be available. We can remove this if condition. 
```suggestion
          headers[HEADERS.ACCESS_TOKEN] = buyerAccessToken;
```

xuanwan05: type OrderActionOptions = {|
    orderID : string,
    paymentID : ?string,
    payerID : ?string,
    restart : () => ZalgoPromise<void>,
    intent : $Values<typeof INTENT>,
    facilitatorAccessToken : string,
    buyerAccessToken : ?string,
    partnerAttributionID : ?string,
    forceRestAPI : boolean,
    onError : OnError,
    experiments: Experiments
|};

function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
    const get = memoize(() => {
        if (experiments?.upgradeLSATWithIgnoreCache) {
            return upgradeFacilitatorAccessTokenWithIgnoreCache(facilitatorAccessToken, buyerAccessToken, orderID)
   

buyerAccessToken is optional, so when we pass it to upgradeFacilitatorAccessTokenWithIgnoreCache, the value could be undefined or null.

ravishekhar: We should make it mandatory. Without this data the mutation will fail.

src/props/onApprovePatchAction.test.js
```diff
@@ -0,0 +1,80 @@
...
+    const buildXApproveOrderActionsResult = buildXApproveOrderActions({
+      ...commonOptions,
+      experiments: { upgradeLSATWithIgnoreCache: true },
+    });
+    const { order } = buildXApproveOrderActionsResult;
+
+    await order.patch();
+
+    // Expect the mock to have been called
+    expect(callGraphQL).toHaveBeenCalled();
```
ravishekhar, line 63: can you pls add validation to check if right graphql mutation was called.

src/props/onApprovePatchAction.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+describe("getOnApproveOrder patch action", () => {
+  test("invoke callGraphQL from onApprove patch action if treatment is present", async () => {
+    const buildXApproveOrderActionsResult = buildXApproveOrderActions({
+      ...commonOptions,
+      experiments: { upgradeLSATWithIgnoreCache: true },
+    });
+    const { order } = buildXApproveOrderActionsResult;
+
+    await order.patch();
```
ravishekhar, line 60: Pls assert if v2 orders api was called when calling patch

ravishekhar (changes requested):

src/api/auth.js
```diff
@@ -128,6 +128,44 @@ export function upgradeFacilitatorAccessToken(facilitatorAccessToken : string, {
...
+                    token: $orderID
+                    buyerAccessToken: $buyerAccessToken
+                    merchantLSAT: $facilitatorAccessToken
+                )
+            }
+        `,
+            variables: { facilitatorAccessToken, buyerAccessToken, orderID }
+        })
+        .then(res => {
+            getLogger().info('upgradeFacilitatorAccessTokenWithIgnoreCache returns successfully');
```
elizabethmv, line 158:
```suggestion
```
This log looks off, but also I don't think we need to add any success logs here. I would recommend removing it.

src/props/onApprove.test.js
```diff
@@ -105,5 +105,51 @@ describe("onApprove", () => {
         expect.anything()
       );
     });
+
+    test("invoke upgradeFacilitatorAccessTokenWithIgnoreCache call if treatment is present", async () => {
+      const getOnApproveOrderResult = getOnApproveOrder({
+        ...getOnApproveOrderOptions,
+        experiments: {
+          btSdkOrdersV2Migration: true,
```
elizabethmv, line 113:
```suggestion
```
Is this needed? It looks more like carry over.

src/props/onApprove.test.js
```diff
@@ -105,5 +105,51 @@ describe("onApprove", () => {
...
+        }),
+        expect.anything()
+      );
+    });
+
+    test("invoke upgradeFacilitatorAccessTokenWithIgnoreCache call if treatment is not present", async () => {
+      const getOnApproveOrderResult = getOnApproveOrder({
+        ...getOnApproveOrderOptions,
+        experiments: {
+          btSdkOrdersV2Migration: true,
```
elizabethmv, line 136:
```suggestion
```
Same comment about needing this here.

src/props/onApproveGetAction.test.js
```diff
@@ -0,0 +1,79 @@
...
+  test("invoke callGraphQL from onApprove get action if treatment is present", async () => {
+    const buildXApproveOrderActionsResult = buildXApproveOrderActions({
+      ...commonOptions,
+      experiments: { upgradeLSATWithIgnoreCache: true },
+    });
+    const { order } = buildXApproveOrderActionsResult;
+
+    await order.get();
+
+    // Expect the mock to have been called
```
elizabethmv, line 62: I would recommend removing all of the "Expect the mock to have been called" comments, the assertion reads the same.
```suggestion
```

elizabethmv (commented):
Mostly clean up comments, looking good so far!

src/props/onApprove.test.js
xuanwan05: it's needed, otherwise it fails at `merchantOnApprove` check

jshawl: I think you can remove `btSdkOrdersV2Migration` here but then in the `expect` down below, remove `paymentID: "abc123"` - fwiw I am hoping to remove this experiment altogether in this sprint so I imagine this test will be updated to account for that (`paymentID` will always be the `orderID` but without the `EC-` prefix)

src/props/onApprove.js
```diff
@@ -253,7 +292,7 @@ type ApproveOrderActionOptions = {|
     experiments: Experiments
 |};
 
-function buildXApproveOrderActions({ intent, orderID, paymentID, payerID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : ApproveOrderActionOptions) : XOnApproveOrderActionsType {
+export function buildXApproveOrderActions({ intent, orderID, paymentID, payerID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : ApproveOrderActionOptions) : XOnApproveOrderActionsType {
```
ravishekhar, line 295: Lets create a followup PR to add test cases for public method getOnApproveOrder and remove this private method from being exported.

ravishekhar (approved):
Looks good 👍 . Lets create a followup PR to address the gaps in unit tests.

elizabethmv (approved):

PR #691, opened 2023-10-02
Remove PayPal on File

jshawl:

❤️  Thank you!


wsbrunson (approved):
Looks great to me but we should get a thumbs up from @ravishekhar

src/button/button.js
```diff
@@ -112,9 +104,9 @@ export function setupButton({
 
     const { merchantID, buyerCountry } = serviceData;
 
-    const paymentSource = enableOrdersApprovalSmartWallet ? FUNDING.PAYPAL : null;
+    const paymentSource = FUNDING.PAYPAL;
```
nbierdeman, line 107: I think payment source had a default value of `null` before PPoF, but am not sure if that's required still: https://github.com/paypal/paypal-smart-payment-buttons/pull/552/files#diff-05448c44a11aec8c317ff422def5b0f7b332d62b0234a10ea45f0021e2c28b14L107

src/button/button.js
```diff
@@ -319,7 +311,7 @@ export function setupButton({
     const setupButtonLogsTask = setupButtonLogger({
         style, env, sdkVersion, sessionID, clientID, partnerAttributionID, commit, sdkCorrelationID,
         stickinessID, buttonCorrelationID, locale, merchantID, buttonSessionID, merchantDomain,
-        fundingSource, getQueriedEligibleFunding, buyerCountry, onShippingChange, wallet, smartWalletOrderID, enableOrdersApprovalSmartWallet, product });
+        fundingSource, getQueriedEligibleFunding, buyerCountry, onShippingChange, product });
```
nbierdeman, line 314: Is `product` still needed?

nbierdeman: Oh, never mind. I see it's still being used for logging.

nbierdeman (approved):
Nice work cleaning this up! I just had one question about the default value of `paymentSource`, but otherwise looks good to go! 🕺

src/button/button.js
jshawl: ah good catch!

elizabethmv (approved):

PR #692, opened 2023-10-02
feat: use id token for cp4 scenarios

cgdibble:
### Description

This will use the idToken if it is provided in place of the auth assertion. This is in support of CP4 vaulting.

It is behind an experiment flag.

❤️  Thank you!


elizabethmv (approved):

src/card/interface/submitCardFields.js
```diff
@@ -111,6 +113,16 @@ export function submitCardFields({
...
+  if (experiments.useIDToken === true) {
+    // CP4 merchants will use the `userIDToken` to auth to associate the transaction with a merchant
+    if (cardProps.userIDToken) {
+      facilitatorAccessToken = cardProps.userIDToken;
+    } else {
+      getLogger().warn(`hcf_missing_userIDToken`)
+    }
+      
+
+  }
```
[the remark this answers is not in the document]
cgdibble: @elizabethmv our thinking was that while this experiment is around we would want to know if that feature flag was hit _and_ if the the userIDToken wasn't provided. 
Ultimately it felt challenging to identify how to approach it, though. We started with what you'd suggested initially, but thought knowing if a merchant is enabled for this but not passing the id token would be good info.

Whatcha think?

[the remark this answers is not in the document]
elizabethmv: If you want to know if the experiment was hit in SPB then maybe a top level info instead of a warning that is on the else case? and then you could include additional parameters on what was passed.

Being in the experiment and not passing the userIDtoken is still a valid/positive use case, so if you'd like to keep that info might make more sense?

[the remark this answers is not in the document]
cgdibble: > Being in the experiment and not passing the userIDtoken is still a valid/positive use case

🤔 ah interesting, this must be the breakdown we mixed up. We were thinking that if they were in the experiment they'd be CP4 and thus needing to pass the `userIDToken`, but what are we missing where that's valid otherwise? I'm having a hard time seeing the case play out but suspect it's entirely just an oversight on my side.

src/card/interface/submitCardFields.js
```diff
@@ -111,6 +113,16 @@ export function submitCardFields({
     experiments,
   });
 
+  if (experiments.useIDToken === true) {
```
wsbrunson, line 116:
```suggestion
  if (experiments.useIDToken) {
```

src/card/interface/submitCardFields.js
```diff
@@ -111,6 +113,16 @@ export function submitCardFields({
...
+    // CP4 merchants will use the `userIDToken` to auth to associate the transaction with a merchant
+    if (cardProps.userIDToken) {
+      facilitatorAccessToken = cardProps.userIDToken;
+    } else {
+      getLogger().info(`hcf_userIDToken_present_false`)
```
wsbrunson, line 121: What information are we trying to understand with this log?

Are we looking to see how often this happens or is this for debugging?

Do we need to know the opposite case where the experiment is true and an id token was passed?

siddy2181: > Are we looking to see how often this happens or is this for debugging?
Our thought was primarily for debugging. We assumed that wouldn't trigger the concern we are facing right now and should be okay to skip? But what do you think?

wsbrunson: yeah if its for debugging i think thats okay. I would just leave a comment to remove this log once you are finished with the experiment

wsbrunson (approved):

PR #693, opened 2023-10-05
[DTPPCPSDK-1279] vitest for Call upgrade LSAT when a merchant uses any actions.order.* method updates

xuanwan05:
### Description
Use Vitest to write the test case covering the 'getOnApproveOrder' function call from the top level down to the 'callGraphQL/CreateUpgradedLowScopeAccessToken' call and the order actions calls. Test all four order actions: 'get,' 'patch,' 'authorize,' and 'capture.'

❤️  Thank you!


ravishekhar (approved):
💯

PR #694, opened 2023-10-06
Update logging of Lsat upgrade, cache buyerAccessToken for all payment flows when experiment in enabled, record metrics for api calls

ravishekhar:
### Description
* Reactor onAuth to log `ignore_lsat_upgrade` instead of `upgrade_lsat_success` when LSAT upgrade (with ignore cache) is not called
* Update logs of all Client Side helper Orders API method to log if lsat upgrade was done with or without ignore cache. 
* Cache the buyer access token in onAuth props so that its available to all payment flows and not just checkout payment flow. 
* Record Metrics for all API calls. 

❤️  Thank you!


src/props/onAuth.js
```diff
@@ -27,20 +27,28 @@ type GetOnAuthOptions = {|
...
+        const isInIgnoreCacheExperiment = experiments?.upgradeLSATWithIgnoreCache;
+
         return ZalgoPromise.try(() => {
             if (accessToken) {
+                if (createSubscription || isInIgnoreCacheExperiment) {
+                    // Cache the buyerAccessToken so that it can be used when creating an upgraded lsat with ignore cache
+                    setBuyerAccessToken(accessToken);
+                    getLogger().info("ignore_lsat_upgrade", {
+                        createSubscription: Boolean(createSubscription),
+                        upgradeLSATWithIgnoreCache: Boolean(isInIgnoreCacheExperiment),
```
elizabethmv, line 39:
```suggestion
                        upgradeLSATWithIgnoreCache: isInIgnoreCacheExperiment,
```

src/api/order.js
```diff
@@ -79,10 +79,12 @@ export function isInvalidResourceIDError(err: mixed) : boolean {
         return detail.issue === ORDER_API_ERROR.INVALID_RESOURCE_ID;
     }))
 }
-
+function lsatUpgradeType() : string {
+    return  getLsatUpgradeWithIgnoreCache() ? 'with_ignore_cache_lsat_upgrade' : 'without_ignore_cache_lsat_upgrade';
+}
 export function getOrder(orderID : string, { facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI = false, experiments } : OrderAPIOptions) : ZalgoPromise<OrderResponse> {
-    getLogger().info(`get_order_lsat_upgrade_${ getLsatUpgradeCalled() ? 'called' : 'not_called' }`);
```
elizabethmv, line 84: Do you know if any dashboards are relying on these that we need to update?

src/api/api.js
```diff
@@ -50,12 +54,20 @@ export function callRestAPI<D, T>({ accessToken, method, url, data, headers, eve
...
+            getLogger().warn(`rest_api_${ eventName }_status_${ status }_error`);
+            sendCountMetric({
+                name : `pp.app.paypal_sdk.buttons.rest_api_${ eventName }.error.count`,
+                dimensions : metricDimensions
+            }).flush();
             throw error;
         }
 
+        sendCountMetric({
+            name: `pp.app.paypal_sdk.buttons.rest_api_${ eventName || 'unknown' }.success.count`,
```
wsbrunson, line 67:
```suggestion
            name: `pp.app.paypal_sdk.buttons.rest_api_${ eventName }.success.count`,
```

src/payment-flows/checkout.js
```diff
@@ -317,8 +317,6 @@ function initCheckout({ props, components, serviceData, payment, config, restart
                 buyerAccessTokenReceivedOnAuth = accessToken ? accessToken : buyerAccessToken;
 
                 return onAuth({ accessToken: buyerAccessTokenReceivedOnAuth }).then(token => {
-                    setBuyerAccessToken(token);
```
wsbrunson, line 320: we don't need this anymore?

wsbrunson (commented):
I would take a look at the places where you are calling `.flush()` and decide if its really the end of the flow and you need to flush or its okay to defer. We're currently adding a lot of flushing and I don't think we want to do that

elizabethmv (approved):
Looks good! A couple of non-blockers

src/payment-flows/checkout.js
ravishekhar: I moved this call inside `onAuth` itself so that same behaviour is exhibited for all payment flows

ravishekhar:
> I would take a look at the places where you are calling `.flush()` and decide if its really the end of the flow and you need to flush or its okay to defer. We're currently adding a lot of flushing and I don't think we want to do that

Yeah makes sense on the flush part, since all these events are in mid of other events so these logs can be safely batched.

wsbrunson (approved):

PR #696, opened 2023-10-13
Add missing onShippingAddressChange and onShippingOptionsChange tests

nbierdeman:
### Description

I noticed we were missing test coverage for `data.errors`, `actions.reject()`, and `selectedShippingOption` while working on [DTPPCPSDK-656](https://paypal.atlassian.net/browse/DTPPCPSDK-656). This PR adds coverage for these items.


src/props/onShippingAddressChange.test.js
```diff
@@ -55,6 +59,22 @@ describe("onShippingAddressChange", () => {
         data: rypData,
         orderID: "",
       });
+      data = buildXOnShippingAddressChangeData(rypData);
+    });
+
+    test("returns undefined when passing data.errors to actions.reject()", async () => {
+      expect(await actions.reject(data.errors.ADDRESS_ERROR)).toEqual(
+        ZalgoPromise.resolve().value
+      );
```
jshawl, line 68: (I had to look this up) there is also [`resolves`](https://vitest.dev/api/expect.html#resolves)

so this could be:

```
await expect(actions.reject(data.errors.ADDRESS_ERROR)).resolves.toEqual(undefined)
```

but also I don't think the `ZalgoPromise` is required here so this could be:

```
expect(await actions.reject(data.errors.ADDRESS_ERROR)).toEqual(undefined);
```

but I think this is only because `ZalgoPromise.resolve() === undefined`

jshawl (approved):
Nice! just 1 non-blocking comment :shipit:

src/props/onShippingAddressChange.test.js
nbierdeman: Thanks, @jshawl! Do you think there's any value in mimicking the return value of `actions.reject()`? It's returning a `ZalgoPromise` that resolves to `undefined` ([here](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/onShippingOptionsChange.js#L41) and [here](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/onShippingAddressChange.js#L45)). If not, I will probably roll with `expect(await actions.reject(data.errors.ADDRESS_ERROR)).toEqual(undefined);`.

nbierdeman: ^I was just thinking of this from more of a documentation perspective, like, "Oh, I see `actions.reject()` is expected to return a `ZalgoPromise`," type of thing.

jshawl: oh I like that idea! (I verified `actions.reject` actually resolves even though the name is reject and it does not `ZalgoPromise.reject`)

wsbrunson (approved):

PR #697, opened 2023-10-15
Improve logging when using client side helpers

ravishekhar:
### Description

This PR addresses 3 things to improve logging when a merchant is using any of client side helper methods.

1. Flush the logs at end of every client side helper invocation. In some instances there is an immediate redirection happening post client side helper invocation resulting in the default flush at regular internal not happening. 
2. Log the lsat metric as String instead of boolean. 
3. Log the payee(s) Id of integrations using client side helper. 

❤️  Thank you!


src/api/order.js
```diff
@@ -82,9 +82,34 @@ export function isInvalidResourceIDError(err: mixed) : boolean {
...
+    } else {
+        return `${cacheType}_not_called`
+    }
+}
+function logPayeeInfoForClientSideHelpers(orderID: string, method: "get" | "patch" | "capture" | "authorize") : void {
+    // eslint-disable-next-line no-use-before-define
+    getSupplementalOrderInfo(orderID).then(order => {
+        const payees = order.checkoutSession.payees || [];
+        const merchantIds = payees.map(p => p.merchantId);
+        getLogger().info(`using_client_side_helper_${ method }`, { payee: merchantIds.join(), orderID });
```
jshawl, line 102: What is the benefit from using `merchantId` instead of `clientId` in this context? I am just being curious.

jshawl (approved):
Looks great! :shipit:

src/api/order.js
```diff
@@ -82,9 +82,34 @@ export function isInvalidResourceIDError(err: mixed) : boolean {
 function lsatUpgradeType() : string {
     return  getLsatUpgradeWithIgnoreCache() ? 'with_ignore_cache_lsat_upgrade' : 'without_ignore_cache_lsat_upgrade';
 }
+function lsatUpgradeMetricValue() : string {
+    const lsatUpgradeCalled = Boolean(getLsatUpgradeCalled());
+    const lsatUpgradeIgnoreCache = Boolean(getLsatUpgradeWithIgnoreCache())
+    const lsatUpgradeError = Boolean(getLsatUpgradeError());
+    const cacheType = lsatUpgradeIgnoreCache ? 'with_ignore_cache' : 'without_ignore_cache';
+
+    if(lsatUpgradeCalled) {
```
nbierdeman, line 91: nitpick:

```suggestion
    if (lsatUpgradeCalled) {
```

nbierdeman (approved):
💯 🚢

src/api/order.js
ravishekhar: There are some cases where payee in the order payload is different than the clientID. So we are trying to figure the actual payee using the client side helper methods.

jshawl: ah thank you!

PR #700, opened 2023-10-19
[DTPPCPSDK-1341]do not create new access token if venmo native

xuanwan05:
### Description
do not create new access token if venmo native

❤️  Thank you!


ravishekhar (approved):
looks great 👍 .

just a nit pick;
Can we create a separate new var in the factory methods and use in the if condition.   e.g. 
`const ulsatNotRequired = paymentSource === FUNDING.VENMO && !buyerAccessToken;`

src/props/onApprove.js
```diff
@@ -142,12 +142,15 @@ type OrderActionOptions = {|
...
+    experiments: Experiments,
+    paymentSource: $Values<typeof FUNDING> | null
 |};
 
-function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments } : OrderActionOptions) : OrderActions {
+function buildOrderActions({ intent, orderID, restart, facilitatorAccessToken, buyerAccessToken, partnerAttributionID, forceRestAPI, onError, experiments, paymentSource } : OrderActionOptions) : OrderActions {
     const get = memoize(() => {
-        if (experiments?.upgradeLSATWithIgnoreCache) {
+        // Does not create a new access token if Venmo native
+        const ulsatNotRequired = paymentSource === FUNDING.VENMO && !buyerAccessToken;
```
wsbrunson, line 152: I'd recommend the following:

1. create a function called `shouldSkipUpgragedLSAT` in `src/props/utils.js`
2. keep this comment `// Does not create a new access token if Venmo native` but expand on it:

```
// Does not create a new access token if Venmo native
// venmo native handles upgrading LSAT so if we upgrade
// it in the SDK the upgrade will fail when we switch to venmo
```

It sounds like we completely forgot this happens so hopefully this will be a hint to anyone in the future if this logic has to change again

wsbrunson (commented):
other than my comment, looks great! great tests

elizabethmv (approved):

src/props/onApprove.js
xuanwan05: It will create a dependency cycle between 'utils' and 'onShippingChange' if we create the function in 'utils.js.' Therefore, instead of creating the function in 'utils.js,' can I create it inside 'onApprove' or just keep the way it is now?

wsbrunson: oh wow that's not great.

I'd recommend renaming `utils.js` to `shippingChangeUtils.js` since they've created a cycle. Then we don't have to clean that up right now.

and then create a new file called `utils.js`.

ravishekhar (approved):
lgtm 👍

src/props/utils.js
```diff
@@ -1,128 +1,4 @@
 /* @flow */
+import { FUNDING } from '@paypal/sdk-constants/src';
 
-import type { Breakdown, CheckoutBreakdown, CheckoutShippingOption, ShippingOption } from '../types';
-
-import { type Query, ON_SHIPPING_CHANGE_PATHS } from './onShippingChange';
-
-
-export const calculateTotalFromShippingBreakdownAmounts = ({ breakdown, updatedAmounts } : {| breakdown : Breakdown, updatedAmounts : {| [string] : ?string |} |}) : string => {
```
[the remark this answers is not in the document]
jshawl: The move looks great to me! I opened up the new file and compared the changes side by side to make sure everything was exported correctly. And I feel extra-confident because we have coverage in both karma and vitest for this code that was moved. Nice job! 🥳

PR #705, opened 2023-10-27
Pass sign_out_user=true for vault and wallet capture

jshawl:
### Description

This PR adds a new query parameter `sign_out_user=true` to the popup window when a buyer clicks on "Pay with a different account".

Currently, buyers cannot change their account when they click on "Pay with a different account". This change will log out the buyer so that they can log in with another PayPal account.

### Reproduction Steps 

Vault the PayPal button. Click the menu, and then click on "Pay with a different account". 

https://github.com/paypal/paypal-checkout-components/pull/2262

- ✅~⚠️ the checkout-components changes need to be released before this PR can be released. ⚠️~
- ✅~⚠️ the router changes need to be released before this PR can be released. ⚠️~
  - ~An easy way to check that the router changes have been released is to obtain a `checkoutnow` URL after creating the order, and then adding `&sign_out_user=true` to the URL. If you are signed out, this PR is ready for release! for example: https://www.sandbox.paypal.com/checkoutnow?sessionID=uid_515f21a7de_mtu6mjg6mjm&token=5MV933856C209100D&sign_out_user=true <- error until the router changes are released.~

❤️  Thank you!


nbierdeman (approved):
💯

wsbrunson (approved):

PR #706, opened 2023-10-27
[DTPPCPSDK-1328] Remove old LSAT code

xuanwan05:
### Description
With the introduction and 100% ramp of using CreateUpgradedLowScopeAccessToken mutation, usage of UpgradeFacilitatorAccessToken mutation is no longer needed.

Clean up 

Any method using that mutation e.g. upgradeFacilitatorAccessToken

The logic insidegetOnAuth from props/onAuth.js module can be removed, ensure that this module continues to return a resolved promise with accessToken and the logs spb_onauth_access_token_present/not_present are emitted. Also the method must set the access to cache using setBuyerAccessToken

❤️  Thank you!


src/props/onApprove.test.js
```diff
@@ -148,42 +148,6 @@ describe("onApprove", () => {
         expect.anything()
       );
     });
-
-    test("invoke upgradeFacilitatorAccessTokenWithIgnoreCache call if treatment is present", async () => {
```
ravishekhar, line 152: Since the default use case now is always upgradeFacilitatorAccessTokenWithIgnoreCache() to be called, so I assume we need a test validating the code path.

src/props/onAuth.js
```diff
@@ -17,47 +14,24 @@ export type XOnAuthDataType = {|
...
-export function getOnAuth({ facilitatorAccessToken, createOrder, createSubscription, featureFlags, experiments } : GetOnAuthOptions) : OnAuth {
+export function getOnAuth({ createSubscription, featureFlags } : GetOnAuthOptions) : OnAuth {
     return ({ accessToken } : XOnAuthDataType) => {
         getLogger().info(`spb_onauth_access_token_${ accessToken ? 'present' : 'not_present' }`);
-        const isInIgnoreCacheExperiment = experiments?.upgradeLSATWithIgnoreCache;
 
         return ZalgoPromise.try(() => {
             if (accessToken) {
-                if (createSubscription || isInIgnoreCacheExperiment) {
+                if (createSubscription ) {
```
ravishekhar, line 27: I think we can even remove this `if (createSubscription )` condition right ? 
We need to ensure we are caching 
`setBuyerAccessToken(accessToken);` for all cases now, as `isInIgnoreCacheExperiment` is always `true` after experiment is ramped to 100%.

src/props/onAuth.js
```diff
@@ -17,47 +14,24 @@ export type XOnAuthDataType = {|
...
                     // Cache the buyerAccessToken so that it can be used when creating an upgraded lsat with ignore cache
                     setBuyerAccessToken(accessToken);
                     getLogger().info("ignore_lsat_upgrade", {
```
ravishekhar, line 30: we can remove this log, as we always ignore lsat upgrade at this point

ravishekhar (changes requested):

mnicpt:
@xuanwan05 There is a ton of changes outside of removing old LSAT code. Are we applying a different lint? I almost feel like the cleanup should happen in another PR and describing what is being cleaned up.

ravishekhar:
> @xuanwan05 There is a ton of changes outside of removing old LSAT code. Are we applying a different lint? I almost feel like the cleanup should happen in another PR and describing what is being cleaned up.

@mnicpt We are in mid of reformatting this PR as per newly formatted  files in main branch after prettier run.

PR #709, opened 2023-11-02
Run Prettier on entire codebase

nikrom17:
### Description

This PR add the default Prettier config to the repo and runs the formatter on all the appropriate files

We don't want to worry about code formatting so we will use prettier to handle that for us

### Important Files to View
.editorconfig
.prettierrc.json
.prettierignore
test/client/mocks.js:2041 (added /* eslint-disable no-undef */ for the entire type block since prettier was removing the eslint-disable-line when placed above line 2043)

Some updates were made to other eslint-disable-line comments but there were no issues moving these to the appropriate place after formatting 

❤️  Thank you!


package.json
```diff
@@ -47,7 +49,7 @@
     "release:dry-run": "DRY_RUN=true grabthar-release",
     "setup": "npm install && npm run flow-typed",
     "start": "npm run webpack -- --progress --watch",
-    "test": "npm run coverage && npm run karma",
+    "test": "npm run coverage && npm run format:check && npm run karma",
```
jshawl, line 52: If you want the `format:check` can come before `npm run coverage` so it will fail a bit faster if unformatted

package.json
```diff
@@ -88,20 +90,21 @@
...
+    "eslint-config-prettier": "^9.0.0",
     "express": "^4.16.3",
     "flow-bin": "0.155.0",
     "flow-typed": "^3.8.0",
     "happy-dom": "^9.0.0",
     "husky": "^8.0.2",
     "jsdom": "^22.0.0",
     "lint-staged": "^13.0.4",
-    "prettier": "^2.8.0",
+    "prettier": "2.8.8",
```
jshawl, line 101: option to add the `^` back

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[**]
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
```
jshawl, line 15: I am good with this as-is but am wondering if it should match checkout-components - https://github.com/paypal/paypal-checkout-components/blob/main/.editorconfig 

but also do you have this plugin installed? https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig in VSCode? 

I've not used `.editorconfig` before so just curious!

.prettierrc.json
```diff
@@ -1 +1,21 @@
...
+  "quoteProps": "as-needed",
+  "requirePragma": false,
+  "semi": true,
+  "singleAttributePerLine": false,
+  "singleQuote": false,
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false,
+  "vueIndentScriptAndStyle": false
+}
```
jshawl, line 21: I am good w/ these changes too but am wondering how you knew what to put in here? I have a slight preference for not overriding prettier defaults (only from a maintenance perspective) but am just curious about how you knew which properties and values to add here.

package.json
```diff
@@ -88,20 +90,21 @@
     "@vitest/ui": "^0.25.3",
     "babel-core": "^7.0.0-bridge.0",
     "cross-env": "^7.0.3",
+    "eslint-config-prettier": "^9.0.0",
```
jshawl, line 93: Are there any conflicting rules in the code now that would cause an issue _without_ this plugin? I'm also thinking about parity w/ checkout-components and in the near future abstracting the prettier config to a shared space (checkout-components does not have this plugin so that prompted my question! https://github.com/paypal/paypal-checkout-components/blob/main/package.json#L69 )

jshawl (approved):
Nice, Nik! I left a few non-blocking questions/comments mostly related to limiting customizations / differing from paypal-checkout-components, but I like what you did here!

.editorconfig
nikrom17: I first started by looking at the .editorconfig files in checkout-components and clientskdnodeweb but a lot of the settings in the .editorconfig file were conflicting with the prettier default values.

The similar settings set in the .editorconfig file would overwrite the default prettier values if they weren't explicitly set in a .prettierrc file. Since we explicitly set the values in the prettierrc, setting similar values in the .editorconfig has no effect.

Some values are possible to be set in the editorconfig that can't be set in the pretterrc (like trim_trailing_whitespace). I kept that value.

The editorconfig settings typically get applied as you write code and prettier is usually run after you write code. This is why I wanted to keep a few basic settings in the .editorconfig file so there isn't such a big difference in the code when you apply prettier. That being said, it is recommend to format the code on save so you don't really notice the .editorconfig styling vs prettier.

This is from the prettier docs and what they recommend for an editor config if you go with the default prettier values.
https://prettier.io/docs/en/configuration#editorconfig

jshawl: ah awesome! thank you for explaining that!

.prettierrc.json
nikrom17: The default prettier values are documented here: https://prettier.io/docs/en/options

They specify the default prettier config on this page and I verified all these settings were the default. 

Here is also where they document the configuration overrides: https://prettier.io/docs/en/configuration

I feel explicitly setting the default values in the rc file will ensure we don't need to worry about values from a .editorconfig file or the prettier extension overriding one of these settings.

package.json
nikrom17: I did not check to see if there were any eslint/prettier conflicts before installing this package. I was going through the prettier install section on their docs and this was recommended as part of the prettier install process

https://prettier.io/docs/en/install#eslint-and-other-linters

The eslint-plugin also ships with a CLI to identify potential eslint/prettier rule conflicts. Below are some screen shtos of running that tool on checkout-components vs SPB (with prettier plugin enabled)

Screenshot 2023-11-03 at 12 35 07 PM

Here is one of SPB after remove eslint-prettier plug in
Screenshot 2023-11-03 at 12 37 59 PM

I feel like add it is the way to go since it seems to resolve the warnings and something the prettier docs officially recommend. I would recommend that all repos using prettier add it as well.

jshawl: ah nice! thank you for digging in on this one!

.prettierrc.json
jshawl: Awesome! thank you for explaining this! (TIL)

nbierdeman (approved):
Nice! 🔥 🥇 🚢

.prettierrc.json
wsbrunson, line 21: We've been leaving this blank in our other repos as well: https://github.com/paypal/paypal-checkout-components/blob/main/.prettierrc.json

the reasoning is we don't want to debate about any of the options or have to do any work when versions change. We take whatever prettier says and we understand that breaking changes to prettier may mean formatting changes

wsbrunson (commented):
one comment but other than that, looks great!

.prettierrc.json
nikrom17: I explicitly added the default prettier configurations so we didn't have to worry about settings from a .editorconfig file or the vs code extension overriding the default settings, but I can remove for constancy.

SPB team talked briefly about creating a shared configuration for all of JS SKD team to use: https://prettier.io/docs/en/configuration#sharing-configurations

What are you thoughts on creating something like this (for a separate ticket/PR)?

wsbrunson: we can keep it on our radar but my 2cents would be: keep this as simple as possible for now until we feel the pain of not having a shared config. So far, going no config has been working well

wsbrunson (approved):

PR #710, opened 2023-11-07
Remove btSdkOrdersV2Migration experiment

nikrom17:
### Description

Remove unused elmo experiment btSdkOrdersV2Migration

[DTPPCPSDK-1271](https://paypal.atlassian.net/browse/DTPPCPSDK-1271)

Experiment will also need to be removed from [smartcomponentnodeweb](https://github.paypal.com/Checkout-R/smartcomponentnodeweb)

❤️ Thank you!

### DIFF
--- a/src/props/onApprove.test.js
+++ b/src/props/onApprove.test.js
@@ -96,26 +96,10 @@ describe("onApprove", () => {
       ...commonOptions,
       beforeOnApprove: vi.fn(),
     };
-    test("invokes the merchant's onApprove call with an undefined paymentID", async () => {
-      const getOnApproveOrderResult = getOnApproveOrder(
-        getOnApproveOrderOptions,
-      );
-      await getOnApproveOrderResult({}, { restart });
-      expect(merchantOnApprove).toHaveBeenCalledWith(
-        expect.objectContaining({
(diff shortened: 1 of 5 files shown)


src/props/onApprove.test.js
```diff
@@ -96,26 +96,10 @@ describe("onApprove", () => {
       ...commonOptions,
       beforeOnApprove: vi.fn(),
     };
-    test("invokes the merchant's onApprove call with an undefined paymentID", async () => {
```
westeezy, line 99: Is this test still good to keep? Is it possible BT sends an undefined paymentId?

nikrom17: I paired with Jesse on this ticket and I believe we reviewed and decided to keep it. That being said, I'm not 100% on that and happy to test/verify anything to double check. I will however require some assistance/guidance with that.

wsbrunson (approved):

src/props/onApprove.test.js
westeezy: This makes sense to me especially as I look at the surrounding tests. Marking as resolved

westeezy (approved):

PR #720, opened 2023-12-15
[DTPPCPSDK-1600] Pass the same value as CMID to OneClickApproveOrder as we are passing in the GetSmartWallet call

xuanwan05:
### Description
Either `CMID` or `sessionId` is passed as the Paypal-Client-Metadata-Id header when calling GetSmartWallet. For the OneClickApproveOrder mutation, either `CMID` or `orderId` is passed in the Paypal-Client-Metadata-Id headers. Therefore, we are changing the Paypal-Client-Metadata-Id headers of OneClickApproveOrder to pass the sessionId instead of the orderId, similar to GetSmartWallet.

Update:
from Shane: clientMetadataID is already set to sessionID if it doesn't exist, this happens in `checkout components`
so we only need to set aypal-Client-Metadata-Id headers to clientMetadataID.

❤️ Thank you!


src/api/order.js
```diff
@@ -1135,11 +1142,7 @@ export function oneClickApproveOrder({
             }
         `,
     variables: { orderID, instrumentType, instrumentID },
-    headers: {
-      [HEADERS.ACCESS_TOKEN]: buyerAccessToken,
-      [HEADERS.CLIENT_CONTEXT]: orderID,
-      [HEADERS.CLIENT_METADATA_ID]: clientMetadataID || orderID,
```
jshawl, line 1141: Could this instead be:

```
[HEADERS.CLIENT_METADATA_ID]: clientMetadataID
```

e.g. remove the `|| orderID,`? I think if `clientMetadataID` is `undefined` it will not be sent in the headers

src/api/order.test.js
```diff
@@ -165,4 +192,39 @@ describe("actions smart api fallback cases", () => {
...
+
+  describe("oneClickApproveOrder calls the mutation", () => {
+    it("with clientMetadataID", async () => {
+      const clientMetadataID = "test-id";
+      await oneClickApproveOrder({
+        clientMetadataID,
+        ...oneClickApproveOrderOptions,
+      });
+
+      expect(callGraphQL).toHaveBeenCalled();
```
jshawl, line 204: you could also test:

```
expect(callGraphQL).toHaveBeenCalledWith(expect.objectContaining({headers: ... }))
```

to make sure the headers are set correctly - and then I don't think we need to check the `resolves.toEqual` below (for this particular test)

src/api/order.js
```diff
@@ -1114,6 +1114,13 @@ export function oneClickApproveOrder({
   buyerAccessToken,
   clientMetadataID,
 }: OneClickApproveOrderOptions): ZalgoPromise<ApproveData> {
+  const headers = {};
+  headers[HEADERS.ACCESS_TOKEN] = buyerAccessToken;
+  headers[HEADERS.CLIENT_CONTEXT] = orderID;
+  if (clientMetadataID) {
```
jshawl, line 1120: I think this will always be true (for the wallet capture) based on this line https://github.com/paypal/paypal-smart-payment-buttons/pull/720/files#diff-a6c146bc209bbfc0b3eaaf82d6df6cb566fe474e27f0416e295712b4b0e284c0R304

but that makes me want to ask: is there a `oneClickApproveOrder` that _should not_ have a `clientMetadataID` set?

jshawl (commented):
Wanted to leave some feedback early but will plan to come back soon! Lookin good!

src/api/order.js
xuanwan05: `clientMetadataID` is either `cmid` or `sessionId` from `props`, and either of them could be null/undefined.

src/api/order.js
xuanwan05: The type of the header value is always a string, making it incompatible with null or undefined

src/api/order.js
```diff
@@ -1114,6 +1114,13 @@ export function oneClickApproveOrder({
   buyerAccessToken,
   clientMetadataID,
 }: OneClickApproveOrderOptions): ZalgoPromise<ApproveData> {
+  const headers = {};
+  headers[HEADERS.ACCESS_TOKEN] = buyerAccessToken;
+  headers[HEADERS.CLIENT_CONTEXT] = orderID;
```
wsbrunson, line 1119:
```suggestion
  const headers = {
    [HEADERS.ACCESS_TOKEN]: buyerAccessToken,
    [HEADERS.CLIENT_CONTEXT]: orderID,
  };
```

src/payment-flows/wallet-capture.js
```diff
@@ -300,14 +301,16 @@ function initWalletCapture({
           throw new Error(`Instrument has no type`);
         }
 
+        const cmid = clientMetadataID || sessionID;
```
wsbrunson, line 304: `clientMetadataID` is already set to `sessionID` if it doesn't exist, this happens in `checkout components
```suggestion
        const cmid = clientMetadataID || sessionID;
```
```suggestion
```

src/payment-flows/wallet-capture.js
```diff
@@ -300,14 +301,16 @@ function initWalletCapture({
...
+
         return ZalgoPromise.hash({
           requireShipping: shippingRequired(orderID),
           orderApproval: oneClickApproveOrder({
             orderID,
             instrumentType,
             buyerAccessToken,
             instrumentID,
-            clientMetadataID,
+            clientMetadataID: cmid,
```
wsbrunson, line 313:
```suggestion
            clientMetadataID,
```

jshawl (approved):
:shipit:

wsbrunson (approved):

ravishekhar (approved):

PR #721, opened 2023-12-20
Clearing out all fields upon form submission

doniarobinson:
We were notified that the .clear() function was not clearing out the credit card number and expiration date for ACDC card fields. It is noted that if all of the user input isn't cleared, it can potentially lead to a CARD_NUMBER_REQUIRED error.

* Submit
* Watch for all fields, including credit card icon, to be cleared out after form submits

❤️ Thank you!


src/card/components/CardExpiry.jsx
```diff
@@ -162,6 +162,7 @@ export function CardExpiry({
         ref={expiryRef}
         type={type}
         className="card-field-expiry"
+        value={maskedInputValue}
```
jshawl, line 165: this codecov comment looks angry but from what I can tell there was no coverage for this group of lines previously - https://app.codecov.io/gh/paypal/paypal-smart-payment-buttons/pull/721/blob/src/card/components/CardExpiry.jsx#L165 and I'm not sure how feasible it would be to add coverage for all of the props (I personally would ignore the codecov comment)

src/card/components/CardNumber.jsx
```diff
@@ -120,6 +120,13 @@ export function CardNumber({
   const numberRef = useRef();
   const ariaMessageRef = useRef();
 
+  const resetDefaultCardType = (
+    element: HTMLElement,
+    isFocused: boolean,
+  ): boolean => {
+    return element && !isFocused && inputState.inputValue.length === 0;
```
jshawl, line 127: nice! this function `resetDefaultCardType` _might_ be a good candidate for a unit test but I'm not sure how feasible it is to test `inputState` as it is out of scope of this function.

jshawl (approved):
Nice! :shipit: also congrats on your first SPB PR!

PR #724, opened 2024-01-03
[DTEVEREST-7976]Limit category check only for donations created with ORDERS V2 API and update product feature of donate txns

arielxin:
### Description

Issue:
1. SPB with donate label are being blocked for vault without payment flow. 
2. DONATION CCO is not being properly passed for donation transactions made via SPB

Existing merchants want to use SPB with donate label with vault without payment API, but the item category check we built earlier in SPB is blocking this kind of integration.

❤️ Thank you!


spencersablan:
Closing since we are no longer accepting PR's here!  Ariel has agreed to reopen a PR in SCNW once SPB is moved over.

PR #725, opened 2024-01-10
 Clearing isValid and isPotentiallyValid back to default values

doniarobinson:
### Description
We were notified that two sub-fields on the main form fields were not being cleared out properly: isValid and isPotentiallyValid. This ticket clears out those fields after returning to the form, and also removes a dependency where CVV isValid and isPotentiallyValid were being affected by CardType.

LI-31748 - current ticket

* Submit
* Run `cardField.getState()` in the console when it returns
* Check to see that value.fields.\*.isValid are all false and value.fields.\*.isPotentiallyValid are all true

Screenshot 2024-01-10 at 10 04 20 AM

❤️ Thank you!


src/card/components/CardCVV.jsx
```diff
@@ -100,22 +100,6 @@ export function CardCVV({
...
-  }, [cardType]);
-
-  useEffect(() => {
-    const validity = cardValidator.cvv(inputValue, cardType?.code?.size);
-    if (touched) {
-      validity.isPotentiallyValid = false;
-    }
-    setInputState((newState) => ({ ...newState, ...validity }));
-  }, [cardType]);
-
```
wsbrunson, line 118: i think we do want this. Lets say you put in a 3 digit CVV code but change the card type from visa to amex which has a 4 digit CVV code. We'd need to re-validate CVV otherwise the card would fail to validate but there'd be no UI to indicate why

doniarobinson: Yeah, that sounds pretty reasonable. I'm going to close this out in favor of a draft so I can keep working on this. Thanks!

PR #727, opened 2024-01-11
Clearing isValid and isPotentiallyValid back to default values

doniarobinson:
### Description
We were notified that two sub-fields on the main form fields were not being cleared out properly: isValid and isPotentiallyValid. This ticket clears out those fields after returning to the form, removes a duplicate useEffect, and modifies the remaining useEffect to not update CVV when the cardType is unknown.

LI-31748 - current ticket

* Submit
* Run `cardField.getState()` in the console when it returns
* Check to see that value.fields.\*.isValid are all false and value.fields.\*.isPotentiallyValid are all true

Screenshot 2024-01-10 at 10 04 20 AM

❤️ Thank you!


doniarobinson:
Closing because it is supposed to be in SCNW now

